PR #1, opened 2016-10-14
Initialize control-node node and tests-machines

MalloZup:


salt/control-node/init.sls
```diff
@@ -0,0 +1,20 @@
+include:
```
MalloZup: this is just a fake configuration, for testing purposes..

salt/control-node/init.sls
```diff
@@ -0,0 +1,89 @@
+init-control-node-repo:
+# control node is always sles-12-sp1, so stay true to KISS.
+  cmd.run:
+    - name: |
+        zypper addrepo http://download.suse.de/ibs/Devel:/Galaxy:/ruby:/test/SLE_12/Devel:Galaxy:ruby:test.repo
```
moio: Please use `file.managed` and `.repo` files here, see for example https://github.com/moio/sumaform/blob/master/salt/client/repos.sls#L5. This makes it easier to adapt for package-mirror later, which helps home office users and people who want to show this while travelling (example: customer demos).

Refreshing can be done like this: https://github.com/moio/sumaform/blob/master/salt/client/repos.sls#L56

salt/control-node/init.sls
```diff
@@ -0,0 +1,89 @@
...
+        zypper addrepo http://download.suse.de/ibs/SUSE:/SLE-12:/Update/standard/SUSE:SLE-12:Update.repo
```
moio: Please use the existing SLE-12 repo definition: https://github.com/moio/sumaform/tree/master/salt/sles
Like this: https://github.com/moio/sumaform/blob/master/salt/client/init.sls#L1

That will pull in the base SLES repos (from package-mirror if so configured) and refresh them.

salt/control-node/init.sls
```diff
@@ -0,0 +1,89 @@
...
+        zypper addrepo http://download.suse.de/ibs/Devel:/SLEnkins/SLE_12_SP1/Devel:SLEnkins.repo
```
moio: What packages are taken from this repo?

salt/control-node/init.sls
```diff
@@ -0,0 +1,89 @@
...
+      - ruby2.1-rubygem-jwt
+      - rubygem-mime-types
+      - ruby2.1-rubygem-builder
+      - rubygem-owasp_zap
+      - rubygem-cliver
+      - rubygem-twopence
+gems-installation:
+
+  gem.installed:
+    - names:
```
moio: Question for later: how difficult is it to package (or use packaged versions of) these gems?

MalloZup:
thx @moio for review,  i will work on this next week.

salt/control-node/init.sls
MalloZup: `http://download.suse.de/ibs/Devel:/Galaxy:/ruby:/test/SLE_12/Devel:Galaxy:ruby:test.repo`

from this repo we take the ruby-gems that run the cucumber suite.

The slenkins repo is mainly only for twopence, the test-executor that run virtio/ssh commands for testing-machines. ( we could move this to the ruby repo).

the update repo, is because some rubygems needed are in this repo. I will than at the end clean up more this repo.

We don't use the sles-12-sp1 repo for ruby test, because some gems are not building in this repo...
SInce is the control-node, we can make for moment this "hack".. until we clean up..

salt/control-node/init.sls
MalloZup: @moio, i didn't investigate for moment how difficult is to package this gems. I think the difficulty level is not high, but it a tedious and could be a long work to package and tests the stuff. sure, this has to be done. Maybe there are already some gems packaged in some other projects, and we have just to migrate this package to our ruby-repo.

moio:
Work continues in the [cucumber_testsuite](https://github.com/moio/sumaform/tree/cucumber_testsuite) branch

PR #4, opened 2016-10-19
Add package-mirror storage pool indication

mbologna:


moio:
Already explained in detail here: https://github.com/moio/sumaform/tree/master/modules/libvirt#choose-whether-to-use-a-package-mirror-host

PR #12, opened 2016-11-11
update main.tf with sles12sp2

MalloZup:
improve the support to sles-12-sp2

### DIFF
--- a/modules/libvirt/base/main.tf
+++ b/modules/libvirt/base/main.tf
@@ -28,6 +28,12 @@ resource "libvirt_volume" "sles12sp1" {
   pool = "${var.pool}"
 }
 
+resource "libvirt_volume" "sles12sp2" {  
+  name = "${var.name_prefix}sles12sp2"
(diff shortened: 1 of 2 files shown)


MalloZup:
i branched already the RPM, and did the build locally and in my home for sles-12-sp2.
obs.suse. seems pretty straight forward to build images.

Question, why in sp1 we had sles-12 repos? 

I think that in the kiwi we should have only sles-12-sp2 repo. what do you think?

moio:
> Question, why in sp1 we had sles-12 repos?
> I think that in the kiwi we should have only sles-12-sp2 repo. what do you think?

I guess you refer to `<repository>` tags in the Kiwi files here.

Those actually refer to BS projects rather than distributed repos - we need them all as packages in the repos distributed to customers might come from earlier versions.

MalloZup:
@moio ok thx. :)

moio:
Reopened, this is still relevant!

PR #15, opened 2016-11-15
Add CentOS 7 image

mbologna:
Add support for spinning up CentOS 7 images.
**WARNING**: due to package-mirror dependency on base/main.tf#configuration , your package-mirror will be **deleted** when applying this patch. Backup accordingly.

### DIFF
--- a/modules/libvirt/base/main.tf
+++ b/modules/libvirt/base/main.tf
@@ -34,6 +34,12 @@ resource "libvirt_volume" "sles12sp2" {
   pool = "${var.pool}"
 }
 
+resource "libvirt_volume" "centos7" {
+  name = "${var.name_prefix}centos7"
+  source = "http://w3.suse.de/~mbologna/sumaform-images/centos7.qcow2"
+  pool = "${var.pool}"
+}
+  
 output "configuration" {
   // HACK: work around https://github.com/hashicorp/terraform/issues/9549
   value = "${
--- a/modules/libvirt/host/variables.tf
+++ b/modules/libvirt/host/variables.tf
@@ -9,7 +9,7 @@ variable "name" {
 }
 
 variable "image" {
-  description = "One of: opensuse421, sles11sp3, sles11sp4, sles12, sles12sp1"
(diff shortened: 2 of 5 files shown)


README_ADVANCED.md
```diff
@@ -118,6 +118,25 @@ module "pg2" {
 }
 ```
 
+## CentOS hosts
```
moio: This is not really needed here, please document instead in

https://github.com/moio/sumaform/blob/master/modules/libvirt/host/variables.tf#L12
https://github.com/moio/sumaform/blob/master/modules/libvirt/minion/variables.tf#L12

and possibly in:

https://github.com/moio/sumaform/blob/master/modules/libvirt/client/variables.tf#L12
(if tested)

modules/libvirt/base/main.tf
```diff
@@ -37,6 +43,7 @@ output "configuration" {
       "sles11sp4", "${libvirt_volume.sles11sp4.id}",
       "sles12", "${libvirt_volume.sles12.id}",
       "sles12sp1", "${libvirt_volume.sles12sp1.id}",
+      "centos7", "${libvirt_volume.centos7.id}",
```
moio: When rebasing simply avoid this line.

salt/minion/init.sls
```diff
@@ -1,12 +1,19 @@
 include:
+  {% if grains['os'] == 'SUSE' %}
```
moio: It sounds wrong to me to solve this here - the client.repos state takes care of client repos, regardless of the distro.

Please solve this in https://github.com/moio/sumaform/blob/master/salt/client/repos.sls.

salt/minion/init.sls
```diff
@@ -1,12 +1,19 @@
...
   - client.repos
+  {% endif %}
   - minion.testsuite
 
-salt-minion:
+
+minion-installation:
```
moio: Why is renaming necessary here?

mbologna:
@moio requested changes pushed.

PR #16, opened 2016-11-16
Fix: Correctly generate machine-id on SLES 11

mbologna:


salt/terraform-resource/init.sls
```diff
@@ -49,9 +49,15 @@ configure-full-avahi-resolution:
 
 setup-machine-id:
   cmd.run:
+    {% if grains['osrelease'] >= '12.0' %}
```
moio: This assumes SLES which might not be the case (CentOS anyone?)

One alternative is to use `unless: systemd-machine-id-setup` which complements the above case

mbologna:
@moio corrected

PR #18, opened 2016-11-17
Feat: package-mirror data disk with XFS

mbologna:
This patch changes the file-system for package-mirror data disk to XFS
(for obvious performance reasons).
This patch applies only to newly created package-mirrors.

NOTE: existing package-mirrors cannot be used with package mirrors created after applying this patch.

### DIFF
--- a/salt/package-mirror/init.sls
+++ b/salt/package-mirror/init.sls
@@ -37,7 +37,7 @@ mirror-script:
 
 mirror-partition:
   cmd.run:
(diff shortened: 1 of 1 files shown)


moio:
I am not accepting this right now as it requires rebuilding the package-mirror's data disk, which is an expensive operation for what I believe to be a very small performance gain given the use-case.

It is possible that in future we will have to do it anyway, in that case I will roll this patch in.

Otherwise feel free to post performance numbers that prove this is really very beneficial!!!

Thanks for this effort anyways :smile:

mbologna:
http://www.linux-magazine.com/Online/Features/Filesystems-Benchmarked

PR #22, opened 2016-11-23
Adding a git-config prerequisite, for cloning cucumber-suite

MalloZup:


salt/control-node/init.sls
```diff
@@ -71,9 +71,25 @@ cucumber-requisites:
 cucumber-testsuite:
   git.latest:
     - name: https://github.com/SUSE/spacewalk-testsuite-base.git
-    - rev: slenkins
+    {% if 'manager30' in grains['branch'] %}
+    - rev: manager30 
+    {% elif 'refhost30' in grains['branch'] %}
```
moio: Reference hosts are a totally different matter and should come as a separate PR.

We might even decide not to set them up with cucumber at all, so this is surely for later.

salt/control-node/init.sls
```diff
@@ -71,9 +71,25 @@ cucumber-requisites:
...
+    - rev: manager30 
+    {% elif 'master' in grains['branch'] %}
```
moio: Please use `else` here - basically "in all other cases use master"

salt/control-node/init.sls
```diff
@@ -71,9 +71,25 @@ cucumber-requisites:
...
+    - rev: master
+    {% endif %}
     - target: /root/spacewalk-testsuite-base
 
+git-config:
+  git.config_set:
+    - name:  http.postBuffer
+    - value: 2M
+    - global: True
+    - require:
```
moio: I think this is the other way round: `cucumber-testsuite` should require `git-config`, right?

MalloZup: yeah, true.

MalloZup:
This can be merged only when the slenkins branch go to manager30 branch.

MalloZup:
not needed, workaround doesn't fix.

PR #25, opened 2016-11-23
 Use zip format instead of git, add branch var

MalloZup:


salt/control-node/init.sls
```diff
@@ -69,10 +71,11 @@ cucumber-requisites:
...
 
 cucumber-testsuite:
-  git.latest:
-    - name: https://github.com/SUSE/spacewalk-testsuite-base.git
-    - rev: slenkins
-    - target: /root/spacewalk-testsuite-base
+  cmd.run:
+    - name: | 
+        wget -P /root/ https://github.com/SUSE/spacewalk-testsuite-base/archive/slenkins.zip
+        unzip /root/slenkins.zip -d /root/
```
moio: Please use `cmd.run` only to call `wget` in order to work around the lack of `skip_verify` in `file.managed` and document it with a comment as a `HACK`, so that we can get rid of it once we get newer Salt versions.

Please continue using `archive.extracted` to extract the archive.

Ping me on IRC if you need help :wink:

MalloZup:
@moio, for me looks ok. now

salt/control-node/init.sls
```diff
@@ -68,11 +70,34 @@ cucumber-requisites:
...
-  git.latest:
-    - name: https://github.com/SUSE/spacewalk-testsuite-base.git
-    - rev: slenkins
-    - target: /root/spacewalk-testsuite-base
+suite-download:
+  cmd.run:
+    # HACK only keep this until we update to salt higher version(2016.3.4.), for skip_verify:False
+    - name: |
+        wget -P /root/ https://github.com/SUSE/spacewalk-testsuite-base/archive/slenkins.zip
+
```
moio: You also need a `creates` argument here to avoid downloading multiple times in case of multiple state application (states should always be idempotent - that is get to the same state regardless of the starting condition). Example:

https://github.com/moio/sumaform/blob/master/salt/suse-manager/rhn.sls#L74

MalloZup: ok thx

MalloZup:
@moio, for me looks ok

moio:
Cherry-picked to `master` after small editorial changes.

Thank you!

PR #28, opened 2016-11-23
Speed-up tests-machines

MalloZup:


MalloZup:
https://github.com/moio/sumaform/issues/27

moio:
Hey, thank you for this PR!

I am afraid I do not agree with the current solution though.

The reason why that dependency is mandatory is that we typically want to wait for VMs - this is needed for correct functioning in all cases apart from the testsuite, which is important but not the main use case for sumaform.

With this patch we "lose" the parser error if we forget `server_configuration`, and that's not a good thing.

Instead, I would propose to specify a custom map for the specific case of the testsuite. This would be optional (the testsuite works with standard configuration anyway, albeit slower) and retain the parser error in case we forget something for non-testsuite use cases.

I am proposing something like this in `main.tf.libvirt-testsuite.example`:
```hcl
server_configuration = { hostname =  "suma3pg.tf.local" }
```

Does this make sense to you?

MalloZup:
@moio fixed, for me make sense what you want. I can agree. with this, we can profite both. Thx for review and suggestion

PR #37, opened 2016-12-05
Sle12sp2 support

mseidl:
This allows changing the base os for suma, a hack for making salt work on sp2(until bug is fixed), updated documentation


moio:
Rebased, slightly corrected and merged in `master`.

Thank you!

PR #38, opened 2016-12-09
Enable first support for centos7 on sumaform-testsuite

MalloZup:


MalloZup:
@moio, for me this PR is fine, for inital supporting the centos7 systems.

Next pr will create the channels

moio:
Content-wise it's OK, but I don't like the mix of "centos" and "rh" or "redhat" in the names.

Please follow the unwritten convention "min<OS_NAME>" - we already have `clisles12sp1` and `minsles12sp1`, this should be named `mincentos7` and all other variables should refer to CentOS rather than Red Hat.

Besides that, [Red Hat is a registered trademark](http://brand.redhat.com/logos/trademarks/) and I do not want to get into the mess of attributing the usage correctly. Please do not refer to Red Hat in variable names, etc. - rename all to be consistent to Centos in naming.

Thanks!!!

moio:
Merged, thanks

PR #40, opened 2016-12-09
Feature: access systems without rsa-key

MalloZup:
Until now, for accessing the centos image, we have to put a ssh-key
This is because of redhat standart policy.
For testing systems we prefer to use password auth.


MalloZup:
something in the HOST ssh_config did the troubleshooting for redhat systems.

PR #41, opened 2016-12-10
perform a salt self update in seperate salt-call to prevent failures

mcalmer:
Doing a salt major version update in a big highstate result into errors.
Seems that salt re-load state and modules code after the update but not
imported utilities.
That means: new modules call old (while in memory) utility functions with
the consequence that a called function might not exist.

### DIFF
--- a/modules/libvirt/host/main.tf
+++ b/modules/libvirt/host/main.tf
@@ -59,6 +59,7 @@ EOF
   provisioner "remote-exec" {
     inline = [
       "salt-call --force-color --local state.sls terraform-resource",
+      "salt-call --force-color --local state.sls selfupdate",
       "salt-call --force-color --local state.highstate"
     ]
   }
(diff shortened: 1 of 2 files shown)


MalloZup:
+1

moio:
My problem with this patch is that sumaform gets to deploy systems using development/unstable salt, while by design I only use stable/SLES salt to keep deployment stable. I did use development Salt in the past but that added uncertainty to the process and I would try to avoid it.

If possible I would like to always use stable/SLES salt for sumaform deployment - the state can upgrade salt or even remove it completely depending on needs, but this is probably the very last thing that needs to be done.

Can this made to work by using dependencies to have the salt upgrade as the last thing in the highstate?

moio:
Alternative fix works. Will merge it soon. Closing this PR, thanks anyway for reporting and proactively coming up with a fix!

PR #45, opened 2016-12-14
Aaa base for testsuites machines

MalloZup:


MalloZup:
https://github.com/moio/sumaform/issues/43

salt/client/testsuite.sls
```diff
@@ -20,6 +20,7 @@ cucumber-requisites:
       - man
       - wget
       - adaptec-firmware
+      - aaa_base-extras
```
moio: This does not exist in CentOS as far as I know. We have to guard against non-SUSE OS's now that we support it.

MalloZup: Ok,  doing jonja ninja then

MalloZup:
@moio, ok for me

PR #46, opened 2016-12-16
BUGFIX: restart salt-api service.

MalloZup:


MalloZup:
@moio , i'm testing it now

PR #48, opened 2016-12-23
Only restart 'salt-api' is salt-api is installed.

meaksh:
Salt is used by sumaform for provisioning, so `salt` package is always installed in every deployed machine but not the `salt-api` package.

This PR fixes the "HACK" to restart the `salt-api` in order to prevent errors during provisioning if `salt-api` doesn't exist.

Also prevent an error trying to run `systemctl restart salt-api` if running OS uses SysV.

### DIFF
--- a/salt/suse-manager/salt.sls
+++ b/salt/suse-manager/salt.sls
@@ -24,6 +24,6 @@ salt:
 # and restart.sls from salt don't work
 salt-api:
   cmd.run:
-   - name: systemctl restart salt-api
+   - name: service salt-api restart
    - require:
-      - pkg: salt
+      - pkg: salt-api


MalloZup:
hi @meaksh , we did this Hack, because the salt-api needed to be restarted after salt was reupdated.

meaksh:
Hey @MalloZup , since `salt-api` requires the same version of the installed `salt` package, then IIUC both `salt` and `salt-api` packages are updated if you upgrade the `salt` package. Therefore, the "HACK" should keep restarting the `salt-api` after `salt` is updated.

MalloZup:
@meaksh, ok for me then. Thx, for fixing the hack of the hack. See you soon, happy hacking.. 

P.S we have to wait for @moio, he press the pedal :)

meaksh:
@MalloZup - Thanks. Happy hacking and have a great vacations!

Let's wait for @moio :smile:

moio:
Fine for the moment, the work I am doing will make this outdated anyway and remove the hack, but it's not ready yet.

Thanks!

PR #56, opened 2017-01-10
Enable var "version" for minion

MalloZup:


MalloZup:
@moio, take a look when you are free

modules/libvirt/minion/variables.tf
```diff
@@ -13,6 +13,11 @@ variable "image" {
   type = "string"
 }
 
+variable "version" {
+  description = "One of: stable,  nightly"
+  type = "string"
```
moio: Please set "stable" as the default so that old `main.tf`s do not break.

main.tf.libvirt-testsuite.example
```diff
@@ -16,19 +14,18 @@ module "base" {
 }
 
 module "suma3pg" {
-  source = "./modules/libvirt/suse_manager"
-  base_configuration = "${module.base.configuration}"
-
+ source = "./modules/libvirt/suse_manager"
```
moio: This change of indentation was probably accidental.

main.tf.libvirt-testsuite.example
```diff
@@ -50,17 +47,18 @@ module "minsles12sp1" {
...
-
+  version = "nightly"
   name = "minsles12sp1ssh"
   image = "sles12sp1"
   server_configuration =  { hostname = "suma3pg.tf.local" }
   for_development_only = false
   for_testsuite_only = true
 }
-
+
```
moio: This new trailing space was probably accidental.

modules/libvirt/minion/variables.tf
MalloZup: done

main.tf.libvirt-testsuite.example
MalloZup: this space is after a module declaration, so it's ok

MalloZup:
@moio,
fixes done.
Readme-advanced note ok

I cannot propagate the var to aws or cloud backends, since this 2 backends has not the same schema for the modules like libvirt. 

Imho the Pr is fine for libvirt. For aws and cloud, a new redisign is needed.

main.tf.libvirt-testsuite.example
moio: Trailing spaces (right before newlines)are always forbidden :smile:

MalloZup: double checked, it's not a trailing space but a new-line. the gui review is showing that

moio:
In the README you mention `version` works for both clients and minions, but the variable was only added to the minion module. Please also add it to the client and try it out once (check that zypper repos are correct).

AWS and OpenStack are even easier to implement: the version variable is already there, it just needs documentation adjustments in their `README`s and `variables.tf` files, as the `host` module is the same regardless if it is a server, client or minion.

moio:
Rebased, added editorial fixes, support for clients (libvirt).

Support for other backends worked out of the box!

Merging.

PR #57, opened 2017-01-11
Fix tomcat 8 remote debug config

cbbayburt:
The following line in the config file prevents Tomcat from applying the configuration.
```
# This variable is used to figure out if config is loaded or not.
TOMCAT_CFG_LOADED="1"
```
This PR modifies the state file to comment-out the specified line.


moio:
Good catch @cbbayburt, thanks for taking the time to fix this!

PR #61, opened 2017-01-25
Instructions on how to run terraform from an arbitrary directory

Bischoff:
Instructions on how to run terraform from an arbitrary directory without getting an error. Feel free to edit for brevity and style.


moio:
Text is perfect, thanks for contributing! Merged.

PR #63, opened 2017-01-26
BUGFIX: use grains.get() instead of iterate empty dict.

MalloZup:


MalloZup:
@moio , rebased, next 64 pr contains the changes

PR #66, opened 2017-02-02
Fix id because it conflicts with something on suma

mseidl:
There is a bug just recently where calling the highstate fails because of a conflicting id.  @mc fixed an issue in SUSE Manager where he used salt-minion in a state.  
>>> output here

    rfsuma3pg:~ # salt rfminsles12sp1* state.highstate
    rfminsles12sp1.tf.local:
        Data failed to compile:
    ----------
        Detected conflicting IDs, SLS IDs need to be globally unique.
    include:
        The conflicting ID is 'salt-minion' and is found in SLS 'base:minion' and SLS 'base:services.salt-minion'
    ERROR: Minions returned with non-zero exit code

Renaming this from salt-minion to just minion makes the highstate apply successfully.

### DIFF
--- a/salt/minion/init.sls
+++ b/salt/minion/init.sls
@@ -2,11 +2,13 @@ include:
   - client.repos
   - minion.testsuite
 
-salt-minion:
+minion:
   pkg.installed:
+    - name: salt-minion
     - require:
       - sls: client.repos
   service.running:
+    - name: salt-minion
     - enable: True
     - watch:
       - pkg: salt-minion


salt/minion/init.sls
```diff
@@ -2,11 +2,13 @@ include:
   - client.repos
   - minion.testsuite
 
-salt-minion:
+minion:
```
mcalmer: I would prefix all ids with sumaform_ .

moio: That's a good proposal, but remember all salt files here can be used without terraform - idea is they can be copypasted manually and should not really depend on it.

Bottom line: I do not like the `sumaform_` prefix, but if anybody finds a better one this is a viable solution!

moio:
This should really be solved in SUMa by prefixing all of SUMa-shipped states, but I am accepting this PR now as a temporary workaround.

PR #70, opened 2017-02-12
configure extra repositories and extra packages via variable

mcalmer:
This allows to configure additional repositories to be used by simply defining them in main.tf as a variable.
(requires a newer version of terraform. Tested with version 0.8.6)

```
module "suma-headpg" {
  source = "./modules/libvirt/suse_manager"

  name = "suma-headpg"
  version = "head"
  memory = 4096
  extra_repos = [ "http://download.domain.top/my/devel/repository" ]
  // see modules/libvirt/suse_manager/variables.tf for possible values
}
```

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -193,3 +193,28 @@ module "suma31pg" {
 
 Note that the version is called `stable` in continuity with the existing naming schema because it will target the final release of SUSE Manager 3.1.
 However, at the time of writing, this version is still in alpha.
+
+### Add custom repos and packages
+
+You can specify custom repos and packages to be installed at deploy time for a specific host:
+
+```hcl
+module "minsles12sp1" {
+  source = "./modules/libvirt/minion"
+  base_configuration = "${module.base.configuration}"
+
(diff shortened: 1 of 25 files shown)


moio:
Wow, I was pretty convinced an upgrade on terraform would break many other things, as when I tested 0.8 betas experience was not completely smooth. Probably they have far better backwards compatibility now.

Thanks for this contribution, I will merge it as soon as we are done with [upgrading terraform](https://github.com/moio/sumaform/issues/71).

mcalmer:
@moio - what do you think? Where does these extra repos makes sense beside suse_manager ?

* client
* control_node
* host
* minion
* package_mirror
* postgress
* suse_manager_proxy

Or is it possible to put it in base and have it globally for all systems ?

mcalmer:
It now has also ```extra_pkgs = []```  and a fix for a nasty systemd bug when /etc/fstab is missing.

@moio  - would be good if you find some time to test and merge this. :-)

moio:
Thanks!!!!

PR #79, opened 2017-02-14
Adding timezone salt states

mseidl:
This adds a salt state to change the timezone.

The default timezone selected is CET, but you can change the timezone by adding a 
   timezone = "$desiredtimezone" 

In the module "base" section.


moio:
Thanks for this contribution! Will take a look [during hackweek](https://hackweek.suse.com/15/projects/1988).

PR #80, opened 2017-02-14
Fix: use `runas` instead of `user` for `cmd.run`

mbologna:
`user` option is deprecated for `cmd.run`

### DIFF
--- a/salt/suse-manager/init.sls
+++ b/salt/suse-manager/init.sls
@@ -73,12 +73,6 @@ suse-manager-setup:
   cmd.run:
     - name: /usr/lib/susemanager/bin/migration.sh -l /var/log/susemanager_setup.log -s
     - creates: /root/.MANAGER_SETUP_COMPLETE
-    {% if grains['saltversion'] == '2015.8.12' %}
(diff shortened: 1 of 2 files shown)


moio:
rfsuma3pg:~ # salt --version
salt 2015.8.12 (Beryllium)

https://docs.saltstack.com/en/2015.8/ref/states/all/salt.states.cmd.html#salt.states.cmd.run

No mention of runas here.

mbologna:
This applies only to `head` and `3.1-stable`

salt/suse-manager/init.sls
```diff
@@ -73,7 +73,7 @@ suse-manager-setup:
   cmd.run:
     - name: /usr/lib/susemanager/bin/migration.sh -l /var/log/susemanager_setup.log -s
     - creates: /root/.MANAGER_SETUP_COMPLETE
-    - user: root
+    - runas: root
```
moio: This will not work on all supported Salt versions.

moio:
Thanks for contributing!

PR #83, opened 2017-02-21
Require terraform 0.8

mbologna:


TROUBLESHOOTING.md
```diff
@@ -1,5 +1,9 @@
 # Troubleshooting
 
+## Q: I get the error "* Unknown root level key: terraform" whenever I try to `terraform apply`
```
moio: `terraform plan` is affected as well, probably others. Just mention the error message.

PR #89, opened 2017-02-23
add smt variable to suse_manager to configure a from mirror

mcalmer:
The light alternative to package-mirror: configure the URL of an SMT server to get the packages from internal.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -194,6 +194,21 @@ module "suma31pg" {
 Note that the version is called `stable` in continuity with the existing naming schema because it will target the final release of SUSE Manager 3.1.
 However, at the time of writing, this version is still in alpha.
 
+### SMT
+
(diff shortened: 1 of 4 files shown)


mcalmer:
@moio - would be good to get this merged. It would help also for our tests if we do not download everytime from the CDN

MalloZup:
@mcalmer , can we put maybe in the description the full-name of what is a SMT server? Just in sake of completeness. I had to google it, what was a smt server (shame on me :D )

moio:
>it would help also for our tests if we do not download everytime from the CDN

sumaform only downloads from download.suse.de or euklid.nue.suse.com, so this should not really make a difference.

Is there anything I am missing?

mcalmer:
When you setup channels 

  mgr-sync add channel

they are using the CDN and download from updates.suse.com/....
The only way to prevent this is using package-mirror which is is overkill when you are in the SUSE Network.

moio:
You are right, I am wrong. Merged.

PR #90, opened 2017-03-01
BUGFIX: fix phantomjs issue with secure websocket

MalloZup:
We need to have the suma-server certificate in order to
visualize the button tristate on phantomjs.


moio:
Thanks for hunting this down and fixing it!

PR #92, opened 2017-03-05
install galaxy GPG key

mcalmer:
This deploys and import the GALAXY GPG key in not SUSE systems.
RedHat requires to have the keys installed. This provide the possibility to install a new salt version
from our galaxy repos.


MalloZup:
ok we need this for centos testhost machines

PR #93, opened 2017-03-07
WIP: Minion-centos7: install latest salt pkg

MalloZup:
In addition created also a centos dir for putting togheter the repos of centos os


salt/centos/repos.sls
```diff
@@ -0,0 +1,24 @@
...
+
+trust_galaxy_key:
+  cmd.wait:
+    - name: rpm --import /tmp/galaxy.key
+    - watch:
+      - file: deploy_galaxy_key
+
+refresh-up2date-repos:
+  cmd.run:
+    - name: yum -y update
```
MalloZup: here maybe we should make an equivalent of zypper refresh

PR #95, opened 2017-03-07
update salt and salt-minion - yum does not take dependencies into acc…

mcalmer:
…ount

yum somehow only update salt but not salt-minion. This change take care that both packages are 
up to date.


moio:
Thanks for taking care of this!

PR #98, opened 2017-03-17
Create Image on demand

MalloZup:
fix issue  https://github.com/moio/sumaform/issues/26.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -302,3 +302,21 @@ module "evil-minions" {
   // see modules/libvirt/evil_minions/variables.tf for possible values
 }
 ```
+## Use specific images on demand
+
+You can select which image, disks you want to use with sumaform.
+This is specially usefull for sparing time and space. By default we build all images.
+For the images we build look at "modules/libvirt/base/variables.tf" to see all images.
+Here an example:
+
+```hcl
+module "base" {
+  source = "./modules/libvirt/base"
+
+  cc_username = "UC7"
+  cc_password = ...
+  images = ["centos7", "sles12sp2"] 
+  }
+
+}
+```
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -302,21 +302,3 @@ module "evil-minions" {
   // see modules/libvirt/evil_minions/variables.tf for possible values
 }
 ```
-## Use specific images on demand
-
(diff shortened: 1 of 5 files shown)


moio:
Hey @MalloZup, thanks for this contribution!

I am not sure to merge it in this form though. It is already an enhancement over current situation, but I am uncertain if we will want to change the "API" in the future, breaking people's `main.tf`s.

Would it be possible/better to optionally specify a list of images, instead of a simple boolean flag?

Eg. if nothing is specified, all images are built. Otherwise, only the specified ones are.

That would not break compatibility, work in typical cases but still allow flexibility for those who know what they need in advance.

MalloZup:
@moio ok i will look for the list

MalloZup:
@meaksh ,@moio, what do you think?
We have 2 lists now ( the Pr is just a prof)

moio:
It looks better, I would still prefer if I could specify a list as a variable in `base`. Eg.

```hcl
module "base" {
  source = "./modules/libvirt/base"

  cc_username = "UC7"
  cc_password = ...

  images = ["centos8", "centos9"] # because you know we live in future
}
```

Would this be possible?

MalloZup:
@moio, @meaksh,  so i adapted the logic.

We can do now:
 ```hcl

  images =  { 
           centos7 = "http://w3.nue.suse.com/~smoioli/sumaform-images/centos7_v2.qcow2"
           sles12sp2 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp2.x86_64.qcow2" 
  }
```
instead of
```hcl
 images = ["centos8", "centos9"]
```
The last one would require an extra internal map variable, where we define the source of the qcow.

With the first approach we are more dynamic, and an advanced user could specify other qcow on the fly if he wish.

moio:
Hmm, I am not sure about this. True it adds flexibility, but the number of adjustments to use a different image is significant - and this comes at the expense of a harder-to-write `main.tf`. With an internal variable, if an advanced user wanted to swap a qcow, all he/she needs to know is the location of `base/main.tf` and edit that directly.

I would vote to go in that direction, sorry. If you don't have time I can take care of that.

Thanks!!!

moio:
tumblr_inline_odd2x9pvyg1raprkq_400

Merged (with some fixes, please take a look at them).

Great work!

PR #105, opened 2017-04-12
Feat: add RES7 mirror (default disabled)

mbologna:


salt/mirror/mirror.lftp
```diff
@@ -58,6 +58,10 @@ mirror --continue --delete --loop --parallel=3 --verbose=3 -x "src/.*" -x "nosrc
 mirror --continue --delete --loop --parallel=3 --verbose=3 mirror/SuSE/build.suse.de/SUSE/Products/SLE-Module-Containers/12/x86_64/product/ mirror/SuSE/build.suse.de/SUSE/Products/SLE-Module-Containers/12/x86_64/product/
 mirror --continue --delete --loop --parallel=3 --verbose=3 -x "src/.*" -x "nosrc/.*" -x ".*\.drpm" -x "update_debug/.*" mirror/SuSE/build.suse.de/SUSE/Updates/SLE-Module-Containers/12/x86_64/ mirror/SuSE/build.suse.de/SUSE/Updates/SLE-Module-Containers/12/x86_64/
 
+# RES 7
+# open http://smt-scc.nue.suse.com
```
moio: Not sure about using this instance by default, I would prefer pulling from official servers. We are aware of slowness right now (was also reported by others), but this server is for SMT development so we should not mess with it.

PR #112, opened 2017-05-11
512 MB are not enough for a proxy + accomodating repository changes

Bischoff:
With 512 MB, oom-killer keeps killing squid or httpd processes


moio:
Looks nice, accepting right away! Thanks!

PR #118, opened 2017-05-22
Adding libvirt support for evil-minions

meaksh:
This PR enables libvirt support for [evil-minions](https://github.com/moio/evil-minions).

`evil-minions` is a Salt load generator that can be used for testing performance of Salt & SUSE Manager with an insane load from evil minions.


modules/libvirt/evil_minions/main.tf
```diff
@@ -0,0 +1,27 @@
...
+  count = "${var.count}"
+  memory = "${var.memory}"
+  vcpu = "${var.vcpu}"
+  running = "${var.running}"
+  mac = "${var.mac}"
+  additional_repos = "${var.additional_repos}"
+  additional_packages = "${var.additional_packages}"
+  grains = <<EOF
+
+version: 3.1-released
```
moio: Since we (presently) support it only on 3.1-released, we could also fix the base OS image to SP2. Only in case we find out there's a reason to have different combinations I would add parameters :angel:

modules/libvirt/evil_minions/variables.tf
```diff
@@ -0,0 +1,64 @@
...
+  type = "map"
+}
+
+variable "minion_count"  {
+  description = "number of minions that this swarm will spawn"
+  default = 200
+}
+
+variable "start_delay"  {
+  description = "number of seconds to wait between spawns"
```
moio: Not needed, please remove

modules/libvirt/evil_minions/variables.tf
```diff
@@ -0,0 +1,64 @@
...
+  default = 6
+}
+
+variable "additional_repos" {
+  description = "extra repositories in the form {label = url}, see README_ADVANCED.md"
```
moio: Not needed (AFAIK), please remove

modules/libvirt/evil_minions/variables.tf
```diff
@@ -0,0 +1,64 @@
...
+  default = {}
+}
+
+variable "additional_packages" {
+  description = "extra packages to install, see README_ADVANCED.md"
```
moio: Not needed (AFAIK), please remove

salt/evil_minions/init.sls
```diff
@@ -0,0 +1,24 @@
+include:
+  - evil_minions.repos
+
+master_configuration:
+  file.managed:
```
moio: I do not think this is needed - if it is, we should base `evil_minions` on top of `minion` instead of `host`. Please choose one of the two.

modules/libvirt/evil_minions/main.tf
meaksh: Ok

modules/libvirt/evil_minions/variables.tf
meaksh: Ok

modules/libvirt/evil_minions/variables.tf
meaksh: Ok

modules/libvirt/evil_minions/variables.tf
meaksh: Ok

salt/evil_minions/init.sls
meaksh: This would be needed in case we want to execute `dumping-salt-minion` on the host, but after conversation, we will run `evil-minions` just after terraform deployment using the parameters supplied, so it does make sense anymore to have a `master configuration` state here.

In case you want to generate a yaml dump file, I would say you could manually run `dumping-salt-minion` on whatever minion you had deployed.

README_ADVANCED.md
```diff
@@ -278,3 +278,19 @@ module "grafana" {
 ```
 
 Grafana is accessible at http://grafana.tf.local with username and password `admin`.
+
+
+## Evil Minions load generator
+
+You can deploy an [Evil Minions](https://github.com/moio/evil-minions) host in order to test load performance in your SUSE Manager server. A libvirt example would be:
```
moio: Nitpick: name of the project is `evil-minions` (not capitalized), please use that.

modules/libvirt/evil_minions/variables.tf
```diff
@@ -0,0 +1,60 @@
...
+  description = "use ${module.<SERVER_NAME>.configuration}, see main.tf.libvirt.example"
+  type = "map"
+}
+
+variable "minion_count"  {
+  description = "number of minions that this swarm will spawn"
+  default = 200
+}
+
+variable "minion_pool" {
```
moio: Nitpick: I would rather not add this, and have it being automatically detected via CPU count. If one wants more, he can just add vCPUs :smile:

modules/libvirt/evil_minions/variables.tf
```diff
@@ -0,0 +1,60 @@
...
+  description = "number of evil-minions processes"
+  default = 2
+}
+
+variable "minion_dump_yml_file" {
+  description = "Salt minion YaML dump file"
```
moio: Please use the `file` terraform interpolation function so that users put a path in here, relative to `main.tf`, and that file's contents are put into a grain.

https://www.terraform.io/docs/configuration/interpolation.html#file-path-

modules/libvirt/evil_minions/variables.tf
meaksh: Yeah! That's much better.

meaksh:
@moio - I think your latest feedback is addressed here now. Only missing thing is allowing different dump files via configuration. I got some errors with `file()` when the file doesn't exist, even if I put that logic inside an if statement.. I guess I'm missing something 😄 

As that missing part is somehow optional, I would say this is ready to be merged (unless more fixes are needed, ofc)

moio:
> I got some errors with file() when the file doesn't exist, even if I put that logic inside an if statement.. I guess I'm missing something

You are missing a known Terraform issue :smile:! https://github.com/hashicorp/terraform/issues/12194

One way to work around this is to point to `salt/evil_minions/minion_dump.yml` by default: we can assume that always exists.

If user specifies a non-existent file on purpose, we would error out which is the desired behavior.

meaksh:
@moio - I made `minion_dump_yaml_file` a configuration parameter. Now, we can set whatever external yaml file to be provided as the minion dump file for `evil-minions`.

Data is passed as grains encoded as base64 to avoid breaking grains yaml structure due `---` marks on the dump file.

modules/libvirt/evil_minions/variables.tf
```diff
@@ -0,0 +1,55 @@
...
+  type = "string"
+}
+
+variable "server_configuration" {
+  description = "use ${module.<SERVER_NAME>.configuration}, see main.tf.libvirt.example"
+  type = "map"
+}
+
+variable "minion_count"  {
+  description = "number of minions that this swarm will spawn"
```
moio: Nitpick: it's not a swarm anymore :smile: - perhaps we could rename this to `evil_minion_count`?

modules/libvirt/evil_minions/variables.tf
```diff
@@ -0,0 +1,55 @@
...
+
+variable "minion_dump_yml_file" {
+  description = "path to salt minion yaml dump file"
+  type = "string"
+  default = "salt/evil_minions/minion-dump-default.yml"
+}
+
+variable "slowdown_factor"  {
+  description = "slowdown factor for evil-minions"
+  default = 1.0
```
moio: Please make it 0 by default, which is the same as in `evil-minions`

modules/libvirt/evil_minions/variables.tf
meaksh: That's true! Copy/paste collateral effects 😄

modules/libvirt/evil_minions/variables.tf
meaksh: Ok

PR #119, opened 2017-05-23
Updated and reworked information about versions

Bischoff:
* latest naming conventions for versions
* more information about versions
* moved to top


moio:
Merged after some minimal editorial change. Thank you very much!

PR #120, opened 2017-05-23
Ensuring zypper is up to date like salt

Bischoff:


salt/default/init.sls
```diff
@@ -2,11 +2,13 @@ include:
   - default.repos
   - default.pkgs
 
-up_to_date_salt:
+up_to_date_salt_and_zypper:
   pkg.latest:
     - pkgs:
       - salt
       - salt-minion
+      - zypper
```
moio: This should at least be guarded by a check for SUSE OSs ([example](https://github.com/moio/sumaform/blob/master/salt/default/repos.sls#L1)) or it will break in CentOS.

Bischoff:
Requested changes done (merely a copy and paste from a few lines below)

Bischoff:
Nah, it's not nitpicking, it's a legitimate question.

I prefer the generic ``up_to_date_pkgs``. Doing the change right now.

Bischoff:
Requested changes done.

meaksh:
@Bischoff - Thanks! Could you please rebase this PR? I'll merge it then.

Bischoff:
I'm not sure to understand what you ask. I already made sure it was holding in only one commit, if what's you mean.

meaksh:
Sorry, I mean doing:
```
git pull --rebase REMOTE_UPSTREAM_SUMAFORM/master
```

That way you'll update your branch to include latest commits that has been added to `master` after you created your branch. Then your all your commits will be placed on the top and we prevent to get a mess of forking and merging branches, they would look cleaner. 

But I just realized that github PR "merge" button now allows you to "rebase & merge", so you can just skip to manual rebasing, I'll do it.

Sorry for the noise and thanks again for the fix! 😄

moio:
@meaksh: JFI: I like the behaviour of the "Rebase and Merge" button when only one commit is involved (it's equivalent to cherry-picking, with the advantage that GitHub will track the PR it's coming from in its metadata).

I also like the behaviour of "Squash and Merge" if squashing makes sense for the PR.

My problem starts when you have several commits, not on top of `master`, and for which squashing does not make sense. Presently AFAIK manual intervention is the only option.

My recommendation in the table below:

| Number of commits | On top of `master`? | Squashing needed? | Suggested Action                          |
|-------------------|-------------------|-------------------|-------------------------------------------|
| One               | Yes               | No                | Click Rebase and Merge (Merge is also OK) |
| One               | No                | No                | Click Rebase and Merge                    |
| Several           | Yes               | Yes               | Click Squash and Merge                    |
| Several           | No                | Yes               | Click Squash and Merge                    |
| Several           | Yes               | No                | Click Merge                               |
| Several           | No                | No                | Rebase manually, then click Merge         |

meaksh:
Yeah, totally agree with your recommendations. Thanks! 👍

PR #121, opened 2017-05-23
Adding support to run "evil-minions" on AWS

meaksh:
This PR adds support to deploy `evil-minion` on Amazon Web Services:

- Provides a vendor version of `evil-minion`, as aws instances cannot access github.
- Makes github installation of `evil-minion` as configurable for libvirt.
- Fixes an error on `configure_aws_tunnels.rb`
- Adds sumaform aws module for evil-minions.

Please @moio , take a look here. Thanks!


modules/aws/evil_minions/variables.tf
```diff
@@ -0,0 +1,100 @@
...
+}
+
+variable "timezone" {
+  description = "Timezone setting for this VM"
+  default = "Europe/Berlin"
+}
+
+variable "evil_minion_count"  {
+  description = "number of minions that will be spawned by evil-minions"
+  default = 200
```
moio: Given today's discussions this could be bumped up to 500 safely right?

PR #125, opened 2017-06-03
there is no 42.1 builded

MalloZup:


MalloZup:
http://download.opensuse.org/repositories/home:/SilvioMoioli:/tools/

PR #129, opened 2017-06-19
Installing "salt-ssh" cli on suma server (for testsuite only)

meaksh:
This PR installs `salt-ssh` CLI on the SUMA server (for testsuite only) since it's now required by `spacewalk-testsuite-base` to check rendered pillar data for ssh minions.


moio:
Thanks!

PR #130, opened 2017-06-19
[WIP] [Experimental]: try to fix race issues on volumes

MalloZup:
- move volume resource to  host module
- change logic on base, no resource there
- the dependencies are changed, added depend_on "main_disk". (see graph comparasion for that)

Contra atm:

- How can we pass array with an Output-variable and iterate overthat? ( for customize image_on_demand)

@moio see also the graph with the new design, ( we avoid one extra-parellism)
 . Do you know if we can pass list to output variables? (if it is possible)

Atm, i'm not sure if it works really, since i'm in holiday, i will just look at it from time to time, but i wanted to share the approach( i think moving the volume resource to host is a good approach)

I did few tests it worked sofar

- [ ] need testing on that


moio:
Problem with this approach is that each host will get its image(s) - they will be duplicated many times.

Unless I am missing something.

MalloZup:
@moio ok you are right, ok i close this 🗻

PR #131, opened 2017-06-20
disable gpgcheck for RES repos. GPG key is not installed on our image

mcalmer:
sumaform do not setup centos correct. It failed to update salt-minion package to new version because the
GPG key is not installed on the image.

Disable GPG check for these repos.


mbologna:
Superseded by https://github.com/moio/sumaform/pull/132

PR #133, opened 2017-06-22
Change salt 'file_root' for sumaform to /opt/sumaform/srv/salt

meaksh:
This PR applies the new `terraform-post-resource` state at the end of a resource deployment for libvirt and aws modules in order to manage cleanup tasks:

- Remove all files from /srv/salt/
- Remove Salt package from traditional clients

I'm using `at` to trigger the cleanups because otherwise Salt would fail during `terraform-post-resource` execution as jinja template and salt itself would disappear at execution time.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -187,3 +187,7 @@ A simple solution is to create a symbolic link pointing to the `salt` directory 
 The error means that there is no such version as "3-stable" for the desired resource. Specifically, all SUSE Manager 3.0 versions have been renamed to get a `3.0-` prefix instead of the old `3-` one to avoid confusion with 3.1, moreover `stable` has been replaced by `released` to take into account beta versions.
 
 In this specific example, use `3.0-released` instead of `3-stable`. In other cases, please check the `variables.tf` file of the module you want to use.
+
+## Q: how do I re-apply the Salt state that was used to provision the machine?
+
+Run `salt-call --local --file-root=/root/salt state.highstate`.
--- a/modules/aws/evil_minions/main.tf
+++ b/modules/aws/evil_minions/main.tf
@@ -33,7 +33,7 @@ resource "null_resource" "host_salt_configuration" {
 
   provisioner "file" {
     source = "salt"
-    destination = "/srv/salt"
+    destination = "/root"
   }
 
   provisioner "file" {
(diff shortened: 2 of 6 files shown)


MalloZup:
We should create a variable, salt cleanup or wtever set by default to false. Atm if we.merge this, we break a lot of tests on the testsuite( removing pkg).

meaksh:
I agree with @MalloZup that a variable might be useful here, but I would suggest in that case, to set that variable by default to `True` and, if necessary, to change the testsuite setting it to `False`.

FMPOV when I deploy a **traditional client** (not minion) using sumaform, by default, it should not contains any Salt packages after deployment. What do you think?

@MalloZup - I only remove Salt package for **traditional clients** (which are not supposed to have it). Why should it be a problem on the testsuite?

moio:
@MalloZup I agree with @meaksh this should be in by default, and am willing to merge the PR even without a variable. That can come later, if needed.

@meaksh I do not agree about emptying `/srv/salt`. That directory is useful if one wants to re-apply the sumaform state for any reason, and I frequently do that (via `salt-call --local state.highstate`).

Now the real problem are conflicting ids and conflicting `top.sls`. Former should be solved on the Manager side, we had a long discussion that all ids coming from Manager should be prefixed, but this is not coming any time soon. Moreover, I now see we have no reasonable solution to handle the sumaform-specific `top.sls` (correct me if I am wrong).

I would then suggest to go back to the approach of changing the file root for sumaform, so that deployment is done via `salt-call --local --file-root=/opt/sumaform/salt state.highstate` or something similar.

About the removal of the salt package: are we sure we want that? Can you explain again why having the package around hurts?

MalloZup:
@meaksh for traditonal-client is ok for removing pkg, i missed the "traditional" part.

The only problem we can have with this is only , when do a migration test ( from trad-client to minion), but imho is ok

meaksh:
@moio - Exactly, apart from the possible conflicting ids of states, the problem is more about `/srv/salt/top.sls`, which in the case of a suma server, will inject the sumaform states for the all the minions.

I agree with your proposal of changing the `file_root` for the sumaform states. That way we preserve the functionality you mentioned of re-applying sumaform highstate. I'll address it now.

About the `salt-minion` package on traditional clients: I have experienced 2 different behaviors during migration from traditional to minion in SUMA because the `salt-minion` package was already installed on the traditional client before migrating it. That's why I suggest to have the environment for a sumaform traditional client as closer as possible to a customer's traditional client.

meaksh:
After discuss with moio on IRC, we have decided to change this PR to only modify the Salt `file_root` for sumaform deployment.

In order to prevent from having "special" machines where `salt` is not installed and sumaform highstate cannot be re-applied, we've decided that a good point would be to use `salt-ssh` for provisioning instead.

New issue created: https://github.com/moio/sumaform/issues/134 😄 

BTW I've rebased and fixed the conflict. Please @moio check it. Thanks

meaksh:
Ha! 😄 Sure, no problem!

moio:
Merged! Thanks a lot

PR #137, opened 2017-06-23
Feat: add support for traceback mail addr override

mbologna:


README_ADVANCED.md
```diff
@@ -335,3 +334,18 @@ module "suma3pg" {
   from_email = "root@mbologna.openvpn.suse.de"
 }
 ```
+
+Internal Server Error and relative stacktrace are sent via e-mail by default to `galaxy-noise@suse.de`.
```
moio: Grammar nitpick: error*s*, stacktrace*s*

modules/libvirt/suse_manager/variables.tf
```diff
@@ -92,3 +92,8 @@ variable "mac" {
   description = "a MAC address in the form AA:BB:CC:11:22:22"
   default = ""
 }
+
+variable "traceback_email" {
+  description = "email address used in SUMa to send ISE stacktraces"
```
moio: * [The abbreviation SUMa should never appear in user-facing material](https://github.com/moio/sumaform/pull/136#discussion_r124197687). Use "SUSE Manager"
* ISE stacktraces are not the only email you get with this parameter, please use a more generic description

salt/suse_manager_server/init.sls
```diff
@@ -107,3 +107,13 @@ substitute_email_sender_address:
     - require:
         - cmd: suse_manager_setup
 {% endif %}
+
+{% if grains['traceback_email'] %}
```
moio: Please use `grains.get(key)` for the state to work correctly in case the grain is not set. We had a similar issue fixed here:

https://github.com/moio/sumaform/commit/47a357715897ec6c1fb22ed153316057ec45a15a#diff-35879e2767a8cba0cce7f156e2a9de24R128

I am addressing this directly.

PR #138, opened 2017-06-27
Fix: rename host in testsuite example

mbologna:
As per convention, we name `sumaheadpg` hosts running `head` version of SUSE Manager, whilst `suma3pg` the corresponding `3.0-released` version.

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -15,10 +15,10 @@ module "base" {
   images = ["centos7", "sles12sp1", "sles12sp2"]
 }
 
-module "suma3pg" {
+module "sumaheadpg" {
   source = "./modules/libvirt/suse_manager"
   base_configuration = "${module.base.configuration}"
-  name = "suma3pg"
+  name = "sumaheadpg"
   image = "sles12sp2"
   version = "head"
   for_development_only = false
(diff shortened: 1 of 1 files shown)


moio:
I like the spirit of this PR, but would prefer to use a stable version for the example file, like 3.0 or 3.1. This comes from the old times when head was still pointing to 3.0 betas!

Remember the example file is what people use in sumaform the very first time they try it. I would like to ensure the experience is smooth :smile:

moio:
@MalloZup would it be possible/would it make sense to use `3.1-nightly` or even `3.1-released` as the default for the testsuite?

MalloZup:
@moio, in the past for the example we used always head because this is the version that we are looking for. Everything released should go to QAM, but we lacks on this people. 
Why you want to point to 3.1? Consider that if it is for stabity, the nightly get broken too

moio:
In theory, developers should make sure their fixes do not break the testsuite or add tests whenever they add features. There is no rule whether to start working on 3.1 and then forward-porting to HEAD, or to start working on HEAD and backport to 3.1 - both are valid approaches.

It's OK, let's keep HEAD for now.

MalloZup:
Yes, sure. Sofar we don't run the acc testsuite on each PR, we will not have a complete CI. I think now we could build a Pipeline with Concourse and Sumaform (since the terraform race condition is fixed), and build the whole needed pkg in other vms. We need simply to bypass OBS. ( at least we need only to build and create repo) 🥇

PR #142, opened 2017-06-29
Fix: make salt-ssh hosts really vanilla hosts

mbologna:
If an host is marked as salt-ssh bootstrapped, we should use its vanilla option.
This needs testing.
@MalloZup @Bischoff

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -47,7 +47,7 @@ module "head-min-sles12sp2" {
 }
 
 module "head-minssh-sles12sp2" {
-  source = "./modules/libvirt/minion"
+  source = "./modules/libvirt/host"
   base_configuration = "${module.base.configuration}"
   version = "head"
   name = "head-minssh-sles12sp2"
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -23,6 +23,7 @@ module "head-srv" {
   version = "head"
   for_development_only = false
   for_testsuite_only = true
+  ssh_key_path = "./salt/controller/id_rsa.pub"
 }
 
 module "head-cli-sles12sp2" {
@@ -33,6 +34,7 @@ module "head-cli-sles12sp2" {
   server_configuration =  { hostname = "head-srv.tf.local" }
   for_development_only = false
   for_testsuite_only = true
+  ssh_key_path = "./salt/controller/id_rsa.pub"
 }
 
 module "head-min-sles12sp2" {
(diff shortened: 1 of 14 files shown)


moio:
Marking this as WIP until @MalloZup or @Bischoff confirm this. Implications are non trivial and I have a feeling this *could* break the testsuite.

MalloZup:
@moio @mbologna  since we want to share the knowledge and responsability for testsuite, i want to show to @mbologna how to do it. Imho it should work the fix, but let see dd

mbologna:
@MalloZup as discussed please make comments on this PR, thanks.

MalloZup:
- [ ] create branch for testing sumaform 
- [ ] learn sumaform controller( see galaxi gitlab) and run with sumaform custom branch)
 Use test jenkins job
- [ ] merge this if test ok

MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base/blob/27a095faaf85ad4e1f38ef758d89acab4a8e3bd2/features/salt_ssh.feature

MalloZup:
@mbologna run the testsuite locally ,.it will.not break, but just for a matter of discipline.

MalloZup:
@mbologna  zio camper 🏕 ⛺️ , rebase your PR, and if you just do ```terraform apply``` you will see that this pr doesn't work. vanilla has not devel and so on parameter

MalloZup:
so after discussion and test, we find out that the testsuite is coded to use clients with rsa-key.

We need to implement in support, a callback function, that if the twopence var is empty (because there is no rsa-key), we use `net-ssh` with user and pwd

- [ ] create PR on spacewalk-testsutite with this

mbologna:
STILL WIP.
We need to support an option `ssh_key_path` on a per-host basis. Something is already done in this branch, but needs fixing. I committed it as WIP, in case somebody wants to take a look.
I will leave it here and resume after vacation.

MalloZup:
@mbologna,   i will busy also with other stuff next sprint, enjoy your vacation.

mbologna:
PR is now complete. Reviewers welcome.

MalloZup:
before merging i would remind that if we merge, this go directly to production. (30,31,HEAD)

@mbologna  can you resume what this PR do?
thx

MalloZup:
@moio, @mbologna 
you need always to update the production, before merging when we do a change like adding variables etc etc.

https://gitlab.suse.de/galaxy/sumaform-test-runner

moio:
@MalloZup the choice of using `master` as the base ref for the automated testsuite was not mine - another manually-advanced ref or branch could have been used instead. Each choice has pros and cons but this is not the place to discuss them.

What I can say is that I am definitely not taking responsibility for the testsuite in the context of sumaform, I simply lack the time to check before each merge, sorry.

PR #144, opened 2017-06-30
Feat: autoselect testsuite branch

mbologna:
Based on the version of the server we are going to launch the testsuite
against, we decide automatically which branch of the testsuite to
checkout.
The association is done in the map defined in:

`modules/libvirt/controller/main.tf`

@MalloZup @Bischoff

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -72,7 +72,6 @@ module "controller" {
   source = "./modules/libvirt/controller"
   base_configuration = "${module.base.configuration}"
   name = "controller"
-  branch = "master"
   server_configuration = "${module.sumaheadpg.configuration}"
   client_configuration = "${module.clisles12sp2.configuration}"
   minion_configuration = "${module.minsles12sp2.configuration}"
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -77,7 +77,5 @@ module "controller" {
   minion_configuration = "${module.minsles12sp2.configuration}"
   centos_configuration = "${module.mincentos7.configuration}"
   minionssh_configuration = "${module.minsles12sp1ssh.configuration}"
-  # branch = "master"
-  # There is a default selection if nothing is specified.
-  # See modules/libvirt/controller/variables.tf for possible values
+  # branch = "master" # See modules/libvirt/controller/variables.tf for possible values
 }
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -77,4 +77,7 @@ module "controller" {
   minion_configuration = "${module.minsles12sp2.configuration}"
   centos_configuration = "${module.mincentos7.configuration}"
   minionssh_configuration = "${module.minsles12sp1ssh.configuration}"
+  # branch = "master"
+  # There is a default selection if nothing is specified.
+  # See modules/libvirt/controller/variables.tf for possible values
 }
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -67,13 +67,12 @@ cucumber_requisites:
       - sls: controller.repos
 
 cucumber_testsuite:
-  cmd.run:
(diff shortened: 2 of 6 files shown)


salt/controller/init.sls
```diff
@@ -67,13 +67,12 @@ cucumber_requisites:
...
 
 cucumber_testsuite:
-  cmd.run:
-    # HACK: work around the lack of skip_verify and enforce_toplevel in archive.extracted before salt 2016.11
-    - name: |
-        wget -P /root/ https://github.com/SUSE/spacewalk-testsuite-base/archive/{{ grains.get("branch" )}}.zip
-        unzip /root/{{ grains.get("branch" )}}.zip -d /root/
-        mv /root/spacewalk-testsuite-base-{{ grains.get("branch" )}} /root/spacewalk-testsuite-base
-    - creates: /root/spacewalk-testsuite-base
+  git.latest:
```
MalloZup: @mbologna i'm not sure if git.latest is the solution.

I remember that we have massively problems in the past using the git state ( like the download was never terminating, and failing)

So that was the reason that we used the tarball.zip, because with git, ( i was using the state) , was causing a lot of problems

main.tf.libvirt-testsuite.example
```diff
@@ -72,7 +72,6 @@ module "controller" {
   source = "./modules/libvirt/controller"
   base_configuration = "${module.base.configuration}"
   name = "controller"
-  branch = "master"
```
MalloZup: @mbologna i tend to like this autoselect feature, one stuff, is that, can we add custom branches like mallozup:space-devel49, ?

In the past we could add custom branches, and we used also this in production with the **test** jobs.
So we could run arbitrary testsuite branches, think wenn you want to test your branch

MalloZup: https://gitlab.suse.de/galaxy/sumaform-test-runner/blob/master/test/main-full.tf#L63

salt/controller/init.sls
MalloZup: just look on the first commit, you will see the history with the state
(ciao bell)

moio: @MalloZup there could be a difference now that we use salt 2016.11 - I would assume the bug we ran into has been closed in the meantime. Please test this.

MalloZup: @moio, we have a jenkins job, that is called Test. This doesn run testsuite in prod by it is like that. What if we create  staging branch in sumaform, where we merge all change that we want to test. And this Job will run on that.

moio: I don't recommend keeping a separate branch with cherry-picked (or merged) commits, as this will diverge with time. I would be OK with a branch which is only a fast-forward away from master and nothing more. @mbologna feel free to create such a branch (eg. `testsuite-master`).

main.tf.libvirt-testsuite.example
mbologna: Added option to specify optional branch in latest commit.

main.tf.libvirt-testsuite.example
```diff
@@ -72,10 +72,12 @@ module "controller" {
...
   server_configuration = "${module.sumaheadpg.configuration}"
   client_configuration = "${module.clisles12sp2.configuration}"
   minion_configuration = "${module.minsles12sp2.configuration}"
   centos_configuration = "${module.mincentos7.configuration}"
   minionssh_configuration = "${module.minsles12sp1ssh.configuration}"
+  # branch = "master"
```
moio: Please remove these two lines from here (example file should be as simple as possible) and document the optional branch switching in https://github.com/moio/sumaform/blob/master/README_ADVANCED.md#cucumber-testsuite

main.tf.libvirt-testsuite.example
MalloZup: Ok boss

MalloZup:
Nice pr! Hcl  fitness for @mbologna . I think it was not implictely needed but (since the real.problem.is that documentation is oft bypassed ) but it is confortable.

mantel:
Looks good to me (did not test it, but I assume it works :)

PR #147, opened 2017-07-07
Reflect production prefix names

MalloZup:
We need a 1:1 production, so people learn right prefixes names

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -10,68 +10,68 @@ module "base" {
   // pool = "default"
   // network_name = "default" // change to "" if you change bridge below
   // bridge = ""
-  name_prefix = "head-"
+  // name_prefix = ""
   // you need always sles12sp1 because controller is based on that.
   images = ["centos7", "sles12sp1", "sles12sp2"]
 }
 
-module "srv" {
+module "head-srv" {
   source = "./modules/libvirt/suse_manager"
   base_configuration = "${module.base.configuration}"
-  name = "srv"
+  name = "head-srv"
   image = "sles12sp2"
   version = "head"
   for_development_only = false
   for_testsuite_only = true
 }
 
-module "cli-sles12sp2" {
+module "head-cli-sles12sp2" {
   source = "./modules/libvirt/client"
   base_configuration = "${module.base.configuration}"
-  name = "cli-sles12sp2"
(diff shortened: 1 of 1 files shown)


moio:
Too bad this is inconsistent with all other defaults in `sumaform` - why were different names chosen? Eg. I expect most developers to be familiar with `minsles12sp2`.

MalloZup:
we took a decision according to this:
https://github.com/SUSE/spacewalk/wiki/Cucumber-Sumaform-Infrastracture#new-naming-scheme

moio:
Then why `srv` and not `suma-head-srv`?

MalloZup:
"suma-head" is the prefix in base module.
it can be then
"dma-srv"

moio:
From my perspective the prefix serves a different purpose, as you explicitly configure `srv` to have `version = "head"`

MalloZup:
@moio  i was not the only one that took the decision. Maybe we misbehaved the Prefix, but the goal was to have
 "suma" (production prefix)
  -head ( version)
  -srv (role)
etc. 
At the time it was more like the highlevel and we overlooked the difference vs sumaform.

Imho now t will take time to change the names on prod

moio:
> @moio i was not the only one that took the decision. [...]

This is irrelevant to my last comment. I am saying that `srv` is wrong, should be `suma-head-srv` or at least `head-srv` as it is specified in the example file that the version is `head`.

MalloZup:
ok i got what you mean, thx

MalloZup:
@moio ok now i added the prefix "head".

thx!

moio:
Are you OK with the [latest change](1deca67a3ab767b1b1ed9a1131d2f4cb762406c3) I'm proposing here?

MalloZup:
sure

@moio thx!

main.tf.libvirt-testsuite.example
```diff
@@ -10,68 +10,68 @@ module "base" {
...
   // you need always sles12sp1 because controller is based on that.
   images = ["centos7", "sles12sp1", "sles12sp2"]
 }
 
-module "srv" {
+module "head-srv" {
   source = "./modules/libvirt/suse_manager"
   base_configuration = "${module.base.configuration}"
-  name = "srv"
+  name = "head-srv"
```
MalloZup: ok

PR #153, opened 2017-07-19
TOMCAT_CFG_LOADED flag only for 3.1 version or later

ncounter:
TOMCAT_CFG_LOADED is not present for older version `[2.1, 3.0]`.

### DIFF
--- a/salt/suse_manager_server/tomcat.sls
+++ b/salt/suse_manager_server/tomcat.sls
@@ -65,7 +65,7 @@ tomcat_config:
     - require:
       - sls: suse_manager_server.rhn
 
-{% if '2.1' not in grains['version'] %}
+{% if '2.1' not in grains['version'] and '3.0' not in grains['version'] %}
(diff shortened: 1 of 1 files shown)


moio:
Please take a look at ba578f4fd98d6f6c8536c664b08a39f24c36e44e, I introduced this exactly because 3.0 broke at some point. Could it be that the tomcat version is different between 3.0 stable and nightly for example?

ncounter:
Ha! That would be a nice catch and interesting question.

Actually I have just tainted and re-downloaded the sles12sp1 image `suma 3.0` is based on, I built two `SUSE Manager server` (one based on `3.0-released` and one on `3.0-nightly`). They have the same tomcat version `tomcat-8.0.43-10.19.1.noarch`, and both fail in case the `tomcat_config_loaded` state runs.

moio:
I guess Tomcat was updated in SLES then. Merging, thanks!

PR #154, opened 2017-07-27
Fix server hostname

mbologna:
Without this patch, resources are created independently (e.g. clients
are created without waiting for the server to be created).

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -30,7 +30,7 @@ module "head-cli-sles12sp2" {
   base_configuration = "${module.base.configuration}"
   name = "head-cli-sles12sp2"
   image = "sles12sp2"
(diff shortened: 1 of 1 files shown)


MalloZup:
we skip the dependency because the minion doesn't need to wait for the server, because the minion will be boostrapped during the testsuite.

mbologna:
There is a problem then, since we are still using the old names in the `hostname` directive.
`    name = "head-srv"`
vs
` hostname = "sumaheadpg.tf.local"`

I am in favor of just slowing down a little the creation of the machines but have these two advantages:
* coherence: sticking to the same syntax used in `main.tf.libvirt.example`
* robustness: do not specify twice the hostname of the machine

MalloZup:
imho we can rename the hostaname  to the new hostnames, but i would not add this dependency because it will increase 20/30 min for each run.

mbologna:
Pushed a commit restoring the old version with correction: new hostnames

PR #157, opened 2017-08-04
Fix: specify defaults for ssh_key_path

mbologna:


moio:
I fixed it directly in `master`, as the ternary condition does not really make sense and not all resources were covered.

PR #158, opened 2017-08-07
Install galaxy.key on all SUSE systems

dincamihai:


hustodemon:
LGTM as a workaround for the certificate issue in the testsuite. Some background:
In testsuite for salt minions and salt-ssh systems, we install packages that are signed by our `galaxy.key`. For regular minions, this key is imported as a side-effect importing tools channels. Installing packages signed by this key (e.g. `virgo-dummy`) works fine here.

For salt-ssh systems, however, no tools channels is assigned and no key is imported. Installing packages signed by that key fails.

moio:
Why is this needed?

MalloZup:
I would revert this as franky sayed. Even for the testsuite it didnt fixed the pb

mantel:
Why didn't it fix the testsuite problem? AFAIK we tried to manually import the key into the rpm database and after that signed packages could be installed just fine. What is still failing?

MalloZup:
@mantel have a look on the results, it is not fixed event with this patch

mantel:
The key is not installed:

suma-head-minssh-sles12sp2:~ # rpm -q gpg-pubkey
gpg-pubkey-39db7c82-5847eb1f
suma-head-minssh-sles12sp2:~ # 

So there is some problem in sumaform not correctly applying this state then.

PR #159, opened 2017-08-07
Fix - install galaxy.key on all systems

dincamihai:
Fix installing galaxy.key on all systems, not just SUSE

### DIFF
--- a/salt/default/repos.sls
+++ b/salt/default/repos.sls
@@ -1,5 +1,3 @@
-{% if grains['os'] == 'SUSE' %}
-
 galaxy_key:
   file.managed:
     - name: /tmp/galaxy.key
@@ -8,6 +6,8 @@ galaxy_key:
     - name: rpm --import /tmp/galaxy.key
     - watch:
       - file: galaxy_key
+
+{% if grains['os'] == 'SUSE' %}
 
 {% if grains['osrelease'] == '42.2' %}
 os_pool_repo:


moio:
Why is this needed?

hustodemon:
See here: https://github.com/moio/sumaform/pull/158

In the end it's not a bug in sumaform, but it seems like wrong configuration of our repos in testsuite. We should probably revert this. IIRC the problem is being solved here https://github.com/SUSE/spacewalk/issues/1715.

mantel:
I disagree. This is NOT a wrong configuration of our test repos. While we are used to the fact that the gpg key is contained in the repo itself, this is just a convenience feature but not mandatory. AFAIK Red Hat does not have this feature. It is the same logic like with gpg itself: If you want to check a signature you need to have the respective key available. How this key ends up in your system is up to you. I would prefer to NOT have the key in the repo itself because this will give us more flexibility in the testsuite. Eg. we could remove the key and try to install a signed package which should fail. Then we could uncheck the button "gpgcheck" and try again and this time it should succeed. Finally we could re-install the key (so maybe we should store it somewhere else than in /tmp), check "gpgcheck" again and then package install should succeed. This test would even be idempotent then.
If the key is already available in the repo itself, there is always the "risk" it will get installed automagically by zypper and test will succeed that is supposed to fail.

PR #162, opened 2017-08-09
Gpgkey list for installing gpgkeys

MalloZup:


MalloZup:
Todo, take ispiraration from additional packages

MalloZup:
@moio @mbologna  feel free to review it. i tested it with 2 minion, one without the list, one with.

selection_012

salt/default/init.sls
```diff
@@ -63,14 +63,15 @@ authorized_keys:
...
 {% for keypath in grains['gpg_keys'] %}
-gpg_key_copy_{{ keypath }}:
+{% set keyname =  salt['file.basename'](keypath)  %}
+gpg_key_copy_{{ keyname }}:
   file.managed:
-    - name: /tmp/{{ salt['file.basename'](keypath) }} 
+    - name: /tmp/{{ keyname }} 
     - source: salt://{{ keypath }}
-install_{{ keypath }}:
+install_{{ keyname }}:
```
MalloZup: may

modules/libvirt/host/variables.tf
```diff
@@ -63,3 +63,8 @@ variable "ssh_key_path" {
   default = "/dev/null"
   # HACK: "" cannot be used as a default because of https://github\.com/hashicorp/hil/issues/50
 }
+
+variable "gpg_keys" {
+  description = "gpg keys that youw want to add and install to your VMs, see libvirt/README.md"
```
mbologna: *you

salt/default/init.sls
```diff
@@ -60,3 +60,17 @@ authorized_keys:
 {% endfor %}
     - makedirs: True
 {% endif %}
+
+{% if grains['gpg_keys'] %}
```
mbologna: Since it is related to repos yes, makes sense to move it to repos.sls

modules/libvirt/host/variables.tf
MalloZup: youw is more hipster

salt/default/init.sls
MalloZup: ok, i'm refractoring there a bit, i will push asap. :tiger:

MalloZup:
@moio  personally, i tend to don"t like to much the `file` function, because it  reads the contents of a file into the string.
Imho we should assume that somebody that want to put extra stuff, like saltstack assume, he need to put it in managed dir, and not allow this kind of freedom.

On the other side i can see the point

MalloZup:
@mbologna  for the travis ci, i would make a script that prints out which steps are executed and which fails.

at moment to have && is to much encapsulate imho.

mbologna:
@MalloZup there is the indication of what's failing. Latest commit at time of writing:

```
Error loading Terraform: Error downloading modules: module suma3pg: Error parsing .terraform/modules/db20d889770318f732fc5cd8dbf79244/variables.tf: At 117:17: Unknown token: 117:17 IDENT gpg
```

MalloZup:
@mbologna sure i saw that, imho we need:

- update the libvirt.example to use the same names as the testsuite, since suma3pg is not more actual

PR #166, opened 2017-08-11
Taint an image resource

ncounter:
It happens sometime you want to refresh your images, so you want to `taint` them to get a new downloaded image.
The syntax is the same when tainting any other resource, but the name of the module and the number of the volume are slightly different. I think it can be useful to have it documented here to get it quickly done.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -20,9 +20,9 @@ module.suma3pg.module.suse_manager.libvirt_volume.main_disk
 $ terraform taint -module=suma3pg.suse_manager libvirt_volume.main_disk
 The resource libvirt_volume.main_disk in the module root.suma3pg.suse_manager has been marked as tainted!
 ```
-#### how to force the re-download of a fresh image resource?
+### Q: how to force the re-download of an image?
 
-*warning: any dependent volume and module should be tainted as well before applying.*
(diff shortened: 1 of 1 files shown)


TROUBLESHOOTING.md
```diff
@@ -20,6 +20,17 @@ module.suma3pg.module.suse_manager.libvirt_volume.main_disk
...
+
+```
+$ terraform state list
+...
+module.base.libvirt_volume.volumes[2]
+
+$ terraform taint -module=base libvirt_volume.volumes.2
+The resource libvirt_volume.volumes.2 in the module root.base has been marked as tainted!
+```
+
```
moio: Please clarify that any resource depending on the volume has to be manually destroyed as well - IOW one should taint the base volume and any dependent modules before applying.

ncounter: Right. Addressed

cbbayburt:
I run this script whenever i need `taint`

```bash
#!/bin/bash

terraform taint -module $1 libvirt_domain.domain && \
terraform taint -module $1 libvirt_volume.main_disk
```

Maybe we could provide this script as a utility?

moio:
@cbbayburt that would be a possibility, the other one would be to fix the problem at the source (libvirt plugin and our definitions). This should really be handled at Terraform level.

PR #168, opened 2017-08-11
salt-travis basic validation

MalloZup:


.travis.yml
```diff
@@ -11,6 +11,7 @@ before_install:
   - sudo alien -i terraform.rpm
   - wget http://download.opensuse.org/repositories/home:/SilvioMoioli:/tools/SLE_12_SP1/x86_64/terraform-provider-libvirt.rpm
   - sudo alien -i terraform-provider-libvirt.rpm
+  - sudo apt-get install salt-common salt-master salt-minion
```
mbologna: We should use the same Salt version as in 3.1-released branch

MalloZup:
Yop, atm it doesnt work, no idea why but the states aren executed lol

.travis.yml
MalloZup: @mbologna can you add this repo

MalloZup:
The first check is good to have because it check for duplicate of name of states:

As you can see, i make it fail explicetely
```console
[INFO    ] Loading fresh modules for state activity

[CRITICAL] Rendering SLS "base:suse_manager_server.postgres" failed: Conflicting ID "postgresql_hba_configuration"

local:

    - Rendering SLS "base:suse_manager_server.postgres" failed: Conflicting ID "postgresql_hba_configuration"
```

MalloZup:
@mbologna  i think we cannot add 3.1 repos because we cannot reach those server from network point of view for adding the repos 3.1, so the ubuntu salt pkg are ok and the only one solution.

mbologna:
We can use different version of SaltStack packaged for Ubuntu from this repo:
https://repo.saltstack.com/#ubuntu

MalloZup:
yop, but this is then the same as doing apt install salt-minion.
or do you want a specific version?

MalloZup:
@mbologna @moio  feel free to review it.

I added at moment only a script for testing the role susemanager server.

We could add more roles, like client and minion. Maybe we could track it in a issue

.travis.yml
```diff
@@ -6,19 +6,16 @@ env:
     - TMPDIR=~/tmp
 
 before_install:
-  - sudo apt install -y alien
+  - sudo apt-get install -y alien
```
mbologna: apt is ok

.travis.yml
```diff
@@ -6,19 +6,16 @@ env:
...
   - wget http://download.opensuse.org/repositories/home:/SilvioMoioli:/tools/SLE_12_SP1/x86_64/terraform.rpm
   - sudo alien -i terraform.rpm
   - wget http://download.opensuse.org/repositories/home:/SilvioMoioli:/tools/SLE_12_SP1/x86_64/terraform-provider-libvirt.rpm
   - sudo alien -i terraform-provider-libvirt.rpm
+  - sudo apt-get install salt-common salt-minion
```
mbologna: same as above

.travis.yml
```diff
@@ -6,19 +6,16 @@ env:
...
-        terraform get &&
-        terraform validate &&
-        mv main.tf.libvirt-testsuite.example main.tf &&
-        terraform get &&
-        terraform validate
-deploy:
+
+script:
+  - bash .ci-travis/terraform-validation
+  - sudo bash .ci-travis/salt-server-validation
```
mbologna: Please separate the parts that require `sudo` (creating files in `/srv` and `/etc/`) and parts that do not require `sudo` (`salt-call`).

MalloZup:
@mbologna  i separeted the sudo stuff inside the file. what do you think

.travis.yml
MalloZup: @mbologna  this will be overcomplicated to do so, since we need to restart service etc.

I would vote to be more relaxed, since is a test machine container

PR #173, opened 2017-08-23
Fix gpg keys

MalloZup:
This renaming https://github.com/moio/sumaform/commit/3e7af3a18beeb8efcfaa02eb4d6a7755df486d6e

was breaking all state that use the gpg keys.

I would vote to conserve my initial naming with gpg_keys with s at the end, since we have more then 1.


moio:
Fixed directly in `master`, sorry, it was a rebase mistake from my part.

PR #180, opened 2017-08-29
Update README.md

keichwa:
The list of images is defined in modules/libvirt/base/variables.tf


moio:
Thanks for fixing this!

PR #182, opened 2017-08-30
Test: run Travis in containers instead of VMs

moio:
See https://docs.travis-ci.com/user/reference/trusty/#Routing-to-Trusty

### DIFF
--- a/.ci-travis/minion
+++ b/.ci-travis/minion
@@ -1,4 +1,0 @@
-file_client: local
-file_roots:
-  base:
-    - /srv/salt/states
--- a/.ci-travis/salt-server-validation
+++ b/.ci-travis/salt-server-validation
@@ -1,21 +1,7 @@
 #! /bin/bash
-
 set -xe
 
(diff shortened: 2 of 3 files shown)


MalloZup:
@moio  we need sudo,  why sudo false, let the sudo be true !!! :)

moio:
@MalloZup: look ma', no `sudo`!

MalloZup:
@moio  where do you reload cached grains? 
are we sure that when removing the grains local and putting a new, we will have not the old data?

This was the reason for restarting the salt service ( i spoked with @meaksh  about that)

moio:
If you use `salt-call` you do not need `salt-master` nor `salt-minion`, it is just one only self-contained execution.

The salt-minion package does not configure any custom grain, all custom grains are set by the script, so I see no reason to reload them.

MalloZup:
ok for me then

merge this also then! :)
https://github.com/moio/sumaform/pull/176

PR #184, opened 2017-08-30
Automatically accept minion keys by default

moio:
Enables `auto_accept` on all Salt masters by default.


MalloZup:
looks good, nice to have it +1

PR #185, opened 2017-08-31
Refactorings

moio:
Just some cleanups in advance of feature additions I am working at.


moio:
Thanks guys, it really helps to have speedy reviews!

PR #186, opened 2017-08-31
Add option to reposync channels at deploy time

moio:
Ever wanted to have some channels and products pre-reposynced in your SUSE Manager Server?

Now you can :smile:

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -62,6 +62,30 @@ module "minionsles12sp1" {
 
 This will create 10 minions connected to the `suma3pg` server.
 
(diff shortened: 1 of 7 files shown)


MalloZup:
@moio looks cool the feature, the only thing that i"m wondering is that, we should make the reposync in background afaik, because waiting will block terraform to deploy the vms.

moio:
That's actually my use case: I want to start onboarding minions only when channels are ready!

If someone really needs the background option, it will be very easy to skip executing `wait_for_reposync.py` in a future PR.

MalloZup:
@moio  ok  then :panda_face:

PR #189, opened 2017-09-05
Add screen package to the controller machine

cbbayburt:
Anytime I setup a new controller, I find myself in need of `screen` tool to run the test suite with, so I can log out of the machine re-attach the terminal which the tests are running at a later time.

I'd be nice to have this tool available by default on the controller machine.

### DIFF
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -92,3 +92,10 @@ testsuite_env_vars:
     - user: root
     - group: root
     - mode: 755
+
+extra_pkgs:
+  pkg.installed:
+    - pkgs:
+      - tmux
+    - require:
+      - sls: controller.repos


MalloZup:
ok for me, if you can add a documentation howto for using it. :+1:

moio:
JFI: why `screen` and not `tmux`?

cbbayburt:
@moio it's just because personally I am more familiar with it.
Here, I just assume it has a bigger user base than tmux, but we can vote of course :)

MalloZup:
i think tmux is better solution, just reading in web, it has more stability and features then screen.

mbologna:
Personal favorite: tmux. They are smaller in size, so I would say we can add them both.

MalloZup:
No one is ok

cbbayburt:
Ok, then let's go on with `tmux`

MalloZup:
@cbbayburt  +1
I would like to have a short Documentation to describe your workflow, otherwise i see this PR as incomplete (just adding a pkg).

For me, in sumaform or spacewalk-testsuite-base, both make sense. Just refer it here.

Thx

moio:
Good from my side, merging.

PR #191, opened 2017-09-15
Add instructions for tmux

cbbayburt:
Instructions to use `tmux` for running the testsuite in the advanced readme.


moio:
I have reverted your `tmux` patch because the package is not there in SLES 12 SP1 :cry: #192 

Please either add a conditional to only have it installed on OSs that ship it, or let's have it `copypac`'d in the tools repo, or let's just use `screen` - include this commit to the PR that adds it back.

Sorry!

cbbayburt:
I think `screen` is the simplest way to solve the initial problem in this case. I'll create a new PR. Thanks.

PR #197, opened 2017-10-04
WIP: set new database password

keichwa:
The SUMA Web UI does not accept dictionary-based passwords.

### DIFF
--- a/salt/postgres/init.sls
+++ b/salt/postgres/init.sls
@@ -82,7 +82,7 @@ postgresql_user:
 {% else %}
     - user: postgres
 {% endif %}
-    - name: echo "CREATE ROLE spacewalk PASSWORD 'spac8ewalk' SUPERUSER NOCREATEDB NOCREATEROLE INHERIT LOGIN;" | psql
-    - unless: psql -c "\du" | grep spacewalk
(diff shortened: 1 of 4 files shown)


MalloZup:
@keichwa  it is not so easy as it seems, to change this password, a bunch of programs rely on this `spacewalk` pwd

moio:
OK @keichwa, that was not the only place in order to have the password correctly changed. I added a commit in this sense and I wonder now: would `susemanager` be an acceptable password?

I am asking because developers have to use this credential quite often, and I would like to use the easiest possible alternative by default.

Bischoff:
To me, ```susemanager``` would be an acceptable password.

And yes, we need to fix it in many programs afterwards. But I don't think we can do anything about that, apart from doing it with care.

keichwa:
Yes, susemanager would is an acceptable password.

While running /usr/lib/susemanager/bin/mgr-setup -m I see some 100 lines of this error prints (ugly, but in the end it seems to work):

ERROR:  role "spacewalk" does not exist
ERROR:  role "spacewalk" does not exist
ERROR:  role "spacewalk" does not exist

MalloZup:
i would just vote for closing this

Bischoff:
@keichwa ?

moio:
Closing for the time being

PR #199, opened 2017-10-10
Head should use packages build for SP3

mcalmer:
Head does not build anymore for SP2. We need to use SP3 now.

### DIFF
--- a/salt/suse_manager_server/repos.d/Devel_Galaxy_Manager_Head.repo
+++ b/salt/suse_manager_server/repos.d/Devel_Galaxy_Manager_Head.repo
@@ -1,6 +1,6 @@
 [Devel_Galaxy_Manager_Head]
-name=Devel Project for SUSE Manager Head (SLE_12_SP2)
+name=Devel Project for SUSE Manager Head (SLE_12_SP3)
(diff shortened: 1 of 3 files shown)


salt/suse_manager_server/repos.d/Devel_Galaxy_Manager_Head.repo
```diff
@@ -1,6 +1,6 @@
 [Devel_Galaxy_Manager_Head]
-name=Devel Project for SUSE Manager Head (SLE_12_SP2)
+name=Devel Project for SUSE Manager Head (SLE_12_SP3)
 type=rpm-md
 enabled=1
-baseurl=http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/Devel:/Galaxy:/Manager:/Head/SLE_12_SP2/
+baseurl=http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/Devel:/Galaxy:/Manager:/Head/SLE_12_SP3/
```
MalloZup: imho this var is hardcoded.
We shoulc create a global var, like HEAD=12sp3, which will change only one time

PR #200, opened 2017-10-11
Add netrc for cloning with git_credentials

MalloZup:
This is needed when we want to clone from private repo

### DIFF
--- a/modules/libvirt/controller/main.tf
+++ b/modules/libvirt/controller/main.tf
@@ -1,11 +1,11 @@
 variable "testsuite-branch" {
   default = {
-    "3.0-released" = "manager30"
-    "3.0-nightly" = "manager30"
-    "3.1-released" = "manager31"
-    "3.1-nightly" = "manager31"
-    "head" = "master"
-    "test" = "master"
+    "3.0-released" = "Manager-3.0"
+    "3.0-nightly" = "Manager-3.0"
+    "3.1-released" = "Manager-3.1"
+    "3.1-nightly" = "Manager-3.1"
+    "head" = "Manager"
+    "test" = "Manager"
   }
 }
 
@@ -22,6 +22,9 @@ module "controller" {
   ssh_key_path = "${var.ssh_key_path}"
   grains = <<EOF
 
+
+git_username: ${var.git_username}
+git_password: ${var.git_password}
 mirror: ${var.base_configuration["mirror"]}
 server: ${var.server_configuration["hostname"]}
 proxy: ${var.proxy_configuration["hostname"]}
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -65,15 +65,20 @@ cucumber_requisites:
     - require:
       - sls: controller.repos
 
-cucumber_testsuite:
+spacewalk_git_repository:
   git.latest:
-    - name: https://github.com/SUSE/spacewalk-testsuite-base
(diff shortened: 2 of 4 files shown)


MalloZup:
star

MalloZup:
https://github.com/moio/sumaform/pull/198

MalloZup:
@moio ready for a review. 
We need to wait before merging, because it will break the production (since the testsuite dir doesn't exist)
a part of this, i think it is ok

modules/libvirt/base/variables.tf
```diff
@@ -8,6 +8,18 @@ variable "cc_password" {
   type = "string"
 }
 
+variable "git_username" {
```
moio: Can you please move these variables out of `base` and in the `controller` module please? We should not really need them in other hosts.

MalloZup: @moio  ok, i was in doubt were to place them.thx

MalloZup:
@moio 
a part of this, i noticed that having spacewalk in the controller make a lot of positive changes:
- the controller could become the dev machine , where we install `ant` and other tools for deploying src code, adn then run directly the test. 
- we could even imagine to create an ant target where we run some tests

MalloZup:
@moio 
i get right know with the actual commit:

when the controller is runned
```
[ERROR   ] Rendering exception occurred: Jinja variable 'dict object' has no attribute 'hostname'
[CRITICAL] Rendering SLS 'banse:default' failed: Jinja variable 'dict object' has no attribute 'hostname'
local:
    Data failed to compile:
---------    Rendering SLS 'base:default' failed: Jinja variable 'dict object' has no attribute 'hostname'
```
Did you ever had this issue? just for an hint

salt/controller/init.sls
```diff
@@ -99,3 +104,22 @@ extra_pkgs:
       - screen
     - require:
       - sls: controller.repos
+
+git_config:
+  file.append:
+    - name: ~/.netrc
```
moio: Is the file existing already at this point or are you creating it from scratch?

moio:
> [CRITICAL] Rendering SLS 'banse:default'

ba**n**se looks like a typo

MalloZup:
ahha

salt/controller/init.sls
MalloZup: from scratch

modules/libvirt/controller/main.tf
```diff
@@ -22,6 +22,9 @@ module "controller" {
   ssh_key_path = "${var.ssh_key_path}"
   grains = <<EOF
 
+
+git_username = "${var.git_username}"
```
MalloZup: @moio  here was the typo :rofl: 
``` git_username: ..```

File: salt/controller/init.sls [hunk 4]
```diff
@@ -65,15 +65,20 @@ cucumber_requisites:
     - require:
       - sls: controller.repos
 
-cucumber_testsuite:
+spacewalk_git_repository:
   git.latest:
-    - name: https://github.com/SUSE/spacewalk-testsuite-base
-    - target: /root/spacewalk-testsuite-base
+    # FIXME: this is hardcoded to SUSE(for opensourcing remove
```
moio: Please remove this FIXME and open an issue instead, so that's easier to track

MalloZup: ok you are right

PR #202, opened 2017-10-16
Support for proxy on controller

Bischoff:
* Added possibility to configure proxy on controller
* Added variable $PROXY on controller

### DIFF
--- a/modules/libvirt/controller/variables.tf
+++ b/modules/libvirt/controller/variables.tf
@@ -13,6 +13,14 @@ variable "server_configuration" {
   type = "map"
 }
 
+variable "proxy_configuration" {
+  description = "use ${module.<PROXY_NAME>.configuration}, see main.tf.libvirt-testsuite.example"
+  type = "map"
+  default = {
+    hostname = "null"
+  }
+}
+
 variable "client_configuration" {
   description = "use ${module.<CLIENT_NAME>.configuration}, see main.tf.libvirt-testsuite.example"
   type = "map"
--- a/.ci-travis/salt-ctl-validation
+++ b/.ci-travis/salt-ctl-validation
@@ -14,6 +14,7 @@ authorized_keys: [ssh-rsa 1233= root@travis.git]
 reset_ids: true
 version: head
 server: { hostname = "travis.tf.local" }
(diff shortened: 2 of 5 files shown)


modules/libvirt/controller/variables.tf
```diff
@@ -13,6 +13,14 @@ variable "server_configuration" {
   type = "map"
 }
 
+variable "proxy_configuration" {
+  description = "use ${module.<PROXY_NAME>.configuration}, see main.tf.libvirt-testsuite.example"
+  type = "map"
+  default = {
+    hostname = "None"
```
MalloZup: @mbologna @moio  i think that setting the value to None here is kind like assigning an Object to nil value.
I would prefer that we use  "false" for the hostname

Bischoff: "None" is what Salt answers if a grain is missing. This line is emulating the same behaviour.

Also, in English, "hostname = one" makes more sense than "hostname = false" :smile_cat: .

moio: This string is passed to Salt via grains, which are encoded in a YAML file.

Per YAML specification, `None` is a string (eg. a `N`, `o`, `n`, `e` character array), while `null` is really a missing value. See this example:

http://yaml-online-parser.appspot.com/?yaml=my+funny+list%3A%0A+-+None%0A+-+null%0A&type=json

Please use `null` here, this is also what we do [in](https://github.com/moio/sumaform/blob/master/modules/aws/host/variables.tf#L63) [other](https://github.com/moio/sumaform/blob/master/modules/aws/host/variables.tf#L93) [modules](https://github.com/moio/sumaform/blob/master/modules/libvirt/suse_manager/variables.tf#L68).

moio:
Good to merge as soon as [this comment](https://github.com/moio/sumaform/pull/202#discussion_r146242705) is addressed.

modules/libvirt/controller/variables.tf
Bischoff: I don't use "None". Nowhere.

The place where I "use" it as a default value is just a safeguard and it is not even being executed. I could put "hostname = "wakajaka"" there with same result.

PR #203, opened 2017-10-17
Answer to "Proxy Version to Activate" with "3.1"

Bischoff:
Ideally, it should be taken from a variable (which one?)


MalloZup:
You take it from susemanager server or you create a proxy var version if not exists

Bischoff:
How do I get it from server? All I found is grains['version'] which might contain things like "3.1-nightly".

Do I realy have to come up with a long disjunction like
```
{% if '2.1' in grains['version'] %}
...
{% endif %}
{% elif '3.0-released' in grains['version'] %}
...
{% endif %}
{% elif '3.0-nightly' in grains['version'] %}
...
{% endif %}
{% elif '3.1-released' in grains['version'] %}
...
{% endif %}
{% elif '3.1-nightly' in grains['version'] %}
...
{% endif %}
{% elif 'head' in grains['version'] %}
...
{% endif %}
```

That looks terrible.

Bischoff:
no answer - doing it with the long disjunction

PR #216, opened 2017-11-04
change hostname using state

mcalmer:
Appending ::1 line to /etc/hosts cause duplicate entry.

just appending the hostnames to ::1 has bad side effects.
At least not setting the short hostname to ::1 prevent returning "localhost" with hostname --fqdn

Without this I had the following effect:

```
$> hostname
suma-31-cli-sles12sp3
$> hostname -f 
suma-31-cli-sles12sp3.mgr.suse.de
$> python -c "import socket; print socket.getfqdn()"
localhost
```

No duplicate ::1 but adding FQDN and short hostname to the existing ::1 entry cause this:
```
$> hostname
suma-31-cli-sles12sp3
$> hostname -f 
localhost
$> python -c "import socket; print socket.getfqdn()"
localhost
```
Do not add the short hostname to the existing ::1 entry print the right things:
```
$> hostname
suma-31-cli-sles12sp3
$> hostname -f 
suma-31-cli-sles12sp3.mgr.suse.de
$> python -c "import socket; print socket.getfqdn()"
suma-31-cli-sles12sp3.mgr.suse.de
```

### DIFF
--- a/salt/default/hostname.sls
+++ b/salt/default/hostname.sls
@@ -1,9 +1,15 @@
-hosts_file:
-  file.append:
-    - name: /etc/hosts
-    - text: |
-        127.0.1.1 {{ grains['hostname'] }}.{{ grains['domain'] }} {{ grains['hostname'] }}
-        ::1 {{ grains['hostname'] }}.{{ grains['domain'] }} {{ grains['hostname'] }}
+hosts_file_ipv4:
+  host.present:
+    - ip: 127.0.1.1
+    - names:
+      - {{ grains['hostname'] }}.{{ grains['domain'] }}
+      - {{ grains['hostname'] }}
+
+hosts_file_ipv6:
+  host.present:
+    - ip: ::1
+    - names:
+      - {{ grains['hostname'] }}.{{ grains['domain'] }}
 
 temporary_hostname:
   cmd.run:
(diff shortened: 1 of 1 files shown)


mcalmer:
This has fixed the tests - check galaxy-ci mails

MalloZup:
lgtm for me

MalloZup:
just a side note, remember to set the sumaform-testrunner to master when we merge this pr. 
:sunflower: :sun_with_face:

Bischoff:
With all respect, Michael, I disagree with this patch :smile: .

Declarations for something else than ```localhost``` of ```127.0.0.1``` and ```::1``` should simply go away.

What we need is a FQDN in ```/etc/hostname```.

mcalmer:
@Bischoff we only want to fix all the problems we have. One was that salt is complaining that without hostname for IPv6 address the generation of grains takes 10 seconds longer. I do not see that your PR solve this problem. 

On the other hand there is a big discussion if /etc/hostname should contain the FQDN or only the short hostname. AFAIK there is a systemd documentation which say 'only the short hostname'

Bischoff:
@mcalmer Right, I was not aware of the situations where sumaform runs without valid DNS.

About the big discussion about /etc/hostname : yes, people disagree on this. On my side, I just experienced that we lose the domain name in the result of "hostname -f" after a reboot if the domain name is not given in /etc/hostname.

moio:
This has been superseded by commits directly in master which tackle the problem from a slightly different angle.

Thanks anyway @mcalmer for the useful analysis!

PR #217, opened 2017-11-07
Fixing FQDN problems

Bischoff:
We need the domain name inside ```/etc/hostname```, this PR adds it


Bischoff:
Does not seem to work - getting crazy.

PR #218, opened 2017-11-13
Fix example in case of proxy

Bischoff:
* if we have a proxy, clients must point to proxy, not to server
* aligned syntax with ```main.tf.libvirt.example```

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -27,10 +27,10 @@ module "head-srv" {
   ssh_key_path = "./salt/controller/id_rsa.pub"
 }
 
-module "head-proxy" {
+module "head-pxy" {
   source = "./modules/libvirt/suse_manager_proxy"
   base_configuration = "${module.base.configuration}"
-  name = "head-proxy"
+  name = "head-pxy"
   version = "head"
   server_configuration = { hostname = "head-srv.tf.local" }
   for_development_only = false
@@ -42,7 +42,7 @@ module "head-cli-sles12sp3" {
   base_configuration = "${module.base.configuration}"
   name = "head-cli-sles12sp3"
   image = "sles12sp3"
-  server_configuration =  { hostname = "head-srv.tf.local" }
+  server_configuration = { hostname = "head-pxy.tf.local" }
   for_development_only = false
   for_testsuite_only = true
   ssh_key_path = "./salt/controller/id_rsa.pub"
@@ -54,7 +54,7 @@ module "head-min-sles12sp3" {
   version = "head"
   name = "head-min-sles12sp3"
   image = "sles12sp3"
-  server_configuration = { hostname = "head-srv.tf.local" }
+  server_configuration = { hostname = "head-pxy.tf.local" }
   for_development_only = false
   for_testsuite_only = true
   ssh_key_path = "./salt/controller/id_rsa.pub"
@@ -75,7 +75,7 @@ module "head-min-centos7" {
   version = "head"
   name = "head-min-centos7"
   image = "centos7"
-  server_configuration = { hostname = "head-srv.tf.local" }
+  server_configuration = { hostname = "head-pxy.tf.local" }
   for_development_only = false
   for_testsuite_only = true
   ssh_key_path = "./salt/controller/id_rsa.pub"
@@ -86,7 +86,7 @@ module "head-ctl" {
   base_configuration = "${module.base.configuration}"
   name = "head-ctl"
   server_configuration = "${module.head-srv.configuration}"
-  proxy_configuration = "${module.head-proxy.configuration}"
+  proxy_configuration = "${module.head-pxy.configuration}"
   client_configuration = "${module.head-cli-sles12sp3.configuration}"
   minion_configuration = "${module.head-min-sles12sp3.configuration}"
   centos_configuration = "${module.head-min-centos7.configuration}"


main.tf.libvirt-testsuite.example
```diff
@@ -32,7 +32,7 @@ module "head-proxy" {
   base_configuration = "${module.base.configuration}"
   name = "head-proxy"
   version = "head"
-  server_configuration = { hostname = "head-srv.tf.local" }
+  server_configuration = {module.head-srv.configuration}
```
moio: The second syntax introduces a dependency between the two modules, that means Terraform will wait for the server to be up before starting deploying the Proxy. This was not wanted in other cases, as it wasted time.

Are you sure this is what you need here?

Bischoff: No. But then why do use that syntax in the other example file ?

main.tf.libvirt-testsuite.example
```diff
@@ -42,7 +42,7 @@ module "head-cli-sles12sp3" {
   base_configuration = "${module.base.configuration}"
   name = "head-cli-sles12sp3"
   image = "sles12sp3"
-  server_configuration =  { hostname = "head-srv.tf.local" }
+  server_configuration =  {module.head-proxy.configuration}
```
MalloZup: why are you changing this to proxy?

Bischoff: because the client is connected to the proxy

MalloZup:
from my pov:

we should just remove the proxy from this example file, since the proxy is not *mandatory*, and will not run every run on production.

This example here is for running the testsuite locally and should be fast, simplier as possible, without to much dependencies

main.tf.libvirt-testsuite.example
MalloZup: we have this for making the testsuite run faster. 

This server configuration variable is needed only when you have a **dev** server.

So we give it a fake variable, so that terraform will not wait the server resource and will spawn machines concurently.

moio:
I would agree with removing the proxy - this file here is mainly for developers and we want to make life easy and fast for them in order to lower the barrier to the use of the testsuite - as a result involving them more.

But this decision I will leave it to you guys.

@MalloZup please discuss with @Bischoff and other interested parties and either fix this PR before merging it or close it and start another.

Thanks

Bischoff:
@mallozup: it is always simpler to ''remove'' than to add.

This is just an example. Someone who does not want the proxy can easily remove it.

Later, when we make centos and ssh optional, it will be the same - easier to remove than to add.

If you want, we can add instructions on how to remove the proxy to that file.

Bischoff:
@moio : there are developers who test with proxy - for example Michele Bologna.

MalloZup:
this file should reflect 100% the production and easy way. 

Why i  don't like this? 

Here we have a **silent dependency** where people could think that the minion are bootstrapped against the server, which is not true, because having this will bootstrap minion against proxy.

( this is in the testsuite itself) and on the terraform level is not clear.

So i see all problems with this kind of silent mechanisms.

Having 2 files  will be the longterm safest mechanism.

Bischoff:
This is just an example file. Anyone can remove anything he does not want.

It's much harder to add something.

When we will make centos and minion ssh optional, it will be the same - the example will show how to do them all, then everyone will be able to remove the things he does not want.

Also, the default *should be* testing with everything.

Bischoff:
@MalloZup. The production will be with proxy too. At least half of the week.

MalloZup:
again this will make bootstrap minions against proxy, which will be mean that people to think wrong behaviour.

You should respect also previous design not destroying it with new.

Bischoff:
It will not make minions bootstrap against proxy if you run the testsuite.
Just like it does not make minion bootstrap against server if you run the testsuite.

If you don't run the testsuite, using a proxy is not "wrong behaviour". This is a central feature of the product.

MalloZup:
@Bischoff  this file here is intended to be run in the testsuite. 

**without any condition** , **1 behaviour**. 

If you add proxy in the testsuite, the skipped test are runned. 


The mean goal of this example. is to make
```bash
terraform apply
run-testsuite
```
Without to think further or if else.

I am speaking about the design of this mechanism, we are not speaking about products etc.

Bischoff:
The default run should be to run with proxy.

MalloZup:
@Bischoff  can you argue why? 

So you are saying that bootstraping against the proxy is more important that bootstraping against the server? Because the 2 behaviours are mutal exclusiv at moment.

Bischoff:
Because testing more is better than testing less.

MalloZup:
@Bischoff you are **not testing more**.

You are testing only 1 behaviour of 2 with this.
And you choose the most used one.
You are testing the** less **used feature (bootstraping against server).

OK, create a card or an email to galaxy-devel for this. I will not approve this PR,

Bischoff:
@moio , please decide what to do with this PR.

main.tf.libvirt-testsuite.example
moio: In the main example file `for_development_only` is implied - that means that the deployment and initial configuration are handled by sumaform. That means we have to honour the deployment ordering, meaning that eg. minions cannot be deployed until the server is up.

The testsuite example file is different, because it assumes the deployment is done by sumaform but the configuration is part of the testsuite itself. In this case we do not want to wait, and deploy all VMs simultaneously to save time.

I thus assume you want to keep the same assumption here.

Bischoff: @moio: this has already been changed to ```hostname = "head-srv.tf.local```. I'm a bit lost, that's what everyone here wants, right?

moio:
I am closing this now and removing the proxy as per the previous comment - we want to keep the barrier of entry to the testsuite as low as possible. This is for users who want to run the testsuite on their machines, we want to make their life as easy as possible.

In that light I am open to accept another PR which adds a `testsuite-with-proxy` example file (and a remark on the README so that interested users know what to do). That will help users who want or need to run the testsuite locally with proxy enabled.

Of course our automated testsuite should really cover both cases - FMPOV they are really equally important.

Bischoff:
OK @moio. Thanks for your time.

PR #219, opened 2017-11-15
Resolve hostname problems

Bischoff:
1) We fix FDDN problems by reverting to old strategy:
* modify ```/etc/hostname```
* change current hostname with ```hostname``` or ```hostnamectl --set-hostname```

but this time we take care of providing the domain when we change the current hostname.

2) We work around the workaround (!) for [this bug](https://bugzilla.suse.com/show_bug.cgi?id=820213) because otherwise at reboot the FQDN gets replaced by a short name, which puzzles avahi and makes salt-minion complain.


Bischoff:
abandoning that path for now...

PR #225, opened 2017-11-22
Refactoring - rearrange code for proxy

Bischoff:
This PR moves a big part of ```init.sls``` to a new file ```configuration.sls```, because it has grown too big IMHO.

The new file contains initialization code that is executed only in case of "development_only" setup.


salt/suse_manager_proxy/configuration.sls
```diff
@@ -0,0 +1,131 @@
+{% if grains['for_development_only'] %}
```
moio: Please rename this file to `development.sls`, similarly to what we do with `testsuite.sls` in some places.

At that point, if contents did not change, feel free to merge (as a single rebased commit please).

Bischoff: ok

PR #226, opened 2017-11-23
More hostname fixes

moio:
- uses the short name in `/etc/hostname` and FQDN in `/etc/HOSTNAME` as [per SUSE recommendation in bug 972463](https://bugzilla.suse.com/show_bug.cgi?id=972463#c25). Using the FQDN in `/etc/hostname` is not OK at least in SUSE systems because of a [downstream patch in systemd](https://github.com/openSUSE/systemd/commit/2ce06a5a49)
 - retains a custom Python script in order to write to `/etc/hosts`
   - this is needed to support cases in which a well-configured DNS server is not available, as avahi alone is not sufficient to have both `hostname` and `hostname -f` return correct values. That is true even when enabling the [--enable-search-domains](http://0pointer.de/lennart/projects/nss-mdns/) option in `nss-mdns` and adding the avahi domain to search domains in `resolv.conf`
   - notr that we now avoid using IPv6 link-local addresses as those are not sufficient without an `%interface` part that Python's `socket.getaddrinfo` currently does not return
 - documents choices better
 - has been tested with all supported scenarios through the following one-liner:
```
echo "short: `hostname`"; echo "long: `hostname -f`"; echo "domain: `hostname -d`"; echo "python long: `python -c "import socket; print socket.getfqdn()"`"; echo "/etc/hostname: `cat /etc/hostname`"; echo "/etc/HOSTNAME: `cat /etc/HOSTNAME`"; echo "/etc/hosts: `tail -n2 /etc/hosts`"; ping -c1 `cat /etc/hostname`
```

### DIFF
--- a/salt/default/hostname.sls
+++ b/salt/default/hostname.sls
@@ -23,13 +23,12 @@ permanent_hostname:
     - name: /etc/hostname
     - contents: {{ grains['hostname'] }}
 
-{% if grains['os_family'] == 'Suse' %}
-SUSE_specific_permanent_hostname:
+# /etc/HOSTNAME is supposed to always contain the FQDN
+legacy_permanent_hostname:
   file.managed:
     - name: /etc/HOSTNAME
     - follow_symlinks: False
     - contents: {{ grains['hostname'] }}.{{ grains['domain'] }}
-{% endif %}
 
 # set the hostname and FQDN name in /etc/hosts
 # this is not needed if a proper DNS server is in place, but when using avahi this
--- a/salt/default/hostname.sls
+++ b/salt/default/hostname.sls
@@ -1,4 +1,43 @@
-hostname:
+# set the hostname in the kernel, this is needed for Red Hat systems
+# and does not hurt in others
+kernel_hostname:
+  cmd.run:
+    - name: sysctl kernel.hostname={{ grains['hostname'] }}
+    - unless: sysctl --values kernel.hostname | grep -w {{ grains['hostname'] }}
+
+# set the hostname in userland. There is no consensus among distros
+# but Debian prefers the short name, SUSE demands the short name,
+# Red Hat suggests the FQDN but works with the short name.
+# Bottom line: short name is used here
+temporary_hostname:
+  cmd.run:
+    {% if grains['init'] == 'systemd' %}
+    - name: hostnamectl set-hostname {{ grains['hostname'] }}
+    {% else %}
+    - name: hostname {{ grains['hostname'] }}
+    {% endif %}
+
+# set the hostname in the filesystem, matching the temporary hostname
+permanent_hostname:
+  file.managed:
+    - name: /etc/hostname
+    - contents: {{ grains['hostname'] }}
+
+{% if grains['os_family'] == 'Suse' %}
+SUSE_specific_permanent_hostname:
+  file.managed:
+    - name: /etc/HOSTNAME
+    - follow_symlinks: False
+    - contents: {{ grains['hostname'] }}.{{ grains['domain'] }}
+{% endif %}
+
+# set the hostname and FQDN name in /etc/hosts
+# this is not needed if a proper DNS server is in place, but when using avahi this
+# might not be the case. The script tries to to use real IP addresses in order not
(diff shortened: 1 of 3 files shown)


moio:
@Bischoff please review

Bischoff:
Is the ```search tf.local``` really not sufficient? (this is a purely academic question, as we will go the ```/etc/hosts``` way anyway.

The removal of the IPv6 line has been tested against the testsuite with success (two errors, but probably unrelated).

moio:
According to my tests, yes. Will ping you privately to let you independently prove that - if I made a mistake and that approach is viable I will be happy to correct course.

salt/default/hostname.sls
```diff
@@ -1,4 +1,43 @@
...
+    {% endif %}
+
+# set the hostname in the filesystem, matching the temporary hostname
+permanent_hostname:
+  file.managed:
+    - name: /etc/hostname
+    - contents: {{ grains['hostname'] }}
+
+{% if grains['os_family'] == 'Suse' %}
+SUSE_specific_permanent_hostname:
```
Bischoff: Please change name here to something like ```unix_permanent_hostname``` or ```legacy_permanent_hostname```. It is by no way SUSE-specific, AFAIK it goes back to the very beginnings of UNIX and TCP/IP.

salt/default/hostname.sls
```diff
@@ -1,4 +1,43 @@
...
+
+# set the hostname and FQDN name in /etc/hosts
+# this is not needed if a proper DNS server is in place, but when using avahi this
+# might not be the case. The script tries to to use real IP addresses in order not
+# to break round-robin DNS resolution and only uses 127.0.1.1 as a last resort.
+{% if grains['use_avahi'] %}
+hosts_file_hack:
   cmd.script:
-    - name: salt://default/set_hostname.py
+    - name: salt://default/set_ip_in_etc_hosts.py
```
Bischoff: Nice name change. Describes much better what it does.

salt/default/set_ip_in_etc_hosts.py
```diff
@@ -0,0 +1,50 @@
...
+import socket
+import subprocess
+import sys
+
+if len(sys.argv) != 3:
+    print("Usage: set_ip_in_etc_hosts.py <HOSTNAME> <DOMAIN>")
+    sys.exit(1)
+
+_, hostname, domain = sys.argv
+fqdn = hostname + "." + domain
```
Bischoff: Just an academic remark: we abuse the term "FQDN" here. A FQDN should end with a dot ```.``` (it starts at root of DNS tree).

But everyone says the same, so okay.

salt/default/set_ip_in_etc_hosts.py
```diff
@@ -0,0 +1,50 @@
...
+        new_hosts, n = pattern.subn(repl, hosts)
+        if n == 0:
+            new_hosts = hosts + repl
+        f.seek(0)
+        f.truncate()
+        f.write(new_hosts)
+
+update_hosts_file(fqdn, hostname, "")
+ipv4 = guess_address(fqdn, hostname, socket.AF_INET, "127\\.0\\.", "127.0.1.1")
+ipv6 = guess_address(fqdn, hostname, socket.AF_INET6, "(::1$)|(fe[89ab][0-f]:)", "# ipv6 address not found for names:")
```
Bischoff: Useless - and dangerous - now - please comment this line. Or add a systematical comment sign. OR do whatever necessary so it becomes clear we don't declare the IPv6 anymore.

salt/default/hostname.sls
moio: Did not know. Does it then make sense to have it for other distros? If it does not hurt.

salt/default/set_ip_in_etc_hosts.py
moio: We do declare it in case a non-link local address can be found, ie. if DHCPv6 is working correctly. Will document better.

Bischoff:
LGTM (apart from exlcude => exclude typo :smile: )

salt/default/hostname.sls
Bischoff: Probably, yes.

PR #228, opened 2017-11-24
SP3 default for server and proxy on TEST AND HEAD

Bischoff:
SP3 is needed on the proxy HEAD:

```
Problem: nothing provides product(SLES) >= 12.3 needed by product:SUSE-Manager-Proxy-3.2-0.x86_64
```

And while I'm at it, I do the same for TEST.


Bischoff:
nks. Merging the HEAD part for the proxy as you advised. The TEST part can wait. My impression was that TEST was always a variant of HEAD, but I could be wrong.

PR #230, opened 2017-11-25
Support for SLE15

mcalmer:
Initial support for SLE15

- declare image
- define repositories (as soon as they are available in euklid we need to change some URLs)

I was able to setup a traditional client and a minion.

### DIFF
--- a/modules/libvirt/base/variables.tf
+++ b/modules/libvirt/base/variables.tf
@@ -55,7 +55,7 @@ variable "name_prefix" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
-  default = ["centos7",  "opensuse422",  "sles11sp3",  "sles11sp4",  "sles12",   "sles12sp1",  "sles12sp2", "sles12sp3"]
+  default = ["centos7",  "opensuse422",  "sles11sp3",  "sles11sp4",  "sles12",   "sles12sp1",  "sles12sp2", "sles12sp3", "sles15"]
   type = "list"
 }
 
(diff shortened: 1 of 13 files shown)


modules/libvirt/base/variables.tf
```diff
@@ -70,6 +70,7 @@ variable "image_locations" {
     sles12sp1 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp1.x86_64.qcow2"
     sles12sp2 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp2.x86_64.qcow2"
     sles12sp3 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp3.x86_64.qcow2"
+    sles15 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles15b3.x86_64.qcow2"
```
MalloZup: just as a comment, why it is called ```sle15b3``` image? and ont just

```sles15```?

I mean ok, it is beta3, should we maybe just call it ```sles15``` ? we could avoid hardcoded names

mcalmer: It will change several times until release of GM. And sumaform cannot detect changes in images if they do not change the name. @moio told me

MalloZup: ok then for me! maybe we could rename also the image to `sle15beta3` so it's obvious, that we need to change it

mcalmer: This is something we can do. But we need to rebuild the image and it will then become a beta4 now 😀

MalloZup: Ok lol. for me then, @mbologna you can merge it, once mc adress this small issue, from my pov it is ok, you can review it of course

mbologna:
Thanks!

PR #239, opened 2017-12-20
remove gemfile

MalloZup:
@moio after the discussion that i had with @julio, we moved the Gemfile in spacewalk repo (this make more sense when we will opensource spacewalk).

The mechanism in sumaform use that Gemfile, so we don't need here the Gemfile.  

As side effect/advantage, we don't get anymore security issues in sumaform, which were not relevant for the sumaform project anyway. so less noise :japanese_goblin: :rocket:


MalloZup:
we can merge it safely

PR #240, opened 2017-12-20
Mock autoinstallation files

Bischoff:
This PR mocks the contents of installation files

Rationale:
* The test suite does not use the real contents of these files
* The downldoad might fail
* The download takes time


Bischoff:
Tested to work. Merging.

PR #243, opened 2017-12-27
migrate from sles12sp1 to sles12sp3 controller

MalloZup:
Move controller to sles12sp3

- [x] test that works

I think is better to use sles12sp3 at this point that leap42.3. 

Anyway sles12sp3 has ruby 2.1 has default, and we need ruby 2.2, so we need to find out ruby-devel repo

### DIFF
--- a/salt/controller/repos.sls
+++ b/salt/controller/repos.sls
@@ -9,10 +9,10 @@ Devel_Galaxy_cucumber_testsuite_repo:
     - require:
       - sls: default
 
-sle_12_sp1_sdk_pool:
+sle_12_sp3_sdk_pool:
   file.managed:
-    - name: /etc/zypp/repos.d/SLE-12-SP1-SDK-x86_64-Pool.repo
-    - source: salt://controller/repos.d/SLE-12-SP1-SDK-x86_64-Pool.repo
+    - name: /etc/zypp/repos.d/SLE-12-SP3-SDK-x86_64-Pool.repo
+    - source: salt://controller/repos.d/SLE-12-SP3-SDK-x86_64-Pool.repo
     - template: jinja
     - require:
       - sls: default
@@ -22,4 +22,4 @@ refresh_controller_repos:
     - name: zypper --non-interactive --gpg-auto-import-keys refresh
     - require:
       - file: Devel_Galaxy_cucumber_testsuite_repo
-      - file: sle_12_sp1_sdk_pool
+      - file: sle_12_sp3_sdk_pool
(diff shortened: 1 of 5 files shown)


MalloZup:
@moio @mcalmer  do you have any idea where we could get the ´ruby 2.2´  for sles12sp3?

Otherwise we could put the ruby 2.2 or 2.3 in our custom repo, this would make sense if we cannot find any other ruby2.2/3 already in obs

MalloZup:
https://build.suse.de/package/show/Devel:Galaxy:cucumber-testsuite/ruby2.4

found 2.4 from factory. :sunglasses:

MalloZup:
since sles12sp3 image is used in all tf in prod, we can safely migrate. THe test with sles12sp3 works fine, so it is good for review. 

@Bischoff  jfyi

salt/controller/repos.sls
```diff
@@ -9,10 +9,10 @@ Devel_Galaxy_cucumber_testsuite_repo:
...
-sle_12_sp1_sdk_pool:
+sle_12_sp3_sdk_pool:
   file.managed:
-    - name: /etc/zypp/repos.d/SLE-12-SP1-SDK-x86_64-Pool.repo
-    - source: salt://controller/repos.d/SLE-12-SP1-SDK-x86_64-Pool.repo
+    - name: /etc/zypp/repos.d/SLE-12-SP3-SDK-x86_64-Pool.repo
+    - source: salt://controller/repos.d/SLE-12-SP3-SDK-x86_64-Pool.repo
     - template: jinja
     - require:
       - sls: default
```
mcalmer: - Does the controller do not use the "Update" repo?
- Is there any reason for this?
- Is this true for all other repos (SLES) as well?

MalloZup: ciao @mcalmer  nope we didn"t use the update repo, for some pkg we use the sdk repo for all sles repos.

I want to upgrade the ctl from 12sp3 then to sles15 because i want to have the latest gem used on the ctl, which with 12sp1 cannot be (cucumber latest need ruby 2.2)


the 3rd point i didn"t get it :dog: , true for which context?

mcalmer: The 3rd point means: Do we use the SLES Update repo or also only the Pool repo?
I can think we might get some trouble with dependencies if we use SLES-Update with just SDK-Pool

MalloZup: I think we dont use updates for ctl. I will verify tomrw

MalloZup: ok no update

moio:
@MalloZup you have not rebased before merging again. Please check with `gitk` and ping me if needed next time, thanks.

PR #246, opened 2018-01-05
[WIP] Make SSH minion optional

Bischoff:
Like for the proxy, one should not need to create all nodes in order to run the testsuite.

This will allow for more lightwight deployments and test suite runs when debugging.


Bischoff:
Testing...

moio:
Thanks @Bischoff!!!

Issue #247, opened 2018-01-10
Tools repository should not be in the server


ktsamis:
After creating some VM's with sumaform, with no additional repos or packages.
To be clear I used 3.0-released for the server.
I do ```zypper patch``` and I have conflicts as if it is a proxy. I checked the repos and these are enabled:

```
qam-srv:~ # zypper lr
Repository priorities in effect:                                                                                                                                                   (See 'zypper lr -P' for details)
      97 (raised priority)  :  2 repositories
      99 (default priority) :  5 repositories

# | Alias                                  | Name                                   | Enabled | GPG Check | Refresh
--+----------------------------------------+----------------------------------------+---------+-----------+--------
1 | SLE-12-SP2-x86_64-Pool                 | SLE-12-SP2-x86_64-Pool                 | Yes     | (r ) Yes  | No     
2 | SLE-12-SP2-x86_64-Update               | SLE-12-SP2-x86_64-Update               | Yes     | (r ) Yes  | No     
3 | SLE-Manager-Tools-SLE-12-x86_64-Pool   | SLE-Manager-Tools-SLE-12-x86_64-Pool   | Yes     | (r ) Yes  | No     
4 | SLE-Manager-Tools-SLE-12-x86_64-Update | SLE-Manager-Tools-SLE-12-x86_64-Update | Yes     | (r ) Yes  | No     
5 | SUSE-Manager-3.0-x86_64-Pool           | SUSE-Manager-3.0-x86_64-Pool           | Yes     | (r ) Yes  | No     
6 | SUSE-Manager-3.0-x86_64-Update         | SUSE-Manager-3.0-x86_64-Update         | Yes     | (r ) Yes  | No     
7 | home_SilvioMoioli_tools                | home_SilvioMoioli_tools                | Yes     | (r ) Yes  | No     

```
The tools repository should not be in the server. It comes from line 228 in sumaform/salt/default/repos.sls that adds the tools repository and is not in an if jinja statement.

By removing the repos immediately on creation there is no conflict.

So the solution would be either to move the lines for tools in all the other modules that need them, or put an if statement that excludes them for server. I think the second is easier.
Thanks to @Bischoff for helping narrow this down.

ktsamis:
Preparing a pull request that adds a conditional for ```role=!suse_manager_server``` to only then add the repos .

Bischoff:
[...]

PR #248, opened 2018-01-11
Remove client tools repos from srv after deployment

ktsamis:
Just 2 new modules in the standard repos.sls state in suse_manager_server that remove the client tools repositories. I deployed VM's with this state and no errors showed up, all the states were applied successfully, the repos were removed and the double ```zypper patch``` package conflict is not there.
Fixes issue #247

### DIFF
--- a/salt/suse_manager_server/repos.sls
+++ b/salt/suse_manager_server/repos.sls
@@ -179,11 +179,21 @@ filebeat_repo:
 
 remove_client_tools_pool:
   file.absent:
+{% if grains['osmajorrelease'] == '12' %}
     - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-12-x86_64-Pool.repo
+{% endif %}
+{% if grains['osmajorrelease'] == '15' %}
+    - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-15-x86_64-Pool.repo
+{% endif %}
 
 remove_client_tools_update:
   file.absent:
+{% if grains['osmajorrelease'] == '12' %}
     - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-12-x86_64-Update.repo
+{% endif %}
+{% if grains['osmajorrelease'] == '15' %}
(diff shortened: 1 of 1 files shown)


ktsamis:
@moio  and @Bischoff for review whenever you have time :smile:

Bischoff:
@ktsamis You have a "reviewers" button on the right ;-)

ktsamis:
For SLE11 it is irrelevant because there is no server based on it. For SLE15 you are right, I did not see that it was there, I will add this as well today.

ktsamis:
btw @Bischoff I can't edit reviewers or labels or anything, I don't have the right access to the repo.

Bischoff:
Ah right, I had same problem in the old times.

ktsamis:
@Bischoff SLE15 done, please note that it is not tested with a deployment because while we have sle15beta4 image, we don't have 3.2 suse manager yet

Bischoff:
Now seems okay to me. We don't have feedback from Silvio yet, but I'll merge.

Thanks for this first PR @ktsamis ! :champagne:

Bischoff:
It was breaking on head due to the problems with `yaml_jinja`.

The following code lets the installation succeed (but it does not remove the tools repositories) if `yaml_jinja` is broken:
```
{% if grains['osmajorrelease'] == '12' %}
remove_client_tools_pool:
  file.absent:
    - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-12-x86_64-Pool.repo
{% elif grains['osmajorrelease'] == '15' %}
remove_client_tools_pool:
  file.absent:
    - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-15-x86_64-Pool.repo
{% endif %}

{% if grains['osmajorrelease'] == '12' %}
remove_client_tools_update:
  file.absent:
    - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-12-x86_64-Update.repo
{% elif grains['osmajorrelease'] == '15' %}
remove_client_tools_update:
  file.absent:
    - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-15-x86_64-Update.repo
{% endif %}
```

PR #249, opened 2018-01-12
Add version information to all modules in libvirt+testsuite main.tf example

Bischoff:
The "version" variable is apparently needed only on the server.

This PR also includes cosmetical changes: English syntax, "optional" indications.

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -11,16 +11,15 @@ module "base" {
   // network_name = "default" // change to "" if you change bridge below
   // bridge = ""
   // name_prefix = ""
-  // you need always sles12sp1 because controller is based on that.
-  images = ["centos7", "sles12sp1", "sles12sp2", "sles12sp3"]
+  images = ["centos7", "sles12sp3"]
 }
 
 module "head-srv" {
   source = "./modules/libvirt/suse_manager"
   base_configuration = "${module.base.configuration}"
+  version = "head"
   name = "head-srv"
   image = "sles12sp3"
-  version = "head"
   for_development_only = false
   for_testsuite_only = true
   auto_accept = false
@@ -30,6 +29,7 @@ module "head-srv" {
 module "head-cli-sles12sp3" {
   source = "./modules/libvirt/client"
   base_configuration = "${module.base.configuration}"
+  version = "head"
   name = "head-cli-sles12sp3"
   image = "sles12sp3"
   server_configuration =  { hostname = "head-srv.tf.local" }
@@ -54,6 +54,7 @@ module "head-min-sles12sp3" {
 module "head-minssh-sles12sp3" {
   source = "./modules/libvirt/host"
   base_configuration = "${module.base.configuration}"
+  version = "head"
   name = "head-minssh-sles12sp3"
   image = "sles12sp3"
   ssh_key_path = "./salt/controller/id_rsa.pub"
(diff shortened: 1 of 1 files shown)


Bischoff:
WARNING - I am not sure about the removal of this "version" variable. Please double-check that it is only needed on server before accepting this PR.

main.tf.libvirt-testsuite.example
```diff
@@ -11,7 +11,7 @@ module "base" {
   // network_name = "default" // change to "" if you change bridge below
   // bridge = ""
   // name_prefix = ""
-  // you need always sles12sp1 because controller is based on that.
+  // you always need sles12sp1 because the controller is based on that
```
MalloZup: we can remove this comment, since the controller is based on sles12sp3 now

Bischoff: OK.

moio:
> The "version" variable is apparently needed only on the server.

It is also needed on clients/minions to select the SUSE Manager Tools (thus Salt) version they get.

https://github.com/moio/sumaform/blob/master/README_ADVANCED.md#changing-product-versions

Please adapt accordingly

Bischoff:
@moio OK, I am lost. I understand what you say, but:

* `main.tf.libvirt.example` contains no `version` variable
* `main.tf.libvirt-testsuite.example` contains a few ones

Which file is doing correctly?

moio:
The default is to use stable Salt ("`released`"), which is good in most cases. In a testsuite you might have a different need - for example to explicitly use Salt from `HEAD` when running `HEAD` tests.

Bischoff:
The testsuite example is based on `head`, and the non-testsuite example is based on `released`.

@moio, If I understand well, instead of removing `version` variables from the testsuite example, I should add new `version` variables to it to make sure the version is defined on all clients?

If yes, I can modify the PR to do that.

main.tf.libvirt-testsuite.example
```diff
@@ -41,7 +40,6 @@ module "head-cli-sles12sp3" {
 module "head-min-sles12sp3" {
   source = "./modules/libvirt/minion"
   base_configuration = "${module.base.configuration}"
-  version = "head"
```
moio: > The testsuite example is based on head, and the non-testsuite example is based on released.
> [...] in the testsuite, version is defined for the minion and the centos minion, but not for for SSH minion, nor for the traditional client

Since the default is `released`, and this example file is for `head`, I would have expected `version = "head"` on all systems in this example file.

> (Ideally, the default on the minions would be "same version as server"...)

Overridable automatic detection based on `server_configuration` should be possible and would be very welcome as a PR.

Bischoff: > Since the default is released, and this example file is for head, I would have
> expected version = "head" on all systems in this example file.

OK, let's do like that.

Apparently, this variable is not needed for traditional client nor on salt minion, but it does not harm to have it with no reason.

> Overridable automatic detection based on server_configuration should be possible
> and would be very welcome as a PR.

OK, maybe I'll do it one of these days.

Bischoff:
Done. I'll change the PR title, because now the commit does exactly the contrary.

main.tf.libvirt-testsuite.example
Bischoff: There was a mistake here, it was not supposed to be removed from normal minion - sorry.

main.tf.libvirt-testsuite.example
```diff
@@ -11,16 +11,15 @@ module "base" {
   // network_name = "default" // change to "" if you change bridge below
   // bridge = ""
   // name_prefix = ""
-  // you need always sles12sp1 because controller is based on that.
   images = ["centos7", "sles12sp1", "sles12sp2", "sles12sp3"]
```
MalloZup: for improving the example here, i would also remove the sles12sp1, since is not used nowhere (controller is sles12sp3).
and remove also the sles12sp2 var

Bischoff: OK, done.

Please check again to make sure it is in a correct shape now.

moio:
@Bischoff, please merge according to https://github.com/moio/sumaform/pull/251#issuecomment-357605319

If in doubt just ping and I will help!

PR #253, opened 2018-01-19
Feat: make NFS mounts available on ports > 1024

mbologna:


GarySmith:
LGTM

moio:
@mbologna I do not have a problem with this, feel free to merge it, but just for the record what is the use case here?

Thanks!

mbologna:
A client was unable to mount NFS share, installed on a different host but inside the same network.
Network policy was a little bit more restrictive, so NFS couldn't use privileged ports (<1024).
With this patch we allow NFS to use the insecure option, which allows NFS server to use ports that are unprivileged (hence the scary name `insecure`).

PR #254, opened 2018-01-22
Fix some issues with SLE15 and Python3

meaksh:
This PR fixes some issues detected when deploying an SLE15 minion with sumaform:
- Python script incompatibility with Python3 on SLE15: wrong shebang (as /usr/bin/python does not exists) and `filter` error.
- Bad comparison because `osmajorrelease` grain returns an Integer depending on the Salt version.


salt/default/set_ip_in_etc_hosts.py.jinja
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/{{ pythonexec }}
```
moio: Is there really no way to make a cross-SLES working shebang? :open_mouth:

meaksh: AFAIK there is not such common shebang between both Python versions.

According to [PEP394 -- The "python" Command on Unix-Like Systems](http://legacy.python.org/dev/peps/pep-0394/) `/usr/bin/python` is meant for Python 2 and `/usr/bin/python3` should be used for Python 3 version (with some exception like for Arch Linux, which triggered PEP394 creation to establish a recommendation).

Looking at our python2 and python3 packages in OBS, the `/usr/bin/python` file is only provided by the python2 package, while python3 package just provides `/usr/bin/python3`.

MalloZup: :rofl:

moio:
@meaksh feel free to merge as long as you tested this at least one on a pre-SLES15 system. Thanks!

salt/default/hostname.sls
```diff
@@ -41,7 +41,7 @@ hosts_file_hack:
     - args: "{{ grains['hostname'] }} {{ grains['domain'] }}"
     - template: jinja
     - context:
-    {% if grains.get('osmajorrelease', None) == 15 %}
+    {% if grains.get('osmajorrelease', None)|int() == 15 %}
```
mcalmer: Is callint int with bracket  "()" correct? I can see only examples without it.

meaksh: @mcalmer yes, parenthesis are allowed for jinja2 filters:

http://jinja.pocoo.org/docs/2.10/templates/#filters

PR #265, opened 2018-02-08
Fix: mirror: start crontab

mbologna:
Ever wondered why you setup a mirror and mirror script does not run,
even if cronjob time is already passed?
In our kiwi distribution cron.service is not started (but enabled),
so you either start cron (this PR does it) or reboot the machine.


moio:
Nice catch, thanks!

PR #256, opened 2018-01-23
Revert "Remove OpenStack support"

MalloZup:
This reverts commit f04ef656b6933d1f073781dc5d668201494fdf9d.

PR #266, opened 2018-02-09
New OpenStack backend

moio:
This re-adds OpenStack as a backend, aligned with the libvirt one.


Bischoff:
example file: I personally prefer to have the openstack credentials in base module.

apart from that, LGTM.

moio:
Big thanks go out to @MalloZup which worked on #256 which was the basis of this work.

And of course @mbologna who started all this back in Hackweek!

PR #267, opened 2018-02-09
Add centos7_openstack_support via packer

MalloZup:
Add centos7_image for openstack backend.

# Task:

- [x] packer build should pass

- [x] image should work (deploy boot and sshable on openstack)

- [ ]  do final test with working image with ssh enabled via pwd

### DIFF
--- a/modules/openstack/base/variables.tf
+++ b/modules/openstack/base/variables.tf
@@ -42,13 +42,14 @@ variable "name_prefix" {
 
 variable "images" {
   description = "list of images to be uploaded to Glance, leave default for all"
-  default = ["opensuse423",  "sles12sp3",  "sles12sp2",  "sles12sp1",   "sles12",  "sles11sp4", "sles11sp3", "sles15beta4"]
+  default = ["centos7", "opensuse423",  "sles12sp3",  "sles12sp2",  "sles12sp1",   "sles12",  "sles11sp4", "sles11sp3", "sles15beta4"]
   type = "list"
 }
 
 variable "image_locations" {
   description = "list of locations to download images, override to add custom ones"
   default = {
+    centos7 = "http://w3.nue.suse.com/~smoioli/sumaform-images/openstack/centos7.qcow2"
     opensuse423 = "https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/openstack/images/opensuse423.x86_64.qcow2"
     sles11sp3 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images:/OpenStack/images/sles11sp3.x86_64.qcow2"
     sles11sp4 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images:/OpenStack/images/sles11sp4.x86_64.qcow2"
--- a/modules/openstack/host/main.tf
+++ b/modules/openstack/host/main.tf
@@ -37,6 +37,12 @@ resource "openstack_compute_instance_v2" "instance" {
     access_network = true
     name = "fixed"
   }
+
+  user_data = <<EOF
+#cloud-config
+ssh_pwauth: 1
+EOF
+
 }
 
 resource "openstack_blockstorage_volume_v2" "extra_volume" {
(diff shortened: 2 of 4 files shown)


MalloZup:
@Bischoff  if the cloud test went well we are done

Bischoff:
@moio  I just checked, the image for libvirt is indeed of 100 GiB too. Can't we do smaller?

I agree that images should all be of same size in libvirt and openstack. I suppose we even want same images for both, don't we?

Bischoff:
WRT to the SSH problem:

```
Error applying plan:

1 error(s) occurred:

* module.mincentos7.module.minion.null_resource.host_salt_configuration: 1 error(s) occurred:

* ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
```

In `var/log/secure`, polkitd complains of an "Unregistered Authentication agent".

Bischoff:
You are also missing `salt-minion package on the image. epel-release`should be installed first. This package is not needed for `cloud-init`, the comment is wrong.

Bischoff:
There is a problem with packer. It times out at waiting for SSH.

moio:
We need 100 GB on libvirt to accommodate for large cases (RHEL reposyncing). We will be able to reduce this once we use cloud-init for libvirt too (because that can resize disks on first boot).

Cloud images should be as small as possible, 4GB being a good base bet.

Bischoff:
Please see https://github.com/moio/sumaform/pull/269 for a version with many fixes (currently unfinished, but already better).

After fixing the problems with SSH, I stiill get
```
    centos7:   File "/usr/bin/yum", line 28, in <module>
    centos7:     import yummain
    centos7: ImportError: No module named yummain
```

@moio Using 10 GB was a way to get alignement with libvirt images which are currently all 10 GiB. If we want 4 GiB, then we shoud use it everywhere.

mcalmer:
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1       197G   12G  176G   7% /
```
We already use 12G. So maybe 10G is just not enough.

Bischoff:
@mcalmer confusion, stop. I was not speaking about the server.

moio:
> This is just aligning openstack with what we already have in libvirt.

Let me clarify.

In libvirt the image size is defined only in Kiwi/Packer, and it is always 200GB. See:

https://build.suse.de/project/show/Devel:Galaxy:Terraform:Images
https://build.opensuse.org/project/show/systemsmanagement:sumaform:images:libvirt
https://github.com/moio/sumaform/tree/master/packer

This used to be 100GB, it was increased to 200GB due to an edge case. I have now discovered the increase was not correct in all projects, and I fixed it. Please note this is the base image - it makes no difference whether it is then used as a client or server.

In OpenStack we have two sizes: the image size in Kiwi/Packer, which should be big enough to contain the image itself plus a bit of headroom, and the actual disk size at deployment time. In this case, we use 4GB for the former, and the latter is defined by Terraform (currently 10 GB by default on most modules, we can increase it where needed on a module-by-module basis). The resizing from 4GB to 10GB or whatever is configured via Terraform is done by cloud-init on first boot.

To sum it up: I expect new CentOS images for OpenStack defined with 4GB disks in Packer.

One extra thing: I will review the PR as soon as it works and the WIP mark is removed. Ping me if you need help along the way.

Bischoff:
@MalloZup I'm not sure your last commit is useful.

What you need is to add `sshd=1` to packer:
```
      "boot_command": [
        "<tab> sshd=1 text ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/ks_centos7.cfg<enter><wait>"
      ]
```

MalloZup:
@Bischoff  why we need that? 

We need btw more space on disk, i have an error

MalloZup:
bildschirmfoto vom 32018-02-12 11-33-04

MalloZup:
@moio  ok this works fine, salt was fine. i pushed latest change with ssh

MalloZup:
this PR can be merged when pass the review, from my pov this is done and we can do another one for pushing the path for downlad image and update the cenots7 variabel

moio:
Fixed and merged, thanks for your help here!

PR #276, opened 2018-02-14
add openstack testsuite example

MalloZup:


moio:
Was this tested at least once?

MalloZup:
@moio sure this is what i used (the instance are destroyed). works fine

moio:
@MalloZup how much red (rough %) did you see compared to the libvirt version?

MalloZup:
@moio  i didn't see red because of the mirror issue, so i stopped until the deploy not running the cucumber since eric did that, https://w3.suse.de/~ebischoff/output.html. the red is aligned (more and less with the production results)

moio:
Hmm, 66 failures vs. 21 on libvirt over about 620 tests. I guess it's OK.

Will merge this after a couple of editorial fixes.

MalloZup:
i am scheduling a next run to have a full picture

moio:
@MalloZup, I am merging this without proxy support.

Reasoning is the same given back when we decided not to add proxies to the testsuite libvirt example, documenting the possibility in `README_ADVANCED.md` instead, namely:

1- we want to make running the testsuite as convenient as possible, in order to encourage people using it. Adding a Proxy requires extra resources (arguably this is more important in the libvirt case, but still)

2- we already have Proxy coverage in the automated testsuite - ironically we do not cover the direct case well at this point in time. At least we get some "cheap coverage" from developers this way.

Side note: ideally the testsuite could cover both, re-running relevant feature sets with and without the proxy. Maybe something for the future.

moio:
Thanks!

PR #277, opened 2018-02-14
fix mirror cucumber_repo

MalloZup:
jfyi


MalloZup:
@moio  do we need something else ? i am not the mirror expert

moio:
Fixed, squashed and merged. Thanks

PR #279, opened 2018-02-19
Fix: openSUSE mirror

mbologna:
Current mirror URL for openSUSE does not have a `repodata` directory, and minima will fail.


moio:
Nice catch! Cool that everything worked despite of this bug :smile:

PR #283, opened 2018-02-23
add python-pip trusted-host flag

MalloZup:
this is needed for cloud vms

We can use pip and downlad pkg on cloud


moio:
Fixed via 78fc1f4, thanks anyway!

MalloZup:
yop thx

PR #284, opened 2018-02-26
Prefer job scraping instead of global one

MalloZup:
for better integration with other jobs


MalloZup:
i'm searching if the "global" attribute is needed

MalloZup:
i will test this pr with a grafana deploy

PR #287, opened 2018-02-27
Split macro-settings into fine-grained settings

moio:
This PR:
 * drops support for SUSE Manager 2.1
 * drops support for the Oracle database (which was only in 2.1)
 * drops support for SLES 11 SP3
 * drops support for the pgpool backend
 * splits `for_development_only` in narrow settings for clients, minions, proxies and SUSE Manager Servers
 * updates sample files accordingly
 * updates Travis CI tests accordingly


moio:
Guys, I am aware this is somehow a monster PR, but still you should take at least a superficial look to see if anything can be done better.

I am testing it in these days and plan to merge it before the end of the week.

TIA

MalloZup:
@moio   add a task: 

- [ ] update all production main.tf, with the new flag ``` testsuite``` instead of ```for_testsuite_only```.

From my pov the old_name is more clear then only ```testsuite```, which is much more generic and not direct, also is compatible backward.

From my first POV, we should conserver the old_name. For the other points, i will review asap

main.tf.openstack-testsuite.example
```diff
@@ -47,9 +48,18 @@ module "srv" {
   version = "3.1-nightly"
   name = "srv"
   image = "sles12sp3"
-  for_development_only = false
-  for_testsuite_only = true
   auto_accept = false
+  disable_firewall = false
+  allow_postgres_connections = false
+  java_debugging = false
```
MalloZup: nice this one, i think we can monitor full without dev flag now!

README_ADVANCED.md
```diff
@@ -62,6 +60,35 @@ module "minionsles12sp1" {
...
+## Turning convenience features off
+
+By default, sumaform deploys hosts with a range of tweaked settings for convenience reasons. If in your use case this is not wanted, you can turn those off via the following variables.
+
+ * `client` module:
+   * `auto_register`: automatically registers clients to the SUSE Manager Server. Set to `false` for manual registration
+ * `minion` module:
+   * `auto_connect_to_master`: automatically connects to the Salt Master. Set to `false` to manually configure
+ * `proxy` module:
+   * `auto_register`: automatically registers the proxy to the SUSE Manager Server. Set to `false` for manual registration
```
MalloZup: i  would vote for ```auto_register``` set by default to false. 
To me seems an invasive default change, meaning that by default a minion is bootstrapped against a server

.ci-travis/salt-client-validation
```diff
@@ -16,8 +16,8 @@ version: head
 server: travis.tf.local
 role: client
 mirror: null
-for_development_only: 0
-for_testsuite_only: 1
```
MalloZup: i don't see any reason for changing the name and break the api

README_ADVANCED.md
```diff
@@ -62,6 +60,35 @@ module "minionsles12sp1" {
...
 This will create 10 minions connected to the `suma3pg` server.
 
+## Turning convenience features off
+
+By default, sumaform deploys hosts with a range of tweaked settings for convenience reasons. If in your use case this is not wanted, you can turn those off via the following variables.
+
+ * `client` module:
+   * `auto_register`: automatically registers clients to the SUSE Manager Server. Set to `false` for manual registration
+ * `minion` module:
+   * `auto_connect_to_master`: automatically connects to the Salt Master. Set to `false` to manually configure
```
MalloZup: same for auto_register: set it by default to false

MalloZup: auto_connect and auto_register need to be false

Bischoff:
I prefer new name `testsuite`.

moio:
I am not breaking backwards compatibility with `for_development_only` here, so `auto_register` and `auto_connect_to_master` must stay `true`. Countless developer hours have already been saved by this and the whole point of sumaform has always been to reduce people's wasted time, so this is not going to change, sorry.

`for_testsuite_only` I do not like because of the `only` - this is not the case any more, and this is the whole point of this refactoring. It made sense when this was mutually exclusive with `for_development_only`. If you have other proposals than `testsuite` I am happy to evaluate them, though.

moio:
> For the testsuite example file, do we really have so many settings that differ from the default, @moio ? I would only specify the ones that differ...

This should already be the case now.

MalloZup:
@moio before merging this, can you reference here a Pull-request against gitlab repo? so we can review it  and sync better, thx

moio:
Will do

moio:
https://gitlab.suse.de/galaxy/sumaform-test-runner/merge_requests/24

Bischoff:
I did not see problems with the gitlab requests. Do we have a way to test it before, or it's just "jump into the swimming pool and start swimming" ?

moio:
Well, I did test these changes with a similar `main.tf` locally, I do not think we can do any better.

MalloZup:
if it works locally should work on production, in any worst case we can revert the commit

PR #290, opened 2018-02-28
Testsuite: clarify need for credentials

moio:
We had users stumbling upon the problem that creating the controller "takes forever" - git waits for user input.

I made the credentials passwords mandatory, so they can't be skipped at plan/apply time.

Additionally, since this does not work for users with 2FA, I added a link to where users can find the credentials we use for CI.


MalloZup:
when its opensource you don't need credentials (git clone is possible on public)

Bischoff:
ok

PR #294, opened 2018-03-07
Fix content staging window

Bischoff:
[-6, -3] minutes before installation => [-3, -1] minutes
0.05 hour = 3 minutes
0.033 hour = 2 minutes

@moio FYI - just different timings in a test suite configuration file

### DIFF
--- a/salt/suse_manager_server/testsuite.sls
+++ b/salt/suse_manager_server/testsuite.sls
@@ -99,7 +99,7 @@ enable_salt_content_staging_window:
   file.replace:
     - name: /usr/share/rhn/config-defaults/rhn_java.conf
     - pattern: 'salt_content_staging_window = (.*)'
-    - repl: 'salt_content_staging_window = 0.05'
+    - repl: 'salt_content_staging_window = 0.033'
     - require:
       - pkg: suse_manager_packages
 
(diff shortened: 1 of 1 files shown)


salt/suse_manager_server/testsuite.sls
```diff
@@ -99,15 +99,15 @@ enable_salt_content_staging_window:
...
+    - repl: 'salt_content_staging_window = 0.033'
     - require:
       - pkg: suse_manager_packages
 
 enable_salt_content_staging_advance:
   file.replace:
     - name: /usr/share/rhn/config-defaults/rhn_java.conf
     - pattern: 'salt_content_staging_advance = (.*)'
-    - repl: 'salt_content_staging_advance = 0.1'
+    - repl: 'salt_content_staging_advance = 0.05'
```
mcalmer: From my point of view we should start to do it right? Changing "default " values is not a good thing.
Why not add `java.XXX` to rhn.conf like customers should do it?

moio: Right, addressed in 0722fc4

Bischoff: Thanks for the fix @moio.

BTW, @mcalmer, just for curiosity, isn't this setting available by UI? For test suite, it would be more comfortable to change it this way...

moio: @Bischoff that was on the roadmap but we never got there. Not many customers currently seem interested.

Bischoff: ok @moio. I was suspecting it was so.

PR #305, opened 2018-03-21
Add apparmor access control facility

Bischoff:
This adds the possibility to optionally install `apparmor` and `auditd` on the server, the proxy, and the minion.

For now, the apparmor profiles are stored inside sumaform. I suppose we will package them for the customers at some point, and then we will probably remove them from sumaform and install that new package. But we're not there yet.


moio:
Will review as soon as the `WIP` tag is removed.

Bischoff:
Thanks @moio . I'm hoping to be able to finalize by the end of the day.

salt/minion/apparmor.sls
```diff
@@ -0,0 +1,34 @@
+{% if grains.get('apparmor') %}
+
+apparmor_packages:
+  pkg.installed:
+    - pkgs:
+      - apparmor-parser
```
moio: Is this package list OK across SLES versions (11-12-15)? What about CentOS?

Bischoff: Good catch. It will certainly not work on CentOS. I'll check on sle 11. I'll fix that.

I'll also fix capitalization of "AppArmor" in visible texts.

README_ADVANCED.md
```diff
@@ -446,6 +446,39 @@ Grafana is accessible at http://grafana.tf.local with username and password `adm
 
 Please note for the Java probes to work the `java_debugging` setting has to be enabled in the `suse_manager_server` module (it is by default).
 
+
+## AppArmor access control
+
+Optional access control via AppArmor and Audit Daemon can be enforced on a SUSE Manager server, proxy, or SLE minion. A libvirt example follows:
```
moio: openSUSE minions should work, right? In that case just use "SUSE minion" here

salt/minion/apparmor.sls
moio: Please guard this file with a Jinja condition on the OS, eg.

```jinja
...
{% endif %}
```

salt/suse_manager_proxy/apparmor.sls
```diff
@@ -0,0 +1,51 @@
+{% if grains.get('apparmor') %}
```
moio: Please also add the OS condition here

salt/suse_manager_server/apparmor.sls
```diff
@@ -0,0 +1,124 @@
+{% if grains.get('apparmor') %}
```
moio: Please also add the OS condition here

Bischoff:
Requested changes done.

moio:
Thanks for this nice contribution!

PR #307, opened 2018-03-22
Remove useless variable from examples

ncounter:
`version` variable is useless for clients, it matters only for SUSE Manager server and proxy.


moio:
> Some modules have a version variable that determines the software version. Specifically:
>  * in suse_manager, suse_manager_proxy etc. version determines the SUSE Manager product version,
> * in minion, client, etc. version determines the SUSE Manager Tools version.

https://github.com/moio/sumaform/blob/master/README_ADVANCED.md#changing-product-versions

ncounter:
Ah, I overlooked that. Thank you @moio

PR #310, opened 2018-03-25
configure grain for use_unreleased_updates in host

mcalmer:
To be able to define "use_unreleased_updates" also for ssh minions we need to configure the grain
in "host"


mcalmer:
@moio The testsuite failed so I pushed a quickfix directly. This PR should make the fix better and work for ssh minions as well.

moio:
Thank you very much for taking the time to address this!

PR #313, opened 2018-04-03
Add support for 3.2-released

Bischoff:
This PR adds support for `3.2-released`, currently defined as latest beta.

Notes:
* the server and proxy pool repository URLs are hardcoded for SLE 12 SP3, but now we also support openSUSE 42.3. I did not find a beta repository for 42.3 though;
* I had to add the development tools repository, otherwise it would not install fine (apache2 was not starting).
* I did not add the other development repositories.

It works with this setup, but please check whether this is correct. Honestly, I'm a bit lost in this repositories jungle.

### DIFF
--- a/salt/default/repos.sls
+++ b/salt/default/repos.sls
@@ -74,7 +74,15 @@ tools_update_repo:
   file.touch:
     - name: /tmp/no_tools_update_repo_needed
 
-{% if 'released' in grains.get('version') | default('released', true) %}
+{% if '3.2-released' in grains.get('version') | default('', true) %}
+
+tools_devel_repo:
+  file.managed:
+    - name: /etc/zypp/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-11-x86_64.repo
+    - source: salt://default/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-11-x86_64.repo
+    - template: jinja
+
+{% elif 'released' in grains.get('version') | default('released', true) %}
 
 tools_devel_repo:
   file.touch:
@@ -211,7 +219,15 @@ tools_update_repo:
     - source: salt://default/repos.d/SLE-Manager-Tools-SLE-12-x86_64-Update.repo
     - template: jinja
 
-{% if 'released' in grains.get('version') | default('released', true) %}
+{% if '3.2-released' in grains.get('version') | default('', true) %}
+
+tools_devel_repo:
+  file.managed:
+    - name: /etc/zypp/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-12-x86_64.repo
+    - source: salt://default/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-12-x86_64.repo
+    - template: jinja
+
+{% elif 'released' in grains.get('version') | default('released', true) %}
 
 tools_devel_repo:
   file.touch:
@@ -361,7 +377,16 @@ suse_res7_key:
     - watch:
       - file: suse_res7_key
 
-{% if 'head' in grains.get('version') | default('', true) %}
+{% if '3.2-released' in grains.get('version') | default('', true) %}
+tools_update_repo:
+  file.managed:
+    - name: /etc/yum.repos.d/SLE-Manager-Tools-RES-7-Beta-x86_64.repo
+    - source: salt://default/repos.d/SLE-Manager-Tools-RES-7-Beta-x86_64.repo
+    - template: jinja
+    - require:
+      - cmd: galaxy_key
+
+{% elif 'head' in grains.get('version') | default('', true) %}
 tools_update_repo:
   file.managed:
     - name: /etc/yum.repos.d/Devel_Galaxy_Manager_Head_RES-Manager-Tools-7-x86_64.repo
@@ -369,6 +394,7 @@ tools_update_repo:
     - template: jinja
     - require:
       - cmd: galaxy_key
+
 {% elif '3.0-nightly' in grains.get('version') | default('', true) %}
 tools_update_repo:
   file.managed:
@@ -377,6 +403,7 @@ tools_update_repo:
     - template: jinja
     - require:
       - cmd: galaxy_key
+
 {% elif 'nightly' in grains.get('version') | default('', true) %}
 tools_update_repo:
   file.managed:
--- a/salt/default/repos.sls
+++ b/salt/default/repos.sls
@@ -64,24 +64,6 @@ test_update_repo:
     - template: jinja
 {% endif %}
 
-{% if '3.2-released' in grains.get('version') | default('', true) %}
-
-tools_pool_repo:
-  file.touch:
-    - name: /tmp/no_tools_pool_repo_needed
-
-tools_update_repo:
-  file.touch:
-    - name: /tmp/no_tools_update_repo_needed
-
-tools_devel_repo:
-  file.managed:
-    - name: /etc/zypp/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-11-x86_64.repo
-    - source: salt://default/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-11-x86_64.repo
-    - template: jinja
-
-{% elif 'released' in grains.get('version') | default('released', true) %}
-
 tools_pool_repo:
   file.managed:
     - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-11-x86_64.repo
@@ -92,19 +74,21 @@ tools_update_repo:
   file.touch:
     - name: /tmp/no_tools_update_repo_needed
 
+{% if '3.2-released' in grains.get('version') | default('', true) %}
+
+tools_devel_repo:
+  file.managed:
+    - name: /etc/zypp/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-11-x86_64.repo
+    - source: salt://default/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-11-x86_64.repo
+    - template: jinja
+
+{% elif 'released' in grains.get('version') | default('released', true) %}
+
 tools_devel_repo:
   file.touch:
     - name: /tmp/no_tools_devel_repo_needed
 
 {% elif '3.0-nightly' in grains.get('version') | default('', true) %}
-
-tools_pool_repo:
-  file.touch:
-    - name: /tmp/no_tools_pool_repo_needed
-
-tools_update_repo:
-  file.touch:
-    - name: /tmp/no_tools_update_repo_needed
 
 tools_devel_repo:
   file.managed:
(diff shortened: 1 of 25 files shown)


salt/default/repos.sls
```diff
@@ -211,7 +211,15 @@ tools_update_repo:
     - source: salt://default/repos.d/SLE-Manager-Tools-SLE-12-x86_64-Update.repo
     - template: jinja
 
-{% if 'released' in grains.get('version') | default('released', true) %}
+{% if '3.2-released' in grains.get('version') | default('', true) %}
+
+tools_devel_repo:
```
moio: @juliogonzalez do we have a 3.2 beta Client Tools repo URL as of today?

salt/suse_manager_proxy/repos.d/SUSE-Manager-Proxy-3.2-x86_64-Pool.repo
```diff
@@ -0,0 +1,11 @@
+[SUSE-Manager-Proxy-3.2-x86_64-Pool]
+name=SUSE-Manager-Proxy-3.2-x86_64-Pool
+type=rpm-md
+enabled=1
+baseurl=http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/SUSE:/SLE-12-SP3:/Update:/Products:/Manager32/images/repo/SUSE-Manager-Proxy-3.2-POOL-x86_64-Media1/
+{#
+   currently not up to date:
```
moio: Well, `3.2-released` at this point should contain 3.2 beta 1. This seems to actually be a repo for 3.2-nightly, since beta 2 has not been released yet.

Is this what you mean when you say that the euklid repo is not up to date?

salt/default/repos.sls
Bischoff: Apparently, not. I'll ask again to Julio.

Clearly, it does not work if I use "no tools needed". But
* I don't understand why...
* I'm not sure I should not use `99` priority
* I'm not sure I should not use other  development repos

salt/suse_manager_proxy/repos.d/SUSE-Manager-Proxy-3.2-x86_64-Pool.repo
Bischoff: Ah. Yes, I need to test beta 2 despite it has not been released yet.

Should I rephrase it all and call it a `3.2-nightly` ? This would explain better the development repos thing.

Or should I moderate my ambitions, and support only beta 2 once it has been released ?

Or do both, a real `3.2-nightly`and a real `3.2-released ?

Bischoff:
See https://w3.suse.de/~ebischoff/output.html , the 8 errors there are old, already fixed errors, suggesting the repos are not good yet.

Also, you're making a point, I did a kind of `3.2-nightly`... This whole thing needs more discussion, I'll re-add the "WIP" tag as it's far from ready...

salt/default/repos.sls
moio: @juliogonzalez they need to add `Repotype: staticlinks` in the Build Service project, that will create symlinks without the build number.

moio: > it does not work if I use "no tools needed". But I don't understand why...

One likely explanation is the Salt version. Because of [how images are designed in sumaform](https://github.com/moio/sumaform/blob/b260ce152b30d615039b7c4d075137586891b6ed/DESIGN.md#requirements-for-base-disk-images), they seldom have the right Salt version (in this case, they actually ship an ancient version, very far away from the version shipped with Server 3.2 beta). So we need to upgrade Salt with Salt, and this is known to be tricky.

Specifically upgrading Salt with Salt typically breaks the current Salt call at the moment of the upgrade, so the recommended workaround is to do that as the very last step of the state. This is what we do when deploying: [we run salt-call twice](https://github.com/moio/sumaform/blob/b260ce152b30d615039b7c4d075137586891b6ed/modules/libvirt/host/main.tf#L73-L74) and the first one is mostly to update Salt (in addition to a handful other very basic settings that we can't live without).

This is the reason why we need the correct repo with the correct Salt version in `default`, very early in the deployment phase - so that the second salt-call does not end up upgrading Salt with Salt and break.

Back to this case: in order to have the same Salt as Manager Server 3.2 beta, we need to have the matching Client Tools repo here.

We can do that as soon as @juliogonzalez talks with Benni and fixes the repo URLs so that they do not change any more.

That should solve your problem - but even if it does not, this is needed.

> I'm not sure I should not use 99 priority

Priority should be consistent with other Client Tools repos - default priority for the "released" repo and 98 for the respective Devel repo (to be layered on top of the released one, only for "nightly").

I'm not sure I should not use other development repos

> I'm not sure I should not use other development repos

Not for the "released" version.

salt/suse_manager_proxy/repos.d/SUSE-Manager-Proxy-3.2-x86_64-Pool.repo
moio: Please keep this PR for `3.2-released`, which we partially need now (beta 1) and will definitely need in a few days (beta 2).

I am open to reviewing a separate PR for `3.2-nightly`, although that should not have a real meaning before 3.2 GA - in the meantime `head` should be used, since it's rolling anyway. Am I missing something?

Bischoff: ok

Bischoff:
According to Michael, current code is good, but we cannot merge it before beta 3 because of packaging problems that prevent it from working correctly.

I leave it open and will retest when beta 3 is out.

salt/default/repos.sls
juliogonzalez: ```Repotype: staticlinks``` added.

@moio, @Bischoff, you already have the repositories without build numbers at http://download.suse.de/ibs/SUSE:/SLE-12:/Update:/Products:/ManagerToolsBeta/images/repo/

Bischoff:
Adapted to beta 3, and tested. Seems to work fine.

salt/default/repos.d/Devel_Galaxy_Manager_3.2_RES-Manager-Tools-7-x86_64.repo
```diff
@@ -0,0 +1,6 @@
+[Devel_Galaxy_Manager_3.2_RES-Manager-Tools-7-x86_64]
```
moio: JFI: I expect this file to be removed after 3.2 GA

salt/default/repos.d/Devel_Galaxy_Manager_3.2_RES-Manager-Tools-7-x86_64.repo
```diff
@@ -0,0 +1,6 @@
...
+name=Devel_Galaxy_Manager_3.2_RES-Manager-Tools-7-x86_64
```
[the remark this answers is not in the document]
Bischoff: Makes sense, ok.

salt/default/repos.sls
```diff
@@ -74,7 +74,15 @@ tools_update_repo:
   file.touch:
     - name: /tmp/no_tools_update_repo_needed
 
-{% if 'released' in grains.get('version') | default('released', true) %}
+{% if '3.2-released' in grains.get('version') | default('', true) %}
+
+tools_devel_repo:
```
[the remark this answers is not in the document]
Bischoff: I am now sure one must either take the beta repos, or the normal update repos, but not both. You're right, I have to fix this.

salt/mirror/minima.yaml
```diff
@@ -215,6 +225,21 @@
   path: /srv/mirror/ibs/Devel:/Galaxy:/Manager:/3.1:/RES7-SUSE-Manager-Tools/SUSE_RES-7_Update_standard
   archs: [x86_64]
 
+# SLE 11 SP4 Manager Tools 3.2 devel
+- url: http://beta.suse.com/private/SUSE-Manager-beta/Beta3/SLE11SP4/x86_64
+  path: /srv/mirror/private/SUSE-Manager-beta/Beta3/SLE11SP4/x86_64
+  archs: [x86_64]
+
+# SLE 12 (GA and all SPs) Manager Tools 3.2 devel
+- url: http://download.suse.de/ibs/SUSE:/SLE-12:/Update:/Products:/ManagerToolsBeta:/Update/standard
```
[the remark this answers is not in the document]
Bischoff: Because beta.suse.com ones did not exist at the time i developed this :fox_face: . I'll fix that.

salt/suse_manager_proxy/repos.sls
```diff
@@ -37,6 +37,20 @@ suse_manager_proxy_update_repo:
...
+  file.managed:
+    - name: /etc/zypp/repos.d/SUSE-Manager-Proxy-3.2-x86_64-Pool.repo
+    - source: salt://suse_manager_proxy/repos.d/SUSE-Manager-Proxy-3.2-x86_64-Pool.repo
+    - template: jinja
+    - require:
+      - sls: default
+
+suse_manager_proxy_update_repo:
+  file.touch:
+    - name: /tmp/no_update_channel_needed
```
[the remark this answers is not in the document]
Bischoff: Sure.

salt/suse_manager_server/repos.sls
```diff
@@ -37,6 +37,20 @@ suse_manager_update_repo:
...
+suse_manager_pool_repo:
+  file.managed:
+    - name: /etc/zypp/repos.d/SUSE-Manager-3.2-x86_64-Pool.repo
+    - source: salt://suse_manager_server/repos.d/SUSE-Manager-3.2-x86_64-Pool.repo
+    - template: jinja
+    - require:
+      - sls: default
+
+suse_manager_update_repo:
+  file.touch:
```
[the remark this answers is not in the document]
Bischoff: Same.

salt/default/repos.sls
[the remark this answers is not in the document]
moio: @mcalmer does that mean that beta customers really need to have them layered on top of normal pool/update repos?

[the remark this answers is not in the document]
Bischoff: When I was testing manually, I had to use the beta tools repo but not the pool / update ones, otherwise i could not reproduce Ornela's problems with rhncfg* packages. That was for SLE12SP3 client. This is what makes me think we should not pile devel tool repos on top of pool/update tool repos.

Anyway... I did the related restructuring in a separate commit. So if that was wrong, it's enough to destroy that last commit.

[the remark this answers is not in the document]
moio: According to an extended discussion with @mcalmer those repos are indeed meant to be stacked after all. This will change for version 12 and on starting with the next Manager major release - initial groundwork was laid to have it supported by 3.2 but that actually was not fully completed in time.

I am fixing this directly in the PR.

moio:
@Bischoff this is OK FMPOV now. Please test (3.2 clients only) and then feel free to merge. You can also rewrite history a bit before merging if you like.

Thank you very much for this contribution!

Bischoff:
Testing latest version right now...

PR #317, opened 2018-04-05
update to sles15rc2

lucidd:


modules/openstack/base/variables.tf
```diff
@@ -61,7 +61,7 @@ variable "image_locations" {
   default = {
     centos7 = "http://w3.nue.suse.com/~smoioli/sumaform-images/openstack/centos7.qcow2"
     opensuse423 = "https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/openstack/images/opensuse423.x86_64.qcow2"
-    sles15beta4 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images:/OpenStack/images/sles15beta4.x86_64.qcow2"
+    sles15rc2 = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images:/OpenStack/images/sles15rc2.x86_64.qcow2"
```
moio: This returns 404, is it expected?

MalloZup: yop i think the github code is ok, but i would assume that the image need to be accepted with a SR.

feel free to post here the submission link so we can have a look on it

mcalmer: Ahh, you have an extra project for `OpenStack`. AFAIK @lucidd only changed the normal image.

lucidd: should be fixed now

moio:
Thanks!

PR #322, opened 2018-04-17
WIP: PTS module

chiaradiamarcelo:


modules/libvirt/pts/variables.tf
```diff
@@ -0,0 +1,204 @@
...
+  description = "use ${module.base.configuration}, see the main.tf example file"
+  type = "map"
+}
+
+variable "name" {
+  description = "hostname, without the domain part"
+  type = "string"
+}
+
+variable "version" {
```
moio: Please only export variables that are really used in `pts/main.tf`.

modules/libvirt/pts/variables.tf
```diff
@@ -0,0 +1,204 @@
+variable "base_configuration" {
+  description = "use ${module.base.configuration}, see the main.tf example file"
+  type = "map"
+}
+
+variable "name" {
```
moio: We need to be flexible here and allow users to define names and MACs for each host we define in `main.tf`. This is the only possible way libvirt users can get proper DNS names if they so want.

modules/libvirt/pts/main.tf
```diff
@@ -0,0 +1,25 @@
+module "server" {
+  source = "../suse_manager"
+  name = "${var.name}-server"
+  base_configuration = "${var.base_configuration}"
+  version = "${var.version}"
+  monitored = true
+  use_unreleased_updates = true
+  pts = true
```
moio: Please make sure all flags in the original `performance/main.tf` are here. I am missing `channels` and `cloned_channels` at the very least, but please check them all.

modules/libvirt/pts/main.tf
```diff
@@ -0,0 +1,25 @@
...
+  name = "${var.name}-evil-minions"
+  server_configuration = "${module.server.configuration}"
+  dump_file = "modules/libvirt/pts/minion-dump.mp"
+}
+
+module "locust" {
+  source = "../locust"
+  source = "../grafana"
+  base_configuration = "${var.base_configuration}"
+  server_configuration = "${module.server.configuration}"
```
moio: `dump_file` seems missing here. The double `source` is also certainly wrong.

modules/libvirt/suse_manager/variables.tf
```diff
@@ -187,6 +187,11 @@ variable "vcpu" {
   default = 2
 }
 
+variable "pts" {
+  description = "Wheter this SUMA server instance is meant for running the PTS"
```
moio: Grammar nitpick: *whether*

main.tf.pts.libvirt.example
```diff
@@ -0,0 +1,30 @@
...
+ images = ["centos7", "sles12sp1", "sles12sp2", "sles12sp3", "opensuse423"]
+}
+
+module "pts" {
+  source = "./modules/libvirt/pts"
+  base_configuration = "${module.base.configuration}"
+  //hostnames
+  suma_server_name = "server"
+  evil_minions_host_name = "evil-minions"
+  //mac addresses
```
moio: Those can be omitted for a default example file. Advanced users will check variables.tf

main.tf.pts.libvirt.example
```diff
@@ -0,0 +1,30 @@
...
+  // name_prefix = ""
+  // timezone = "Europe/Berlin"
+ images = ["centos7", "sles12sp1", "sles12sp2", "sles12sp3", "opensuse423"]
+}
+
+module "pts" {
+  source = "./modules/libvirt/pts"
+  base_configuration = "${module.base.configuration}"
+  //hostnames
+  suma_server_name = "server"
```
moio: Please provide and use default names by default

modules/libvirt/pts/main.tf
```diff
@@ -0,0 +1,29 @@
...
+  source = "../suse_manager"
+  name = "${var.suma_server_name}"
+  base_configuration = "${var.base_configuration}"
+  version = "3.1-nightly"
+  monitored = true
+  use_unreleased_updates = true
+  pts = true
+  mac = "${var.suma_server_mac}"
+  channels = ["sles12-sp3-pool-x86_64", "sles12-sp3-updates-x86_64", "sle-manager-tools12-pool-x86_64-sp3", "sle-manager-tools12-updates-x86_64-sp3"]
+  cloned_channels = "[{channels: [sles12-sp3-pool-x86_64, sles12-sp3-updates-x86_64, sle-manager-tools12-pool-x86_64-sp3, sle-manager-tools12-updates-x86_64-sp3], prefix: cloned-2017-q3, date: 2017-09-30}, {channels: [sles12-sp3-pool-x86_64, sles12-sp3-updates-x86_64, sle-manager-tools12-pool-x86_64-sp3, sle-manager-tools12-updates-x86_64-sp3], prefix: cloned-2017-q4, date: 2017-12-31}]"
```
moio: Please double check defaults for RAM and vCPUs - they should more or less match what we use in OpenStack

modules/libvirt/pts/main.tf
```diff
@@ -0,0 +1,29 @@
...
+
+  name = "${var.evil_minions_host_name}"
+  server_configuration = "${module.server.configuration}"
+  dump_file = "modules/libvirt/pts/minion-dump.mp"
+  mac = "${var.evil_minions_host_mac}"
+}
+
+module "locust" {
+  source = "../locust"
+  base_configuration = "${var.base_configuration}"
```
moio: name missing here

modules/libvirt/pts/variables.tf
```diff
@@ -0,0 +1,31 @@
...
+  type = "map"
+}
+
+variable "suma_server_name" {
+  description = "hostname, without the domain part"
+  type = "string"
+  default = "pts-server"
+}
+
+variable "evil_minions_host_name" {
```
moio: Please be consistent with variable names, and use <MODULE_NAME>_name in all cases.

modules/libvirt/pts/variables.tf
```diff
@@ -0,0 +1,31 @@
...
+  description = "hostname for the evil-minions host, without the domain part"
+  type = "string"
+  default = "pts-evil-minions"
+}
+
+variable "suma_server_mac" {
```
moio: Same here, <MODULE_NAME>_mac

modules/libvirt/pts/main.tf
```diff
@@ -0,0 +1,45 @@
...
+  memory = 4096
+}
+
+module "locust" {
+  source = "../locust"
+  name = "${var.locust_name}"
+  base_configuration = "${var.base_configuration}"
+  server_configuration = "${module.server.configuration}"
+  mac = "${var.locust_mac}"
+  memory = 2048
```
moio: In this case only, I would change the default locust module memory value to this value, removing this parameter from here. I do not think to make sense to default to a value that is not safe.

modules/libvirt/pts/variables.tf
```diff
@@ -0,0 +1,53 @@
...
+
+variable "monitored" {
+  description = "whether the suse manager server should be monitored via Prometheus and Grafana"
+  default = false
+}
+
+variable "suse_manager_name" {
+  description = "hostname, without the domain part"
+  type = "string"
+  default = "pts-server"
```
moio: I think defaults should not contain the `pts` prefix - that should be set in `base` (if needed at all).

salt/suse_manager_server/pts/run-pts.py
```diff
@@ -0,0 +1,105 @@
...
+    global locust_url
+    global patching_only
+    global locust_only
+    global run_all
+
+    for opt, arg in options:
+        if opt in ('--evil-minions-hostname'):
+            evil_minion_hostname = arg
+        elif opt in ('--locust-hostname'):
+            locust_url = 'http://' + arg + '.tf.local'
```
moio: We cannot assume avahi is available. Please make sure that the full hostname is passed, so that it's not necessarily a .tf.local name

chiaradiamarcelo:
looks good to me

moio:
Added some changes and merged to try in Jenkins. Please review!

PR #324, opened 2018-04-18
Add name variable for locust module

chiaradiamarcelo:


moio:
Thanks for this contribution!

PR #327, opened 2018-04-20
Locust task for downloading packages from SUMA server

chiaradiamarcelo:


salt/locust/locustfile.py
```diff
@@ -9,16 +11,31 @@ def on_start(self):
         # don't verify ssl certs
         self.client.verify = False
         self.login()
```
moio: Can be removed, the login is just for the Web UI

salt/locust/locustfile.py
```diff
@@ -9,16 +11,31 @@ def on_start(self):
...
+        self.get_download_package_url()
```
moio: I would probably use:

```python
self.download_path = self.get_download_path()
```

To make the declaration of the field more visible (style only). As to why `path` instead of `url`, see below.

salt/locust/locustfile.py
```diff
@@ -9,16 +11,31 @@ def on_start(self):
...
 
     def login(self):
         self.client.post("/", {"username": os.environ["SERVER_USERNAME"], "password": os.environ["SERVER_PASSWORD"]})
 
+    def get_download_package_url(self): 
+        channel_name = 'cloned-2017-q4-sles12-sp3-updates-x86_64'
+        package_name = 'at-3.1.14-8.3.1.x86_64.rpm'
+
+        try:
+            response = urllib2.urlopen('http://server.tf.local/pub/pillar_data/pillar_evil-minions-0.yml')
```
moio: Server's FQDN should not be hardcoded. You can just remove it, as `self.client.get("/path/not/full/url")` takes care of injecting the server name.

PR #340, opened 2018-05-23
[WIP] Adding new role: SSH minion

Bischoff:
"SSH minion" clients were previously managed with the generic "host" module. Since we want to add test repositories when we are using them from the testsuite, this is not enough anymore.

This PR adds the following:
* a new `ssh_minion` terraform module for both `libvirt`and `openstack`providers
* a new `role:ssh_minion` Salt grain
* a new `ssh_minion/` subdirectory of `salt/` directory, with a `testsuite.sls` file in it


Bischoff:
Closing in favour of less ambitious PR.

PR #341, opened 2018-05-23
Offer test suite packages on SSH minions

Bischoff:
"SSH minion" clients are managed with the generic "host" module and no role.

They however differ from a generic host in the sense that they need to be able to benefit from test suite's repository and packages. This is what this PR adds.


salt/client/testsuite.sls
```diff
@@ -18,12 +18,6 @@ client_cucumber_requisites:
 
 {% if grains['os'] == 'SUSE' %}
 
-testsuite_build_repo:
-  file.managed:
-    - name: /etc/zypp/repos.d/Devel_Galaxy_BuildRepo.repo
```
moio: Just to double check: is this really never needed for traditionally-managed clients?

salt/default/testsuite.sls
```diff
@@ -0,0 +1,45 @@
+{% if grains.get('testsuite') | default(false, true) %}
+{% if grains.get('role') == '' %}
```
moio: Shouldn't this be `is None`?

Bischoff: i looked at the doc of get() and they said that if default value is absent, then return value is empty string. However i might be wrong and it's not tested yet. Feel free to object stronger ;-) .

salt/client/testsuite.sls
Bischoff: traditional clients are also "default" hosts.

.... argh! I test for no role then! *bug*, thanks

salt/default/testsuite.sls
mateiw: Hmm, maybe `grains.get('role', '') == ''` would be better. Afaik `get(key)` without a default value throws an err if the key is missing.

salt/client/testsuite.sls
Bischoff: ok, fixing by doing a permissive clause as you suggested in irc

salt/default/testsuite.sls
Bischoff: I've put a | default() guard cause anyway... Let's see if that works...

salt/default/testsuite.sls
```diff
@@ -0,0 +1,40 @@
+{% if grains.get('testsuite') | default(false, true) %}
+{% if (grains.get('role') | default('ssh_minion', true)) in ['client', 'minion', 'ssh_minion'] %}
```
moio: `grains.get('role') in ['client', 'minion', None]` should work

Bischoff: Yes, but i find this more expressive so, as it explicitely mention a pseudo-role "ssh-minion".

If you insist I'll change.

Bischoff:
Full scale testsuite test is running currently.

In last version, I had to move "testsuite" grain to "host" top-level module in openstack and libvirt providers. Having a real "ssh_minion" role would have avoided that :smile_cat: .

salt/default/testsuite.sls
Bischoff: That's precisely the problem with not having a dedicated role... I'm not sure this should apply to any vanilla minion... anyway. I'll change it.

Bischoff:
Tests were green. OK to merge?

Bischoff:
Done, thanks.

PR #348, opened 2018-05-28
Change the default memory in controller. Fixes #347

### LINKED ISSUE #347: controller default memory is not enough
In our jobs we have the controller ```memory = 2048 ``` for all the jobs because Capybara apparently needs more than 1G. We should change the default if all jobs are using it.

I will open a PR soon for this if you agree, I already talked with @MalloZup and @Bischoff about this last week. Please assign the issue to me if you agree.

ktsamis:
If you agree this changes one line to change the memory from 1024 to 2048 so we can remove it from the jobs specific conf and leave the default when this gets merged. Fixes #347

### DIFF
--- a/modules/libvirt/controller/variables.tf
+++ b/modules/libvirt/controller/variables.tf
@@ -82,7 +82,7 @@ variable "ssh_key_path" {
 
 variable "memory" {
   description = "RAM memory in MiB"
-  default = 1024
+  default = 2048
 }
 
 variable "running" {


mcalmer:
I think more important is, to change this in sumaform-test-runner repo.

moio:
@mcalmer apparently we did not override the default in sumaform-test-runner, so this should suffice.

@ktsamis, please report if this addresses your issue.

mcalmer:
@moio @ktsamis - we do, but the setting is already 2048
https://gitlab.suse.de/galaxy/sumaform-test-runner/blob/master/QAM/31/main_test.tf#L127

ktsamis:
@moio We already had the setting for 2048 in sumaform-test-runner. It does address the issue, and since we need for all the configured jobs 2048 I thought we would move the setting from the test runner to sumaform default. So now that it's merged, all good from my side

PR #352, opened 2018-05-30
add cloud script to sles7.4 and enable img for both backend

MalloZup:
- [x] update enablement 
git clone https://src.fedoraproject.org/rpms/cloud-init.git
- [x]  tests new libvirt img

- [x]  tests cloud img

### DIFF
--- a/packer/http/ks_sles-es7.cfg
+++ b/packer/http/ks_sles-es7.cfg
@@ -24,6 +24,9 @@ selinux --disabled
 timezone --utc Europe/Berlin
 bootloader --location=mbr
 
+# needed for cloud-init
+url --url=http://mirror.centos.org/centos/7/os/x86_64/
+#repo --name=basecentos7 --baseurl=http://mirror.centos.org/centos/7/os/x86_64/
 # repo to install salt-minion
 repo --name=sumatools-final --install --baseurl=http://nu.novell.com/repo/$RCE/RES7-SUSE-Manager-Tools/x86_64/
 
@@ -222,7 +225,7 @@ openssh-clients
 avahi
 qemu-guest-agent
 salt-minion
-
+cloud-init
 %end
 
 %post --log=/root/post.log --nochroot
--- a/packer/http/ks_sles-es7.cfg
+++ b/packer/http/ks_sles-es7.cfg
@@ -25,11 +25,9 @@ timezone --utc Europe/Berlin
 bootloader --location=mbr
 
 # needed for cloud-init
-url --url=http://mirror.centos.org/centos/7/os/x86_64/
-#repo --name=basecentos7 --baseurl=http://mirror.centos.org/centos/7/os/x86_64/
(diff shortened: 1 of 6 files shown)


packer/sles-es7.json
```diff
@@ -3,6 +3,7 @@
     "type": "shell",
     "scripts": [
       "scripts/base-es.sh",
+      "scripts/cloud.sh",
```
moio: Why was this file modified in-place and not duplicated? Please open a new PR just for this one, I will build it and upload it ASAP. Thanks

MalloZup: i didn't get the question.

from my pov, would be optimal if we have 1 img for libvirt and cloud same time.

But if you want to have 2 files json (libvirt/cloud) i am ok for that.

moio: Let's keep them separate for now like all others. Later I can evaluate a PR to unify them all. I just don't want to have a different rule for this image alone

MalloZup: ok

moio:
Please resolve conflicts and ping me when this needs to be uploaded for testing. Will merge after testing

MalloZup:
@moio i found out that the cloud-init is only on centos7 repo os base.

What do you suggest as approach? 

- enabling temporaly the repo and instal the pkg and disable repo
- move cloud-init pkg in some internal repo, like sumaform ones?

moio:
@MalloZup best would be to build it in a new subproject of systemsmanagement:sumaform, eg.

https://build.opensuse.org/project/show/systemsmanagement:sumaform:sles-es7

just like we do in:

https://build.opensuse.org/project/show/systemsmanagement:sumaform:opensuse423


As a second alternative the EPEL repo can be used.

MalloZup:
@moio i have some issue with the obs repo on rhel7, i just  put the source pkg from fedora server but we have a lot of dependency issues. ( https://build.opensuse.org/project/show/systemsmanagement:sumaform:sles-es7)

i dunno if project-wise we need to add extra-dependencies or pkgs to fix them. 

If you could have a look once you have time would be great.

Otherwise we might add the epel-repo.

packer/http/ks_sles-es7.cfg
```diff
@@ -24,6 +24,8 @@ selinux --disabled
 timezone --utc Europe/Berlin
 bootloader --location=mbr
 
+# needed for cloud-init
+repo --name=basecentos7 --url=http://mirror.centos.org/centos/7/os/x86_64/
```
MalloZup: this should resolve, i am testing the img building.

from kickstart doc
```
--install - Save the provided repository configuration on the installed system in the /etc/yum.repos.d/ directory. Without using this option, a repository configured in a Kickstart file will only be available during the installation process, not on the installed system. 
```

MalloZup:
ok @moio i was able to build it with cloud-init, works as expected and i had a build
yeah

PR #357, opened 2018-06-02
Specify 3.2 nightly

mcalmer:
We are short before branching and we need to run cucumber tests for 3.2 on nightly build.

### DIFF
--- a/salt/default/repos.sls
+++ b/salt/default/repos.sls
@@ -39,7 +39,7 @@ tools_update_repo:
 tools_additional_repo:
   file.touch:
     - name: /tmp/no_tools_additional_repo_needed
-{% endif %}
+{% endif %} {# grains['osrelease'] == '42.3' #}
 
 
 {% if grains['osrelease'] == '11.4' %}
@@ -100,8 +100,8 @@ tools_additional_repo:
 
 tools_additional_repo:
   file.managed:
-    - name: /etc/zypp/repos.d/Devel_Galaxy_Manager_3.1_SLE-Manager-Tools-11-x86_64.repo
-    - source: salt://default/repos.d/Devel_Galaxy_Manager_3.1_SLE-Manager-Tools-11-x86_64.repo
+    - name: /etc/zypp/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-11-x86_64.repo
+    - source: salt://default/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-11-x86_64.repo
     - template: jinja
 
 {% elif 'head' in grains.get('version') | default('', true) %}
@@ -114,7 +114,7 @@ tools_additional_repo:
 
 {% endif %}
 
-{% endif %}
+{% endif %} {# grains['osrelease'] == '11.4' #}
 
 
 {% if '12' in grains['osrelease'] %}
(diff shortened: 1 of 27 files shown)


mcalmer:
@moio I have some trouble to get the server run. As they are in US, download.suse.de seems to point to a local mirror which has invalid data sometimes. Repos do not refresh correctly, because the signature file is still old or repomd.xml contains outdated data.

So I changed download.suse.de to download.nue.suse.com which provide correct data, but it means we access not local resources.

I think we need to discuss how we can run these tests in US.

salt/default/hostname.sls
```diff
@@ -46,4 +46,17 @@ hosts_file_hack:
     {% else %}
       pythonexec: python
     {% endif %}
+{% else %}
+add_domain_to_searchlist:
```
moio: Why is this needed?

salt/default/repos.sls
```diff
@@ -305,8 +287,8 @@ tools_additional_repo:
 
 tools_additional_repo:
   file.managed:
-    - name: /etc/zypp/repos.d/Devel_Galaxy_Manager_3.1_SLE-Manager-Tools-15-x86_64.repo
-    - source: salt://default/repos.d/Devel_Galaxy_Manager_3.1_SLE-Manager-Tools-15-x86_64.repo
+    - name: /etc/zypp/repos.d/Devel_Galaxy_Manager_3.2_SLE-Manager-Tools-15-x86_64.repo
```
moio: Will this also work for `3.1-nightly` clients/mirrors?

salt/default/repos.sls
```diff
@@ -395,27 +377,18 @@ tools_update_repo:
     - require:
       - cmd: galaxy_key
 
-{% elif '3.0-nightly' in grains.get('version') | default('', true) %}
-tools_update_repo:
-  file.managed:
-    - name: /etc/yum.repos.d/Devel_Galaxy_Manager_3.0_RES-Manager-Tools-7-x86_64.repo
```
moio: It was my understanding that we could use `Devel:` client tools repos for any OS and any supported Manager client version - so we should have `Devel_Galaxy_Manager_3.{1,2,3}_{SLE-Manager-Tools-11,SLE-Manager-Tools-12,SLE-Manager-Tools-15,RES-Manager-Tools-7}`.

Does this make sense?

salt/default/hostname.sls
mcalmer: I opened a PR to the SCC dhcp config which provide a searchlist.
I hope the approve this fast and then we do not need this anymore.

salt/default/repos.sls
mcalmer: We only use the "latest" for all SUSE Manager versions. And we are now about to switch to 3.2 versions of the repos.

I have removed the unused repo definitions from salt and minima (https://github.com/moio/sumaform/pull/357/commits/30901814a20f3bf3faf8e7a654b919adbe1f2685)

salt/default/repos.sls
mcalmer: Any OS yes, but not any SUMA version. All Clients use the same version independent of the Server version where they are connected to.

mcalmer:
@moio - I think now we are ready to merge. Maybe you want to check again?

PR #359, opened 2018-06-04
Switch to SLE15, Centos URL fix and switch of default for the EC2 cloud 

brejoc:
* Switching the image URLs from SLE15beta4 to SLE15. Not sure if anything else is needed.
* Image URL for Centos7 seems to have changed. There is no longer a centos7_v3…
* Seems that I've missed to create an PR for this. I think "Making SUSE ECP Cloud the default" would make sense for our target user group.


MalloZup:
@brejoc  i think got fixed. thx for pr anyway

brejoc:
Awesome. Then let me close this.

PR #370, opened 2018-06-13
WIP: add parameter for logging performance queries

MalloZup:


MalloZup:
@moio  feel free to review it

salt/suse_manager_server/postgres.sls
```diff
@@ -12,6 +12,9 @@ postgresql_main_configuration:
       - fsync = off
       - full_page_writes = off
       {% endif %}
+      {% if grains['log_min_duration_statement'] != "nolog" %}
```
MalloZup: i found out that 'false' is 0 in Jinja2 so we cannot  use really false here.

This solution work fine

PR #371, opened 2018-06-14
Bugfix: in case of sle11 systems require is empty.

MalloZup:
Fix this jinja2 bug.

I didn't tested it but should work.

In case of sles11sp4 the require is empty which cause a jinja2 error breaking sumaform
https://ci.suse.de/view/Manager/view/Manager-Head/job/manager-Head-cucumber-sle11sp4/140/consoleFull


Bischoff:
ok, thanks.

PR #372, opened 2018-06-14
exporter packages include services

mcalmer:
The exporter packages now include the systemd service files. 
This requires changes in the configuration.

In addition I have added a jmx_exporter for taskomatic as jmx_exporter is now a template service which can run multiple instances on a system with different configurations.


moio:
Thanks for this PR and sorry for taking this long to merge it. I am currently blocked by #387 which has impacts everywhere, my plan is to go ahead here as soon as that is done

mcalmer:
There are SRs open for the packages agains OBS systemsmanagement:sumaform:tools

moio:
Reviewed, tested, fixed, merged.

Sorry about taking so long! :cry:

PR #377, opened 2018-06-22
Rework the whole repo assignment mechanism

### LINKED ISSUE #369: Hardcoded repo for patterns-suma_server package broke MU and is not needed
In the file ```sumaform/salt/suse_manager_server/init.sls```
we have this:

```
suse_manager_packages:
  pkg.latest:
    {% if 'head' in grains['version'] or 'test' in grains['version'] %}
    - fromrepo: Devel_Galaxy_Manager_Head
    {% elif '3.0-released' in grains['version'] %}
    - fromrepo: SUSE-Manager-3.0-x86_64-Pool
    {% elif '3.0-nightly' in grains['version'] %}
    - fromrepo: Devel_Galaxy_Manager_3.0
    {% elif '3.1-released' in grains['version'] %}
    - fromrepo: SUSE-Manager-3.1-x86_64-Pool
    {% elif '3.1-nightly' in grains['version'] %}
    - fromrepo: Devel_Galaxy_Manager_3.1
    {% elif '3.2-released' in grains['version'] %}
    - fromrepo: SUSE-Manager-3.2-x86_64-Pool
    {% elif '3.2-nightly' in grains['version'] %}
    - fromrepo: Devel_Galaxy_Manager_3.2
     {% endif %}
    - name: patterns-suma_server
    - require:
      - sls: suse_manager_server.repos
      - sls: suse_manager_server.firewall
```

since the module is pkg.latest and the repos are set correctly in ```sumaform/salt/suse_manager_server/repos.sls``` the check for the grain is not needed. It's already checked when the repos are set and if we hardcode the repos where the package is coming from, then if I add ```additional_repos``` that contain this package (as was the MU I was doing) then it's not updated, its dependencies are not updated/installed and the server is not properly set up.

@moio can you assign this to me? I will open a PR when I test this and it doesn't break anything.

moio:
This is a major refactoring/reworking that, among other things, fixes #369.

Before we set repos on a role-by-role basis. Advantage: roles were self-contained in directories. Disadvantages: we had to resort to hacks when repo definitions crossed role boundaries (eg. SUSE Manager Server repos conceptually needed to be set in `default`, we worked around that via repo removals and setting fixed repos for the pattern package, which is #369), some repo refresh was unnecessarily repeated.

With this patch repos are set across roles, before anything else. This allows to pick the Salt version that will end up in the final product from the very beginning (`minimal.sls`).

In the end we traded good design with less hacks, and at least one less bug.


moio:
@MalloZup @Bischoff I plan to merge this on Monday

mcalmer:
@moio please check, what salt will be installed on the Controller node. Currently we have 2018.3.1 which is not supported by us and currently broken. We cannot say when we will fix it. Let us use only supported versions please.

moio:
Checked and merged. @MalloZup, @Bischoff: please ping me in case there is any testsuite disruption

PR #379, opened 2018-06-25
Fix check for grain with right value for role suse_manager_server

ktsamis:
The check for grain role is not matching:
```
https://github.com/moio/sumaform/blob/master/salt/repos/default.sls:

{% if 'suse_manager' not in grains.get('role') %}
```
while in:
```
https://github.com/moio/sumaform/blob/master/salt/top.sls:

  'role:suse_manager_server':
    - match: grain
    - suse_manager_server

```

This should fix that.

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -175,7 +175,7 @@ test_update_repo:
 
 {% endif %}
 
-{% if 'suse_manager' not in grains.get('role') %}
+{% if 'suse_manager_server' not in grains.get('role') %}
 tools_pool_repo:
   file.managed:
     - name: /etc/zypp/repos.d/SLE-Manager-Tools-SLE-12-x86_64-Pool.repo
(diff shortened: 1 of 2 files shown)


ktsamis:
This check also fails because the 
```
{% if 'suse_manager' not in grains.get('role') %}
```
should be:

```
{% if 'suse_manager' != grains.get('role') %}

```
I tried with the above fix and I could create a minssh while before I couldn't. I will add the commit for it

ktsamis:
```is not``` makes more sense than ```!= ``` Testing now and will report later

ktsamis:
Ok I don't know why my comment wasn't saved. The ```!=``` is the correct one. ```grains.get()``` returns a string, or an empty string if no value is there, so comparing strings is fine with ```!=```. Now for my [main.tf](https://gitlab.suse.de/galaxy/sumaform-test-runner/blob/master/QAM/31/main_no_updates.tf) with pretty much all kinds of servers (roles) it renders and creates the minion that was failing before.

ktsamis:
So this is now invalid due to [merge request 380](https://github.com/moio/sumaform/pull/380). The java commit was meant to be separate PR. @moio Please reject this

PR #382, opened 2018-06-27
Add right version of java fixes #378

### LINKED ISSUE #378: java openjdk instead of ibm version installed in server
While using 3.2-released to setup the qam servers, the server failed to set up because all the states depending on this one failed:
```
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):           ID: suse_manager_setup
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):     Function: cmd.run
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):         Name: /usr/lib/susemanager/bin/migration.sh -l /var/log/susemanager_setup.log -s
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):       Result: False
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):      Comment: Command "/usr/lib/susemanager/bin/migration.sh -l /var/log/susemanager_setup.log -s" run
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):      Started: 15:51:26.971114
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):     Duration: 22.441 ms
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):      Changes:
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):               ----------
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):               pid:
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):                   12082
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):               retcode:
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):                   1
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):               stderr:
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):                   update-alternatives: error: alternative /usr/lib64/jvm/jre-1.8.0-ibm/bin/java for java not registered; not setting
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):               stdout:
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):                   Asserting correct java version...
module.srv.module.suse_manager.libvirt_domain.domain (remote-exec):                   Failed to set /usr/lib64/jvm/jre-1.8.0-ibm/bin/java as default java version!
```
Checking the server, indeed the openjdk version was installed. I will open a PR when I test the change. I will make the above state to depend on the package with ibm being installed. Not sure if this happens in other versions as well. @moio  Please assign to me

ktsamis:
New commit for java issue. For some reason putting:

```
suse_manager_setup:
  cmd.run:
    - require:
      - pkg: suse_manager_packages
+    - pkg: java-1_8_0-ibm
      - file: environment_setup_script
      {% if grains.get('apparmor') %}
      - sls: suse_manager_server.apparmor
      {% endif %}

```
or making the suse_manager_setup require the java_ibm file, fails. Trying with just the state without the setup requiring it works for both 3.1 and 3.2 released for me.

### DIFF
--- a/salt/suse_manager_server/init.sls
+++ b/salt/suse_manager_server/init.sls
@@ -28,6 +28,12 @@ environment_setup_script:
     - source: salt://suse_manager_server/setup_env.sh
     - template: jinja
 
+java_ibm:
+  pkg.installed:
+    - name: java-1_8_0-ibm
+    - require:
+      - sls: repos
+
 suse_manager_setup:
   cmd.run:
     - name: /usr/lib/susemanager/bin/migration.sh -l /var/log/susemanager_setup.log -s
--- a/salt/suse_manager_server/init.sls
+++ b/salt/suse_manager_server/init.sls
@@ -29,14 +29,11 @@ environment_setup_script:
     - template: jinja
 
 # HACK: work around bsc#1099454
-
 java_ibm:
   pkg.installed:
     - name: java-1_8_0-ibm
     - require:
       - sls: repos
-
-# end of HACK
 
 suse_manager_setup:
   cmd.run:
--- a/salt/suse_manager_server/init.sls
+++ b/salt/suse_manager_server/init.sls
@@ -28,11 +28,15 @@ environment_setup_script:
     - source: salt://suse_manager_server/setup_env.sh
     - template: jinja
 
+# HACK: work around bsc#1099454
+
 java_ibm:
   pkg.installed:
     - name: java-1_8_0-ibm
     - require:
       - sls: repos
+
+# end of HACK
 
 suse_manager_setup:
   cmd.run:


moio:
This should really not be needed, the pattern should require the correct Java version. I actually wonder why it only happens in your setup - do you have Java-providing custom repos enabled perhaps?

ktsamis:
No I don't have any other repos besides the default sumaform ones. Take a look at the [jenkins job](https://ci.suse.de/view/Manager/view/Manager-3.2/job/manager-3.2-qam-cucumber/4/consoleFull) that is 3.2 and failing to setup the server because state suse_manager_setup fails because the wrong version of java is installed. 

Seeing the script:
```
qam-32-srv:~ # grep java /usr/lib/susemanager/bin/migration.sh                                           │2018-06-26 16:23:31 <1> mgr-srv-31-3(2390) [zypper] main.cc(main):74 ===== Hi, me zypper 1.13.28
    sles)      JVM='/usr/lib64/jvm/jre-1.8.0-ibm/bin/java';;                                             │2018-06-26 16:23:31 <1> mgr-srv-31-3(2390) [zypper] main.cc(main):75 ===== 'zypper' '-n' 'ar' '-ckn' 'SLE
    opensuse)  JVM='/usr/lib64/jvm/jre-1.8.0-openjdk/bin/java';;                                         │S:12-SP3::pool' 'http://download.suse.de/ibs/SUSE/Products/SLE-SERVER/12-SP3/x86_64/product/' 'SLES:12-SP
# ensure correct java version is being used (bsc#1049575)                                                │3::pool' =====
echo "Asserting correct java version..."                                                                 │2018-06-26 16:23:31 <1> mgr-srv-31-3(2390) [zypper] main.cc(~Bye):59 ===== Exiting main() =====
update-alternatives --set java ${JVM}                                                                    │2018-06-26 16:23:31 <1> mgr-srv-31-3(2417) [zypper] main.cc(main):74 ===== Hi, me zypper 1.13.28
    echo "Failed to set ${JVM} as default java version!"                                                 │mgr-srv-31-3:/var/log # vi /usr/lib/susemanager/bin/migration.sh

```

I saw the bug and the only requirement is that the right version is being used, not installed, or able to install. We don't have a dependency for it that I can see:
```
qam-32-srv:~ # zypper info --requires patterns-suma_server
Loading repository data...
Reading installed packages...


Information for package patterns-suma_server:
---------------------------------------------
Repository     : SUSE-Manager-3.2-x86_64-Pool     
Name           : patterns-suma_server             
Version        : 3.2-8.1                          
Arch           : x86_64                           
Vendor         : SUSE LLC <https://www.suse.com/> 
Support Level  : Level 2                          
Installed Size : 57 B                             
Installed      : Yes                              
Status         : up-to-date                       
Source package : patterns-suse-manager-3.2-8.1.src
Summary        : SUSE Manager Server              
Description    :                                  
    SUSE Manager Server packages.
Requires       : [24]                             
    salt-master
    susemanager-docs_en
    susemanager
    SUSE-Manager-Server-release
    release-notes-susemanager
    spacewalk-base-minimal
    susemanager-tools
    osa-dispatcher
    python-ws4py
    spacewalk-postgresql
    spacewalk-reports
    susemanager-advanced-topics_en-pdf
    susemanager-best-practices_en-pdf
    susemanager-getting-started_en-pdf
    susemanager-reference_en-pdf
    py26-compat-salt
    python3-salt
    salt-api
    smdba
    supportutils-plugin-susemanager
    susemanager-branding-oss
    syslinux
    postgresql-contrib >= 9.6
    postgresql-server >= 9.6

```
Not sure if some other package depends on it. Can't do a reverse search on dependencies for java. So for me this is pretty much another bug for the product. That we need to depend on the right version of java, but for now I need a workaround for my jenkins jobs.

moio:
In this case please open a Bugzilla report and include a comment like this:
```
#HACK: work around bsc#....
```

Also put a condition around to check on the affected SUSE Manager version(s).

ktsamis:
I added the comment but not the condition. I think it affects both 3.2 and 3.1. In any case making sure that java-1_8_0-ibm is installed is not a problem for the server.

moio:
What about 3.0? I do not remember the Java version there

ktsamis:
I actually thought that 3.0 is out of maintenance now, Klaus told me we won't have any other MU for 3.0 so I don't think we care about it.

moio:
We do, it's still supported in sumaform as developers need to take a look from time to time for bugfixes and workarounds. Same for support, who will have to deal with 3.0 for longer, due to customers migrating late.

I think I will drop 3.0 support from sumaform sometime next year - until that point we must try not to break it.

mcalmer:
Please check ci.suse.de cucumber tests for HEAD. They are not setting up correctly. AFAIK also because of wrong java installed. But 3.2 and 3.1 works fine. I wonder a bit why we see these problems only sometimes.

I think it would be good to find out. A job for the Round Robin Testsuite Guy?

ktsamis:
My jobs for 3.2, 3.1 and 3.0 all succeeded the setup. I can't run jobs for HEAD atm, that might be a job for the Round Robin Testsuite Guy. For now I think making sure that the right version of java is installed doesn't break anything. At worst it won't succeed for some combination of version and git branch. Very unlikely since the right version of java should be installed on its own already.

mcalmer:
If only HEAD is affected, the reason might be Uyuni changes where we only have openjdk.
The solution could be to now enforce the switch to openjdk for HEAD .

ktsamis:
No, my jobs for 3.2-released are failing because of this. The setup of the script fails. I need this state at least for 3.2-released but I think I had some random failures because of this in 3.1-released as well. It doesn't hurt to include this in the server states, and for Uyuni we can include a jinja conditional for it. I am really not familiar with the settings needed for Uyuni though. Maybe moio can take a look at that? Or the round robin testsuite guy?

moio:
I am merging this now. Uyuni is being reworked anyway, so if problems pop up there we will just fix them.

PR #386, opened 2018-07-05
Adjustments for SUSE manager for retail - Rough network setting

Bischoff:
This pull request adds support for SUSE Manager for retail:
```
    +------------+-----------+------------+  10.160.0.0/16
    |            |           |            |
+---+----+   +---+---+   +---+----+   +---+----+
|  eth0  |   |  eth0 |   |  eth0  |   |  eth0  |
|        |   |       |   |        |   |        |
| server |   | proxy |   | client |   | minion |
|        |   |       |   |        |   |        |
+  eth1  |   |  eth1 |   |  eth1  |   |  eth1  |
+---+----+   +---+---+   +---+----+   +---+----+
    |            |           |            |
    +------------+-----------+------------+  192.168.5.0/24
```
- variable geometry bits:
  - new `retail` base setting
  - corresponding `retail` grain on the hosts
  - corresponding `$RETAIL` variable in the `.bashrc` on the host
- minimalistic network setup if in retail mode:
   - additional private network
   - a second interface to that network on each host
- hacks: needed repos and packages
  - retail "devel" repo (*)
  - branch network formula (**)
  - dhcp server and SuSEfirewall2 packages (**)

The plan is to rebase the finer setup from #383 on top of this PR.

(*) should go away when merged with suse manager
(**) should go away when we have minions as proxies

### DIFF
--- a/modules/libvirt/base/main.tf
+++ b/modules/libvirt/base/main.tf
@@ -65,6 +65,14 @@ resource "libvirt_volume" "sles-es7_volume" {
   pool = "${var.pool}"
 }
 
+resource "libvirt_network" "additional_network" {
+  count = "${var.additional_network ? 1 : 0}"
+  name = "${var.name_prefix}private"
+  mode = "none"
+  addresses = [ "192.168.5.0/24" ]
+  dhcp { enabled = "false" }
+}
+
 output "configuration" {
   depends_on = [
     "libvirt_volume.centos7_module",
@@ -88,6 +96,8 @@ output "configuration" {
     name_prefix = "${var.name_prefix}"
     use_shared_resources = "${var.use_shared_resources}"
     testsuite = "${var.testsuite}"
+    additional_network = "${var.additional_network}"
+    additional_network_id = "${join(",", libvirt_network.additional_network.*.id)}"
 
     // Provider-specific variables
     pool = "${var.pool}"
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -148,6 +148,23 @@ Additionally, it is possible to have only one user to upload images and other sh
  * add a `use_shared_resources = true` variable to the `base` module of all users but one
  * make sure there is exactly one user that does not have the variable set, make sure this user has no `name_prefix` set. This user will deploy shared infrastructure for all users
 
+
+## Additional network and SUSE Manager for Retail
+
+You may get an additional, isolated, network, with neither DHCP nor DNS by specifying:
+
+```hcl
+module "base" {
+  [...]
+  additional_network = true
+  [...]
+}
+```
+
+This will create a network named `private`, with your prefix in front of the name (eg. `hmu-private`). Every VM will get a second network interface `eth1` connected to that network.
+
+You may use that additional network to test SUSE Manager for Retail with the test suite or manually.
+
 ## Custom SSH keys
 
 If you want to use another key for all VMs, specify the path of the public key with `ssh_key_path` into the `base` config. Example:
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -4,14 +4,17 @@ provider "libvirt" {
 
 module "base" {
   source = "./modules/libvirt/base"
+
   cc_username = "UC7"
   cc_password = ...
   testsuite = true
+
   // optional parameters with defaults below
   // pool = "default"
   // network_name = "default" // change to "" if you change bridge below
   // bridge = ""
   // name_prefix = "" // if you use name_prefix, make sure to update the server_configuration for clients/minions below
+  // additional_network = false // needed for testing Retail
 }
 
 module "ctl" {
--- a/main.tf.openstack-testsuite.example
+++ b/main.tf.openstack-testsuite.example
@@ -21,6 +21,7 @@ module "base" {
   // optional parameters with defaults below
   // name_prefix = ""
   // timezone = "Europe/Berlin"
+  // additional_network = false // needed for testing Retail
 
   // comment-out the following two lines if you are not targeting the SUSE internal "ECP" Cloud 
   mirror = "mirror.tf.local"
--- a/modules/libvirt/base/variables.tf
+++ b/modules/libvirt/base/variables.tf
@@ -48,6 +48,11 @@ variable "testsuite" {
   default = false
 }
 
+variable "additional_network" {
+  description = "whether to use an additional network or not"
+  default = false
+}
+
 // Provider-specific variables
 
 variable "pool" {
--- a/modules/libvirt/host/main.tf
+++ b/modules/libvirt/host/main.tf
@@ -31,12 +31,20 @@ resource "libvirt_domain" "domain" {
     var.additional_disk
   )}"]
 
-  network_interface {
-    wait_for_lease = true
-    network_name = "${var.base_configuration["network_name"]}"
-    bridge = "${var.base_configuration["bridge"]}"
-    mac = "${var.mac}"
-  }
+  network_interface = ["${slice(
+    list(
+      map("wait_for_lease", true,
+          "network_name", var.base_configuration["network_name"],
+          "bridge", var.base_configuration["bridge"],
+          "mac", var.mac
+         ),
+      map("wait_for_lease", false,
+          "network_id", var.base_configuration["additional_network_id"]
+         )
+    ),
+    0,
+    var.base_configuration["additional_network"] ? 2 : 1
+  )}"]
 
   connection {
     user = "root"
@@ -54,6 +62,7 @@ resource "libvirt_domain" "domain" {
 hostname: ${var.base_configuration["name_prefix"]}${var.name}${var.count > 1 ? "-${count.index  + 1}" : ""}
 domain: ${var.base_configuration["domain"]}
 use_avahi: ${var.base_configuration["use_avahi"]}
+additional_network: ${var.base_configuration["additional_network"]}
 timezone: ${var.base_configuration["timezone"]}
 testsuite: ${var.base_configuration["testsuite"]}
 use_released_updates: ${var.use_released_updates}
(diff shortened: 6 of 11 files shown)


MalloZup:
cc @aaannz

aaannz:
One thing; retail as a solution can have two network topologies. One specified in PR (machine with dedicated network interface for internal network) and one where there is only one NIC.

Example:
```
WAN <----> [ROUTER] --+-----------+------------+------------+--
                      |           |            |            |
                   [BRANCH]  [TERMINAL1]  [TERMINAL2]  [TERMINAL3]  ...
```

That's why I am not so sure about naming this toggle `retail`. I would prefer something descriptive i.e. `dedicated_network` or `internal_net` or similar.

Bischoff:
@aaannz ok, what you say makes sense.

However, one thing you don't know is that I need a `retail` toggle anyway, It is needed to trigger installation of needed packages and to trigger the tests in the test suite.

Perharps we could use two different settings, one for the network setup (and in that case `branch_network` would be a good name IMHO) and another one for the testing itself (`retail` as previously).

@Moio, you decide? I can split into two different settings if you decide so.

aaannz:
On July 24, 2018 6:05:59 PM GMT+02:00, E Bischoff <notifications@github.com> wrote:
>@aaannz ok, what you say makes sense.
>
>However, one thing you don't know is that I need a `retail` toggle
>anyway, It is needed to trigger installation of needed packages and to
>trigger the tests in the test suite.
>
>Perharps we could use two different settings, one for the network setup
>(and in that case `branch_network` would be a good name IMHO) and
>another one for the testing itself (`retail` as previously).
>
>@Moio, you decide? I can split into two different settings if you
>decide so.
>
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/moio/sumaform/pull/386#issuecomment-407461413

This is exactly why I think using `retail` for network config is not right. I assumed you will want to use that keyword to set up suma as retail solution.

moio:
`retail` in `base` surely does not sound good to my ears. How about a separate "retail-only" module that defines the extra network? (I am just brainstorming at this point)

aaannz:
Still, even in "retail-only" module there are two options for network - one separate and one to leave it as is, but setup libvirt dhcp to use next-server and dns from branch.

The thing is that I thought for years that having dedicated internal network was how the retail solution was used. This kickoff I learned from Jeff that it is not the case and instead this shared network is used in majority.

MalloZup:
summarizing from my pov:

'retail' module is not base. this will install pkg for retail
'networking op' can be in base from my pov

Bischoff:
In the context of this PR, there are 3 things:
 1 - "variable geometry" bits - needed in all the cases to make retail tests optional. It could be controlled by a `retail` flag.
 2 - a special network setup, that I need if I want to test in place the branch network, DHCP, and DNS formulas. It could be controlled by a second flag, for example `branch_network`. 
 3 - hacks, that are temporary only ; some will go away when Retail team's work is merged, and the others will go when we have the "proxy as minion" functionality. For the sake of simplicity, these hacks should also be controlled by `retail` flag.

As Ondrej pointed out, Retail can be tested in a second network setup, where the formulas are not used because DHCP and DNS are on a separated box (the so-called "router"). This is not covered in this PR, but if it is needed, I could add it without too much effort. I'm not sure though this would bring much to the testing side of things...

To sum up the needed flags:
 - `retail` : variable geometry bits, temporary hacks
 - `branch_network` : a special network including the clients and the branch server
 - a third flag that would control whether the branch server is also a router / dhcp / dns box or whether those functionalities are external.

I have to check, but the retail flag is probably not needed in base module, while the other one probably needs to be there.

Does that make sense to you, @aaannz ? @moio ?

Bischoff:
> summarizing from my pov:
>
> 'retail' module is not base. this will install pkg for retail
> 'networking op' can be in base from my pov

ah I missed that  Dario said in a shorter way what I tried to say in previous message :smile: .

(but it's not a "module" in the terraform sense, and installing packages is part of the hacks that are supposed to go away)

MalloZup:
@moio can you have a look on this?

moio:
@Bischoff is the following correct?
 - `retail` would be a new variable on the `suse_manager` and `suse_manager_proxy` modules
 - `branch_network` would be a new variable on the `base` module

moio:
I suggest renaming the `retail` variable in `base` as `additional_network` and its output variable `additional_network_id`.

It would be cool if it also had `additional_network_name`/`additional_network_bridge`/`additional_network_mac` to have the same expressing power of the primary network, but this is only "nice to have" - I am fine with anything that makes testing retail scenarios possible.

What will be needed is OpenStack support as this is not a libvirt exclusive. I can offer to help testing.

Thank you very much

Bischoff:
> I suggest renaming the `retail` variable in `base` as `additional_network` and its output variable 
> `additional_network_id`.

OK, will do.

Do we keep a "retail" variable in parallel? It's used to install additional repos and to do a few hacks. I'm not sure when this all can go. I will ask the Retail guys.


> It would be cool if it also had 
> `additional_network_name`/`additional_network_bridge`/`additional_network_mac` to have the same 
> expressing power of the primary network, but this is only "nice to have" - I am fine with anything that makes 
> testing retail scenarios possible.

OK, but later. Especially that the implementation will depend on the network patch we use for terraform (either mine or the refactoring from the other guy), and I don't see a progress on this.


> What will be needed is OpenStack support as this is not a libvirt exclusive. I can offer to help testing.

This one is more urgent. I will tackle it too. I can do the testing.


> Thank you very much

Sure. Thanks for the review.

moio:
> Do we keep a "retail" variable in parallel? It's used to install additional repos and to do a few hacks. I'm not sure when this all can go. I will ask the Retail guys.

Sure, we need that, but please make it a variable only on modules for which it makes sense (server/minion/proxy)

> OK, but later. Especially that the implementation will depend on the network patch we use for terraform (either mine or the refactoring from the other guy), and I don't see a progress on this.

Assume yours until news.

> This one is more urgent. I will tackle it too. I can do the testing.

Fantastic, I am available in chat if you need any help!

Bischoff:
Current status: I have divided `retail` variable into 3 variables into 3 different modules:
* `private_network` (`base`): use a private network?
* `retail_repo` (`server`): install Retail team repository and packages? **temporary**
* `retail` (`controller`): should we run SUSE Manager for Retail tests?

What is missing:
* support in cloud
* possibility to have the private network defined externally

MalloZup:
@Bischoff  for retail_repo i would not create a variable but use the built-in:
https://github.com/moio/sumaform/blob/master/README_ADVANCED.md#add-custom-repos-and-packages

(this is more flexible then adding new temp variable)

for : 

- retail (controller): should we run SUSE Manager for Retail tests?

My opinion is that we can have a mechanism for disabling Retail tests but imho we should run them together with the HEAD testsuite.

If we have 2 differents testsuite to me is hardware/resource waste and also people are struggling revieing 1 head, imagine to have 2 differents stuff. ( this is the same pattern we had seen with SLE15 tests)

For the rest

moio:
> @Bischoff for retail_repo i would not create a variable but use the built-in:
https://github.com/moio/sumaform/blob/master/README_ADVANCED.md#add-custom-repos-and-packages
> (this is more flexible then adding new temp variable)

+1, good idea!

Bischoff:
> @Bischoff for retail_repo i would not create a variable but use the built-in:

yes, good idea.

> retail (controller): should we run SUSE Manager for Retail tests?
> My opinion is that we can have a mechanism for disabling Retail tests but imho we should run them together
> with the HEAD testsuite.

It was the plan....

Well, not immediately in HEAD, but in 3.2, and once it works in 3.2, in HEAD too. Why so? Because Retail is a 3.2 MU in the first place.

> If we have 2 differents testsuite to me is hardware/resource waste

I never wanted to do that :yum: .

Bischoff:
Doc and templates added.

Bischoff:
@aaannz @MalloZup @moio maybe something was not clear: even if the setup will not be done via formulas but through sumaform (what you call an "external router"), we will still need the private network to simulate the network infrastructure.

modules/libvirt/base/main.tf
```diff
@@ -65,6 +65,14 @@ resource "libvirt_volume" "sles-es7_volume" {
   pool = "${var.pool}"
 }
 
+resource "libvirt_network" "private_network" {
```
moio: Would it make sense to name this resource `additional_network` for consistency?

Bischoff: OK, will do.

Bischoff:
Ok, I'm convinced by @MalloZup, let's remove `retail` variable completly and have only one test suite.

However, I will need *not* to merge the new testsuite features now as they would be red. But waiting is enough.

If or when we add the possibility for an external "router", we will have to be creative for a new variable.

Bischoff:
@mbologna the retail repos are installed on the minions too because of Kiwi packages. What about moving this hack out of sumaform as well and into the main.tf files with @MalloZup's trick?

mbologna:
> @mbologna the retail repos are installed on the minions too because of Kiwi packages. What about moving this hack out of sumaform as well and into the main.tf files with @MalloZup's trick?
> 
> (since I have everything ready, I can do it for you).

I agree with moving this HACK into main.tf but please remember that it's temporary and it will be removed as soon as the FATE entry linked will be closed.
Thanks for taking care of that

Bischoff:
> I agree with moving this HACK into main.tf but please remember that it's temporary
> and it will be removed as soon as the FATE entry linked will be closed.

Of course. My hack too is temporary.

> Thanks for taking care of that

No worries, while I'm at it...

mbologna:
My advice is: if you don't need to invest too much time in it && you like more having this HACK in main.tf, then yes, go ahead.

MalloZup:
i see temporary hack for 10 years :+1:  so we can :rofl:

Bischoff:
Temporary hack removed, merging.

PR #389, opened 2018-07-31
Fix: append replaced values if not found

mbologna:
Replaced values are not inserted into rhn.conf because Salt cannot
find them (as they are not part of the default config file).
This PR uses Salt option append_if_not_found to ensure that these
settings will land on rhn.conf


Bischoff:
LGTM

moio:
Thanks Michele

PR #390, opened 2018-08-01
Feat: temporary workaround for kiwi-desc-saltboot

mbologna:
Workaround: until `kiwi-desc-saltboot` is part of Manager:tools , we need
to manually add this repo that contains `kiwi-desc-saltboot`. Can be removed
when https://github.com/SUSE/spacewalk/issues/5202 is closed


Bischoff:
@moio @mbologna: adding the suse manager for retail repo can only be a temporary workaround. This is supposed to go when Retail team merges their work into SUSE Manager itself. So let's not forget to remove it at that time !

Also, this will conflict with my own PR, because I was doing the same :smile_cat: . No worries, I'll fix that.

PR #387, opened 2018-07-24
make sumaform terraform v11 compatible

MalloZup:
This pr is not taking care of the update to v11 , it is a prerequisite step that we need to implement.

fix #285 

# Left to do:

# gitlab
https://gitlab.suse.de/galaxy/sumaform-test-runner/merge_requests/41

# backup centos7 or sles img

### communication:

- [ ] send an-email to galaxy-devel about the change of API (version != product_version)

PR #393, opened 2018-08-08
[Depends on PR#387] Two more changes for terraform v11

Bischoff:
Transition from `version` to `product_version` in a line that is not covered by Dario's PR (#387).

### DIFF
--- a/salt/minion/testsuite.sls
+++ b/salt/minion/testsuite.sls
@@ -53,7 +53,7 @@ update_ca_truststore:
     - require:
       - pkg: suse_minion_cucumber_requisites
 
-{% if '3.2' in grains['version'] or 'head' in grains['version'] %}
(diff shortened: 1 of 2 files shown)


mbologna:
Since you're targeting dmaiocchi's PR https://github.com/moio/sumaform/pull/387 I think it's better either to rebase this on top of his branch or to clearly state that this PR is depending on that (so it cannot be merged by accident on master)

Bischoff:
In any case, this is not working. I'll simplify it, and this will go on top of his branch as you recommand.

Bischoff:
Simplified...

MalloZup:
i pushed this commit directly on my branch!
https://github.com/moio/sumaform/pull/387/commits/065fcd9f06e53cc6fb337fecd007433cd3b7ede4

thx!

PR #397, opened 2018-08-24
Add 3.2 server update repositories

Bischoff:
This PR finished the work for 3.2 update repositories now we have them after GA.


Bischoff:
You're welcome.

PR #401, opened 2018-09-04
Remove hack for bsc#1099454

juliogonzalez:
Having this hack breaks sumaform for the server for anything != SLE12 (including Leap 42.3)

Not needed anymore, since bsc#1099454 is fixed.

Ping @ktsamis for awareness (can't assign him as reviewer).

### DIFF
--- a/salt/suse_manager_server/init.sls
+++ b/salt/suse_manager_server/init.sls
@@ -28,13 +28,6 @@ environment_setup_script:
     - source: salt://suse_manager_server/setup_env.sh
     - template: jinja
 
-# HACK: work around bsc#1099454
-java_ibm:
-  pkg.installed:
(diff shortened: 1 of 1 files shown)


Bischoff:
Excellent, @juliogonzalez.

This is why I hate those hacks in the testsuite: they tend to be forgotten :crying_cat_face: .

moio:
> This is why I hate those hacks in the testsuite or sumaform: they tend to be forgotten

And this is also why I insist so much in using an explicit `HACK` comment (not `TODO`!) preferably with an indication over when it can be removed.

That is, in the simplifying assumption the hack is temporary.

Bischoff:
ok for `HACK` comment, I'll try to remember.

PR #405, opened 2018-09-06
Remove Ethernet type from ifcfg-ethX as we might use virtio

Bischoff:
Comment out `TYPE=Ethernet` in `/etc/sysconfig/network-scripts/ifcfg-ethX`


Bischoff:
Well, the problem is a consequence of the card being badly declared that is very far away from its root cause.

Giving the original problem here would seem quite unrelated.

moio:
Please also adapt base-es.sh. I expect the behavior to be totally identical there.

Bischoff:
OK.

PR #407, opened 2018-09-07
Use phantomjs and twopence from systemsmanagement-sumaform-tools so they are available for public CI

### LINKED ISSUE #403: Move Devel_Galaxy_cucumber_testsuite_repo to an external repo
Current repository Devel_Galaxy_cucumber_testsuite_repo at https://github.com/moio/sumaform/blob/master/salt/repos/controller.sls will not work as soon as we have a public CI.

I can only see two dependencies there: phantomjs (AFAIK we don't use it anymore) and twopence (I guess we took it from https://build.opensuse.org/package/show/devel:tools/twopence so our stuff doesn't break if there are updates there).

It seems we have two alternatives:

1. Move it to a new subproject at https://build.opensuse.org/project/show/systemsmanagement:Uyuni
2. Move it to https://build.opensuse.org/project/show/systemsmanagement:sumaform

Both options are valid for me. We just need to pick one, copy the package there, and as soon as we have the build, update sumaform.

Ping @moio @MalloZup, I can help, but we need to decide where the twopence should live.

juliogonzalez:
Fixes #403 

Packages are already available at https://build.opensuse.org/project/show/systemsmanagement:sumaform:tools

### DIFF
--- a/salt/mirror/minima.yaml
+++ b/salt/mirror/minima.yaml
@@ -179,10 +179,6 @@ http:
   - url: http://dist.nue.suse.com/ibs/Devel:/Galaxy:/Manager:/TEST/SLE_12_SP3
     archs: [x86_64]
 
-  # Testsuite Controller repo
-  - url: http://dist.nue.suse.com/ibs/Devel:/Galaxy:/cucumber-testsuite/openSUSE_Leap_42.3/
-    archs: [x86_64]
-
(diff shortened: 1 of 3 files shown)


juliogonzalez:
Waiting until 3.2 MU is submitted before merging (just in case)

juliogonzalez:
3.2 submitted. Merging.

PR #409, opened 2018-09-07
[WIP] Experimenting - trying to build image for latest CentOS 7

Bischoff:
**DO NOT MERGE**

Trying to fix this thing.

Root problem is that root partition is saturated. With 2 GB, no wonder.
```
 Filesystem             Size  Used Avail Use% Mounted on
 /dev/mapper/live-rw    2.0G  1.4G  643M  68% /
```
However it looks like, at the time `base.sh` is called, that `/dev/vda` is not even formatted yet.

So we have 200 GB on our qcow2 disk, but at script time, we are not on this disk.

### DIFF
--- a/packer/centos7.json
+++ b/packer/centos7.json
@@ -17,7 +17,7 @@
       "iso_checksum": "714acc0aefb32b7d51b515e25546835e55a90da9fb00417fbee2d03a62801efd",
       "iso_checksum_type": "sha256",
       "iso_url": "http://mirror2.hs-esslingen.de/centos/7/isos/x86_64/CentOS-7-x86_64-Minimal-1804.iso",
-      "ssh_wait_timeout": "30s",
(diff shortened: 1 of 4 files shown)


Bischoff:
It's deeply broken. The new disk does not get formatted even when the installation completes:
```
# qemu-nbd --connect=/dev/nbd0  /home/ebischoff/hard_disks/centos7.qcow2 
# fdisk -l /dev/nbd0 
Disque /dev/nbd0 : 200 GiB, 214748364800 octets, 419430400 secteurs
Unités : secteur de 1 × 512 = 512 octets
Taille de secteur (logique / physique) : 512 octets / 512 octets
taille d'E/S (minimale / optimale) : 512 octets / 512 octets
```

Bischoff:
It was due to `sshd.inst`. Closing.

PR #411, opened 2018-09-11
Make Docker and Kiwi profiles location customizable

Bischoff:
- to enable for tests with private profiles
- to avoid putting credentials in the testsuite itself


modules/libvirt/controller/main.tf
```diff
@@ -11,6 +11,19 @@ variable "testsuite-branch" {
...
 
+variable "testsuite-profiles" {
+  default = {
+    "3.0-released" = "https://${var.git_username}:{var.git_password}@github.com/SUSE/spacewalk.git#Manager-3.0:testsuite/features/profiles"
+    "3.0-nightly" = "https://${var.git_username}:{var.git_password}@github.com/SUSE/spacewalk.git#Manager-3.0:testsuite/features/profiles"
+    "3.1-released" = "https://${var.git_username}:{var.git_password}@github.com/SUSE/spacewalk.git#Manager-3.1:testsuite/features/profiles"
+    "3.1-nightly" = "https://${var.git_username}:{var.git_password}@github.com/SUSE/spacewalk.git#Manager-3.1:testsuite/features/profiles"
+    "3.2-released" = "https://${var.git_username}:{var.git_password}@github.com/SUSE/spacewalk.git#Manager-3.2:testsuite/features/profiles"
+    "3.2-nightly" = "https://${var.git_username}:{var.git_password}@github.com/SUSE/spacewalk.git#Manager-3.2:testsuite/features/profiles"
+    "head" = "https://github.com/uyuni-project/uyuni.git#:testsuite/features/profiles"
```
MalloZup: is this something that will stay forever?
we might call this uyuni head.
I would might avoid this trap of naming variable and then it will be changed

Bischoff:
> i would call the variable uyuni-head instead of head

impossible, unless you redefine the possible values of `product_version`.

juliogonzalez:
WARNING, WARNING, WARNING: It's highly likely that we'll be testing both Devel:Galaxy:Manager:Head (SLE as base) and systemsmanagement:Uyuni:Master (openSUSE Leap as base) at the same time.

So take care of choosing names!!!!!

Bischoff:
I did not choose names @juliogonzales, just using what exists already. And it's not about OBS here, but git.

Is it a correct assumption that both Devel:Galaxy:Manager:Head and systemsmanagement:Uyuni:Master (in OBS) will use uyuni-project/uyuni source code (in git) ?

That being said, it's likely that what you describe, Julio, will bring heavy changes to sumaform. But this is outside the scope of this PR.

Bischoff:
OK, it gives expected results. Now testing in conjonction of testsuite.

modules/libvirt/controller/variables.tf
```diff
@@ -84,6 +84,11 @@ variable "ssh_key_path" {
   # HACK: "" cannot be used as a default because of https://github.com/hashicorp/hil/issues/50
 }
 
+variable "git_profiles" {
```
moio: I do not like this name too much, it sounds confusing when read together with the existing `git_username`, `git_password` and `git_repo`.

`git_profile_repo` sounds better, although at that point we would have `git_repo` and `git_profile_repo` - we should think about renaming the former as well to differentiate them better.

Bischoff: Makes sense.

Bischoff:
This pull request might be dropped, or not. The problem is that github.com does not accept URL-encoded credentials. There were two reasons for writing this was 1) to provide those credentials 2) to enable personalized GIT profiles.

Do we still want this, given that one one of both reasons for it remain? @moio @MalloZup @juliogonzalez, opinions?

If we keep it, code will be substantially simplified.

moio:
What is different in this case from the handling of git credentials we do now for the main testsuite repo (`git_username`, `git_password`)?

Bischoff:
Simplified code (no more credentials, only one branch), and changed variable name as requested.

juliogonzalez:
About having the profile path configurable: I think it will be useful for testing, or if in the future we decided to move the profiles to their own repository (testing Manager-3.2 with docker files from the master branch at uyuni-project/uyuni sounds weird, but it's fine for the time being).

Bischoff:
@juliogonzalez If that happens, then it's indeed enough to change default value in `modules/*/container/main.tf`.

@moio your opinion ? keep or to the bin ?

juliogonzalez:
Or if we by any chance we find how to do it with netrc and curl, or just want to try different profiles :-D

PR #413, opened 2018-09-12
fix: this typo was causing race condition and

MalloZup:
errror when destroying disk because depends_on was based on wrong names
(inexistent)


moio:
Whoa, great catch!

juliogonzalez:
Destroy works fine. now Testing how apply goes.

Thanks @MalloZup!!!

PR #416, opened 2018-09-13
Now osmajorrelease grain is an integer

Bischoff:
This is the root cause to the CentOS problems: now Salt defines `osmajorrelease`as an integer and not as a string anymore.


moio:
Great catch!

Since what Salt version is this true?

Remember we still support SLES 11, so we still have Salt 2016 there and have to support it.

Bischoff:
It changed in nitrogen 2017.7.0:

  https://docs.saltstack.com/en/latest/topics/releases/2017.7.0.html#grains-changes

You are right @moio, my patch is wrong. It will work with newer salt but break with older versions. I need to redo it in a more flexible way.

PR #417, opened 2018-09-14
Added example how to use git_profiles_repo

Bischoff:
doc only commit

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -31,6 +31,7 @@ module "ctl" {
   git_username = ...
   git_password = ...
   git_repo = "default"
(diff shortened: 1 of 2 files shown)


main.tf.libvirt-testsuite.example
```diff
@@ -31,6 +31,7 @@ module "ctl" {
   git_username = ...
   git_password = ...
   git_repo = "default"
+  // git_profiles_repo = ... // if you don't want to use embedded Docker and Kiwi profiles
```
moio: May I suggest "uncomment to use alternative Docker and Kiwi profiles"?

Bischoff: Sure.

PR #419, opened 2018-09-14
Use Uyuni repositories for openSUSE Leap

juliogonzalez:
This also required using patterns from systemsmanagement:Uyuni:Master


juliogonzalez:
Issue created for the doc and assigned to me.

PR #421, opened 2018-09-14
Added CentOS repos for 3.1

Bischoff:
We were putting 3.2 repos on 3.1 CentOS.

I am not sure at all about the difference between
  `Devel_Galaxy_Manager_3.2_RES-Manager-Tools-7-x86_64.repo`
and
  `Devel_Galaxy_Manager_3.2_RES-Manager-Tools-7-Beta-x86_64.repo`.

I suspect an error there. Cen you please review that too?

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -321,11 +321,38 @@ suse_res7_key:
     - watch:
       - file: suse_res7_key
 
-{% if '3.2-released' in grains.get('product_version') | default('', true) %}
+{% if '3.1-released' in grains.get('product_version') | default('', true) %}
 tools_update_repo:
   file.managed:
     - name: /etc/yum.repos.d/SLE-Manager-Tools-RES-7-Beta-x86_64.repo
     - source: salt://repos/repos.d/SLE-Manager-Tools-RES-7-Beta-x86_64.repo
+    - template: jinja
+    - require:
+      - cmd: galaxy_key
+
+{% elif '3.1-nightly' in grains.get('product_version') | default('', true) %}
+tools_update_repo:
+  file.managed:
+    - name: /etc/yum.repos.d/Devel_Galaxy_Manager_3.1_RES-Manager-Tools-7-x86_64.repo
+    - source: salt://repos/repos.d/Devel_Galaxy_Manager_3.1_RES-Manager-Tools-7-x86_64.repo
+    - template: jinja
+    - require:
+      - cmd: galaxy_key
+
+{% elif '3.2-released' in grains.get('product_version') | default('', true) %}
+tools_update_repo:
+  file.managed:
+    - name: /etc/yum.repos.d/SLE-Manager-Tools-RES-7-Beta-x86_64.repo
+    - source: salt://repos/repos.d/SLE-Manager-Tools-RES-7-Beta-x86_64.repo
+    - template: jinja
+    - require:
+      - cmd: galaxy_key
+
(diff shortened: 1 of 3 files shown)


moio:
Client repos are just one version per OS, irrespective of the Manager version. We might have a different version for Uyuni.

Bischoff:
Isn't that covered by 'Head`?

Honestly, I'm a bit lost here. Can you please tell me exactly what I should change?

Bischoff:
I humbly suggest we drop this PR, that you do the PR exactly how you want it, and that I adapt in the testsuite.

Bischoff:
Here is what we have in the various branches of the test suite:
```
     When I enable repository "Devel_Galaxy_Manager_3.1_RES-Manager-Tools-7-x86_64" on this "ceos-traditional-client"
     When I enable repository "Devel_Galaxy_Manager_3.2_RES-Manager-Tools-7-x86_64" on this "ceos-traditional-client"
     When I enable repository "Devel_Galaxy_Manager_Head_RES-Manager-Tools-7-x86_64" on this "ceos-traditional-client"
```

MalloZup:
@moio : 

1.) This should be removed 
https://github.com/moio/sumaform/blob/master/salt/repos/default.sls#L324

especially the `beta` for a released tag. https://github.com/moio/sumaform/blob/master/salt/repos/default.sls#L327.

It should be  https://github.com/moio/sumaform/blob/master/salt/repos/default.sls#L345
( i think this should be adressed on other PRS ( i think all the `released` tag are affected)

2) @Bischoff  what moio mean here is:

the client tools in SUSE-Manager are valid for all SLE12 family clients.

so having https://github.com/moio/sumaform/blob/master/salt/repos/default.sls#L342
we can alllow `3.1-nightly`, `3.2-nightly` etc.

In this centos7 case, we don\t need other repos https://github.com/moio/sumaform/blob/master/salt/repos/default.sls#L306, because this tool are ok

salt/repos/default.sls
```diff
@@ -330,23 +330,42 @@ tools_update_repo:
...
+    - source: salt://repos/repos.d/Devel_Galaxy_Manager_3.1_RES-Manager-Tools-7-x86_64.repo
     - template: jinja
     - require:
       - cmd: galaxy_key
 
-{% elif 'nightly' in grains.get('product_version') | default('', true) %}
+{% elif '3.2-released' in grains.get('product_version') | default('', true) %}
+tools_update_repo:
+  file.managed:
+    - name: /etc/yum.repos.d/SLE-Manager-Tools-RES-7-Beta-x86_64.repo
```
MalloZup: released beta :rabbit: 

I think this need a separate PR @moio for checking all this

Bischoff:
@moio @MalloZup would it be acceptable to merge this PR "as is" to fix the 3.1 production testsuite, and then to refine in a separate PR?

MalloZup:
@Bischoff  to me, i am thinking that the `client tool model doesn`t apply for centos7 but only for SLE12.

I would ping @juliogonzalez or @mcalmer  for this.

For the `released beta` bug open an issue

Bischoff:
I only copied and pasted `-Beta`lines. I am unable to judge of their relevance. Better discuss and fix it in a separate PR. I created issue #422 for that.

Bischoff:
I needed 3.1 to be green again. I think it's clean enough to be reworked. I merged.

PR #424, opened 2018-09-15
Use mirror again

mcalmer:
Artem said changes on the proxy environment was made and if we can try out if the problem was fixed.
I made these changes to point to the mirror address.


mcalmer:
@moio - JFYI: 3.1 cucumber tests currently run with my fork and this change. We can have a look how good the results are before we merge this.

mcalmer:
I think this error is not caused by provo mirror problem. It is a problem with the result of a SCC query or with its handling:

Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "prdext_ber_id_uq"
18:50:31         Detail: Key (base_pdid, ext_pdid, root_pdid)=(173, 547, 1000) already exists.

PR #425, opened 2018-09-17
Added dependency to additional network

Bischoff:
Wait for network to be created, if any.


Bischoff:
currently testing with and without the additional network...

PR #426, opened 2018-09-18
Remove beta repos

### LINKED ISSUE #422: "Released beta" repos
We need to investigate why we use `-Beta` repos for `-released`products.

Maybe there was a good reason for that, but we have lost it from sight.

moio:
This removes repos used during the beta phase of SUSE Manager 3.2.

Fixes  #422

cc @Bischoff


Bischoff:
Thanks Silvio.

PR #429, opened 2018-09-19
Enable Uyuni Stable

juliogonzalez:


README_ADVANCED.md
```diff
@@ -10,11 +10,12 @@ Legal values for released software are:
  * `3.0-released` (latest released Maintenance Update for SUSE Manager 3.0 and Tools)
  * `3.1-released` (latest released Maintenance Update for SUSE Manager 3.1 and Tools)
  * `3.2-released` (latest released alpha/beta/gold master candidate for SUSE Manager 3.2 and Tools)
+ * `stable`       (latest released version for Uyuni Server and Proxy, from systemsmanagement:Uyuni:Stable, for Tools use `head`)
```
moio: Would you consider using `released` instead of `stable` to keep consistency?

We could allow that also for SUSE Manager, pointing to the latest stable release, so it would make always sense. Indeed we already support `released` in clients/minions, so this would be a natural extension.

juliogonzalez: I think we better use `uyuni-released` so it's perfectly clear what we are talking about. Besides, that would allow us (in theory) to install uyuni on top of both SLE and openSUSE (keeping in mind Uyuni is only built for openSUSE of course).

Allowing it for SUSE Manager has some implications, for example at branch selection, so we better tackle that with a separate PR.

moio: Could we also add `uyuni-nightly`?

juliogonzalez: There's no `uyuni-nightly`, only `head` (in fact master).

Keep in mind the releasing cycle for Uyuni is not the same as we have for SUMA as we will have a rolling release approach, so there's only `head` (master, in development) and `uyuni-released` (stable, the most recent released version)

In fact the name "nightly" is problematic it self. It should be `whatever-development`. Nightly implies some kind nightly snapshot, but nightlies are changing all the time, not just at night :-D

juliogonzalez:
This is tested and working. Merging now :-)

README_ADVANCED.md
moio: I know, this comes from Spacewalk where they really had one build at night time. Some day we could change the naming.

PR #435, opened 2018-10-04
add implicit host

MalloZup:


MalloZup:
from upstream 
`host - The address of the resource to connect to. This is usually specified by the provider.`
https://www.terraform.io/docs/provisioners/connection.html#host

PR #432, opened 2018-09-25
Cloudinit- terraform 

MalloZup:
Ubuntu 18.04 with cloudinit

## TOdos: 
- [x] after discussion on irc would be nice to have a DISK resize to 200G for ubuntu image.

## Requirements of this PR:

- we need the `0.5.0` version of terraform-libvirt-provider in sumaform before merging this.

PR #436, opened 2018-10-10
Add support for serial consoles

Bischoff:
(extracted from #432 from @MalloZup )

This PR adds a serial console to our VMs. It enables:
* to use `virsh console` to connect to a VM, especially when network is unavailable
* to avoid Ubuntu to run at 100% CPU and become unusable

### DIFF
--- a/salt/default/init.sls
+++ b/salt/default/init.sls
@@ -29,6 +29,11 @@ timezone_setting:
     - require:
       - file: timezone_symlink
 
+# serial_console:
+#   service.running:
+#     - name: serial-getty@ttyS0
+#     - enable: True
+
 {% if grains.get('use_unreleased_updates') | default(False, true) or grains.get('use_released_updates') | default(False, true) %}
 update_packages:
   pkg.uptodate:


Bischoff:
There are two ways to do this, through a kernel parameter or through a systemd service.

For the moment i'm doing it the easy way, through the service, but on the long term the kernel parameter approach  could give us access to the VM earlier during the boot process.

Yes, I will test CentOS as well.

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/domain.html.markdown

Bischoff:
To anyone interested, the magic kernel parameter is `console=ttyS0,9600`.

Still testing...

Bischoff:
I don't know why, the deployment fails with:
```
* libvirt_domain.domain: timeout while waiting for state to become 'all-addresses-obtained' (last state: 'waiting-addresses', timeout: 5m0s)
* module.client.module.client.libvirt_domain.domain: 1 error(s) occurred:
```
is this linked to some other change of you, or do I miss something, @MalloZup?

MalloZup:
That failure is because sumaform has the variabile wait_for_lease enabled true. This variable basicaly make a for loop over 5 min and wait for ip of the domain(in bridge or not bridge network ) .this is the failure. So in your case the client had pb getting the ipforum. Can you reproduce it always ?  In maybe your client, module of you reproduce always has pb with that kernel arg.it can be the diff img

Bischoff:
@MalloZup I'm not even using the kernel arg, I'm trying with systemd.

I'll try to run without this patch and see if I get an IP. It could be sumaform is trying to use the system console that is now on another tty. @moio, does that hypothesis sound realistic?

I suppose I just have the standard images. I use a fresh sumaform.

moio:
terraform-provider-libvirt gets the VM's IP address(es) via the QEMU guest agent (typically after they have been assigned by the DHCP). I have just checked and yes, the communication with the guest agent uses a serial console, so that sounds like the culprit

https://wiki.qemu.org/Features/GuestAgent

Bischoff:
Thanks @moio, that's probably the explanation.

Did you get same problem in your own PR @MalloZup ? You define more "terminal"s but I tried to do the same, and it does not solve the problem. Are you doing some other magic I missed?

Bischoff:
I might have the solution. The problem might be that qemu-agent is detecting the serial console and tries to use it, but there's no login on it yet.

I will use the kernel parameter approach. With that one, you get a longin prompt very early.

MalloZup:
@Bischoff i think i tested only with the ubuntu and there was no problem i could log in with the qemu-agent
It might be specific issues with SLES images used by sumaform or SLES maybe generic

MalloZup:
but take care that the ubuntu tf has other consoles, this would be your problem imho

Bischoff:
There is some progress: I can now see the kernel booting in my `virsh console`.

However the boot is broken, I probably need some more kernel arguments :smile_cat: .

Bischoff:
It does not work, I was not able to make "Direct kernel boot" feature work.

However, I don't think that's the good strategy. I realized that the good strategy is to add `console=ttyS0,115200` inside grub configuration, on the image.

Which means image rebuilding. Given my past experience with that, I will stop at this point.

MalloZup:
@Bischoff what is the 2nd parameter mean?

MalloZup:
@Bischoff i am still on Ubunt patch waiting on review there, i might help once i have finised.

Is really making images the only solution ? :name_badge:

moio:
@Bischoff JFYI kernel parameters can be specified in Terraform, as `terraform-provider-libvirt` supports them IIUC

https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/domain.html.markdown

search for "params"

Bischoff:
@moio as you might see above, dario already gave the pointer, it's what I tried and it does not work.

Furthermore, it's bad because:
- it uses the "external kernel" feature of libvirt that I kind of dislike
  (I'm not comfortable bypassing grub)
- you can't just add a parameter, you have to specify all of kernel, initrd, and all boot parameters
  (which as a side effect might need you to distinguish SLES / CentOS / etc.)

moio:
> The correct solution is really to put it into the image.

OK. Do we want to keep this PR open then?

This sounds like we want to change all images, and at that point, I wonder if it would make sense to switch to JeOS/upstream + cloud init for most of them.

MalloZup:
The switch to jeos will take more time. E.g JeOs need to be resized root partition is to small.  I would do first ubuntu the jeos

Bischoff:
> OK. Do we want to keep this PR open then?
>
> This sounds like we want to change all images, and at that point, I wonder if it would make sense
> to switch to JeOS/upstream + cloud init for most of them.

It's independant problems, no? I would keep it open.

moio:
The second sentence was a remark, but the question was relative to the quote. Quoting again:

> > The correct solution is really to put it into the image.
> OK. Do we want to keep this PR open then?

moio:
What changes are needed exactly on which images?

Bischoff:
@moio Add `console=ttyS0,115200` to kernel command line parameters on all images.

Bischoff:
Thanks so much @moio !!!

moio:
Right, there was a problem with OBS publishing, now fixed. I tested both SLES and openSUSE and they do the right thing now.

New CentOS images will be provided via the [sumaform-images](https://github.com/moio/sumaform-images) project and they are ready, including console support.

Only problem is a slight incompatibility of GitHub Releases with terraform-provider-libvirt, solved in this PR currently pending:

https://github.com/dmacvicar/terraform-provider-libvirt/pull/472

Plan:
 - I am closing this PR now as no code change is needed anyway
 - as soon as https://github.com/dmacvicar/terraform-provider-libvirt/pull/472 is accepted, I will update [sumaform's build of terraform-provider-libvirt](https://build.opensuse.org/package/show/systemsmanagement:sumaform/terraform-provider-libvirt)
 - as soon as that is ready, I will be committing a change in sumaform to switch over to [sumaform-images](https://github.com/moio/sumaform-images) built images

In future, we will not have hosting problems any more as GitHub releases are publicly available. Moreover, maintenance of Packer images will be easier as [all needed software is containerized](https://github.com/moio/sumaform-images/blob/master/CONTRIBUTING.md#trying-out-a-build-locally), so reproducing the build is not machine dependent any more.

Bischoff:
All clear, thanks Silvio for your help on this long-awaited feature of using virsh console :smiley: .

PR #439, opened 2018-10-16
Add new minion that is acquired through PXE booting (two modules)

Bischoff:
This PR defines two new modules:
* `unprovisioned`: like `host`, but no Salt at all for an authentic vanilla flavor 🍨 
* `pxe_boot`: a kind of unprovisioned host that can boot on its network card

Superseded by #451 (same, but only one module).

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -460,6 +460,43 @@ module "vanilla" {
 }
 ```
 
+## Unprovisioned hosts
+
+Plain hosts still have some provisioning, for example SSH keys. If you want completely
+unprovisioned hosts (no initialization with Salt), use the `unprovisioned` module.
+
+A libvirt example follows:
+
+```hcl
+module "unprovisioned" {
+  source = "./modules/libvirt/unprovisioned"
+  base_configuration = "${module.base.configuration}"
+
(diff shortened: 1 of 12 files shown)


README_ADVANCED.md
```diff
@@ -456,6 +456,43 @@ module "vanilla" {
...
+### PXE boot hosts
+
+PXE boot hosts are special unprovisioned hosts that are capable
+of booting on their networking card. Additionally, they have a hardware
+type of "Genuine Intel" to make provisioning via Retail easier.
+
+A libvirt example follows:
+
+```hcl
+module "pxeboot"
```
moio: Do you already have two different use cases that would justify the existence of two separate modules?

Otherwise I would probably prefer having just one, at the cost of complicating the testsuite's `main.tf` to set the extra three bits (`pxe`, `connect_to_base_network` and `connect_to_additional_network`, which IIUC are the only differences)

Bischoff: A pure vanilla host is something that seemed to make sense to me, yes, independantly of PXE dirty stuff.

But I can merge if you prefer.

Bischoff:
Merged both modules in #451 as requested, closing.

PR #440, opened 2018-10-17
[WIP] Connect each machine only to some networks and not others

Bischoff:
This PR adds two new options to decide to which network(s) a VM should connect to.

This information is fixed for some machines:
* machines that connect only to base network
  * test suite controller
  * SUSE Manager server
* machines that connect to base and additional networks
  * SUSE Manager proxy
  * all clients
* machine that will connect only to additional network
  * pxebooted minion (in another PR)


Bischoff:
Tested, works fine, but doc is missing.

modules/libvirt/host/variables.tf
```diff
@@ -49,6 +49,16 @@ variable "gpg_keys" {
...
 }
 
+variable "connect_to_base_network" {
+  description = "true if you want a card connected to the main network, see README_ADVANCED.md"
+  default = true
+}
+
+variable "connect_to_additional_network" {
+  description = "true if you want a card connected to the additional network (if any), see README_ADVANCED.md"
+  default = true
```
moio: Shouldn't this be false by default, since `additional_network` itself is false by default?

My thinking is that in the majority of use cases will not be about SUSE Manager for Retail, especially in Uyuni, so it would make sense to only enable this when needed.

Bischoff: If you look at the preparation of interfaces, this is ignored if there is no additional network.

This default makes sure that, if additional network is enabled, then you get an interafce on it, which is reasonable, I think.

moio: Sure, the mechanism is understood. I would still prefer to disable by default even when an `additional_network` exists  as I see the existence of an additional network and the connection to it to be separate needs. What I mean is, although the current use case is Retail we want to keep the door open to alternative uses (that's why the parameter was called `additional_network`) so it makes sense to be more explicit.

Apart from this the PR can be merged, :+1:

Bischoff: I agree that the existence of the additional network and the connection to it are two separate things (that's the whole point of this PR).

As a user, I don't like much things that "do not work by default", but OK.

moio: > As a user, I don't like much things that "do not work by default", but OK.

That I would solve with a new template example file or, even better, a testsuite module with a "retail" flag.

Bischoff: OK, I gave an example in the non-testsuite example file (since this is not linked to Retail, why not... :smiley: ).

Bischoff: I just realized one good thing with these two new settings: they also open the door for the scenario where the proxy is connected to the additional network only, and the routing done by a separate box...

PR #446, opened 2018-11-07
Enable only eth0 for avahi

Bischoff:
On machines with several network cards, avahi listens on all. This patch enables avahi only on `eth0`.

We also experience `hostname` command not returning anything on openSUSE 42.3. This might be linked, but I am not sure (and I have been unable to reproduce the problem manually).

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -1,36 +1,50 @@
 include:
   - default.hostname
 
-avahi:
+avahi_pkg:
   pkg.installed:
     {%- if grains['os_family'] == 'Debian' %}
     - name: avahi-daemon
     {% else %}
     - name: avahi
     {% endif %}
+
+avahi_change_domain:
   file.replace:
     - name: /etc/avahi/avahi-daemon.conf
     - pattern: "#domain-name=local"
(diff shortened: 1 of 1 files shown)


moio:
> On machines with several network cards, avahi listens on all. This patch enables avahi only on eth0.

I would actually be happier with the default. Why is this needed?

Bischoff:
@moio first it *might* be needed to fix uyuni branch where the testsuite fails at the very beginning because the
` hostname` command returns nothing on the proxy. However I am not sure this is the solution, I would need to test.

Second, why would we need avahi on the private network? I don't like much letting useless ports open.

It's okay to leave this PR on hold until I have time to test further.

moio:
Seems like this is not really needed for the `hostname` problem, and I would prefer to leave avahi disabled on all interfaces. Should this be closed?

Bischoff:
Er, the problem is that avahi is _enabled_ on all interfaces.

moio:
Sorry, typo, I meant *enabled* (as per previous comment).

Bischoff:
The hostname problem was other. I still think it's not too good to let avahi run on all interfaces, but I can live with it. Closing.

PR #447, opened 2018-11-14
Allow proxy to become salt proxy

aaannz:
Proxy module created only traditional proxy for managed clients.
By including minion sls this proxy is then reregistered as a
salt minion with proxy funcionality

This is required for Retail branchserver


modules/libvirt/suse_manager_proxy/main.tf
```diff
@@ -39,6 +39,8 @@ server_password: ${var.server_configuration["password"]}
 generate_bootstrap_script: ${var.generate_bootstrap_script}
 publish_private_ssl_key: ${var.publish_private_ssl_key}
 apparmor: ${var.apparmor}
+salt_proxy: ${var.salt_proxy}
```
mcalmer: I think the name for this option is not good. There is also a package / service named salt-proxy but it does something completly different.

I think you mean proxy_as_minion or similar.

moio: Looking at [the README_ADVANCED.md section](https://github.com/moio/sumaform/blob/50c9a4d20de34f752c5ac386bdceb0ca347962fb/README_ADVANCED.md#turning-convenience-features-off) I would rather opt for `auto_connect_to_master` to be coherent with the existing `minion` option.

I would also welcome changing the default to `true` and `auto_register` to `false` in order to get a wider exposure of the new codebase, which is the recommended default nowadays anyway.

aaannz: You are right. I renamed the variable as suggested.

aaannz: @moio I don't know what you mean by `auto_connect_to_master` to be coherent with the existing `minion` option.

moio: Quoting the link I posted above:

> ## Turning convenience features off
> 
> By default, sumaform deploys hosts with a range of tweaked settings for convenience reasons. If in your use case this is not wanted, you can turn those off via the following variables.
> 
>  * `client` module:
>    * `auto_register`: automatically registers clients to the SUSE Manager Server. Set to `false` for manual registration
>  * `minion` module:
>    * `auto_connect_to_master`: automatically connects to the Salt Master. Set to `false` to manually configure
>  * `proxy` module:
>    * `auto_register`: automatically registers the proxy to the SUSE Manager Server. Set to `false` for manual registration

It would sound natural to add `auto_connect_to_master` here (although I would agree if you told me `master` was a poor choice as part of that name).

aaannz: Right, now I understand. I added it to proxy module as well. And info that it has any effect only when `proxy_as_minion` is `true`

aaannz:
Renamed `proxy_as_minion` to `minion` and enabled by default.
These changes were enough for my tests to convert proxy to proxy minion. I don't know what you mean by ignored `auto_connect_to_master`. I added it to proxy module as well and enabled by default (it does nothing if `minion` is false however)

moio:
> I don't know what you mean by ignored `auto_connect_to_master`

Well, looking at https://github.com/moio/sumaform/pull/447/files and searching for `auto_connect_to_master` does not find any change under the [salt/suse_manager_proxy](https://github.com/moio/sumaform/tree/master/salt/suse_manager_proxy) directory, which is a little surprising - what is the effect of setting that flag?

Note that currently the only use of `auto_connect_to_master` is in:

https://github.com/moio/sumaform/blob/579e8a695edf01a4f93b5271c13e512375057951/salt/minion/init.sls#L40-L65

and any state under the `salt/minion` directory is run exclusively for [hosts with the `role: minion` grain set](https://github.com/moio/sumaform/blob/579e8a695edf01a4f93b5271c13e512375057951/salt/top.sls#L17-L19), which is not the case for Proxies.

aaannz:
There is a change in file [modules/libvirt/suse_manager_proxy/main.tf](https://github.com/moio/sumaform/pull/447/files#diff-e4884a08d669694413c4a694547e2422R43) where I add `auto_connect_to_master` to the grains of proxy. Then if `minion` variable is true, `minion.sls` is [included in proxy salt states](https://github.com/moio/sumaform/pull/447/files#diff-87ac9b491d5f8842c6d5f8b52d423963R4) and reuses minion salt configuration.
But this got me thinking, maybe instead of introducing new grain `minion` I should reuse `role:minion` grain. I'll need to check that.

moio:
Arrgh, sorry, I totally misread that line!

Yes, in principle that would be better as the assumption is one role <=> one subdirectory in salt/

aaannz:
I tried some ways to reuse `role:minion`, but I can't find a way which I would like.

When I add another role - `role: [suse_manager_proxy, minion]` that works fine with `top.sls` but there is a problem then in `default.sls` where it expectes grain `role` to be string and not an array. The same for if fences in `suse_manager_proxy/init.sls`.
Then I tried adding new role `role: suse_manager_proxy_minion` which would include `suse_manager_proxy` and then `minion`. This however requires another or option in `suse_manager_proxy/init.sls` and others.
At the end, I would prefer to stay with the way I present in this PR.

PR #451, opened 2018-11-20
Add new minion acquired through PXE booting (single module)

Bischoff:
This PR defines a new module:
* `pxe_boot`: an unprovisioned host that can boot on its network card

It supersedes #439 (same, with two modules).


Bischoff:
Testing...

PR #452, opened 2018-11-23
Use new name of "head" branch

Bischoff:
This PR renames "Manager" into "master", since this branch does not exist for some time now. Furthermore, newcomers to Uyuni won't have the history for this name.

There is a related PR for the runner:
https://gitlab.suse.de/galaxy/sumaform-test-runner/merge_requests/67

As this will break the `main.tf`file for everyone, the plan is to send an email to the susemanager mailing list.


Bischoff:
@moio as a tech debt, why not. But not now (lack of time).

In this pecular case, there would be a trick to avoid changing the `main.tf`s:
test "Manager" OR "master"...

Bischoff:
Timeout, merging ;-) .

PR #455, opened 2018-11-30
WIP: Initial commit for custom_channels

ktsamis:
This is not ready to merge, I am having problems with the variable format for HCL. Just a prototype to keep track of things. I need to change the format of the new variable to abide by HCL syntax.

It should add custom_channels variable to suse_manager module that contains a url to be mirrored on the server. It depends on ```channels``` being there for having a base channel.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -173,6 +173,32 @@ At deploy time the `spacewalk-clone-by-date` will be used for each channel set. 
 
 Activation keys are also automatically created for each clone with the name `1-<CLONE_PREFIX>`.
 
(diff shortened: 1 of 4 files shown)


ktsamis:
Basically this is a rework of [this bash script](https://gitlab.suse.de/asemen/suma-custom-channel/blob/master/add_custom_channel.sh) that we use on the long running servers that we use.

Bischoff:
After discussion with @ktsamis, that's nice, but it should not be done in sumaform, because in the testsuite context, you don't have an initialized server with an admin/admin account (it's set up by the core features).

So we will rewrite this in a cucumber feature, @aliouliaki .

PR #457, opened 2018-12-11
[WIP] Allow to set up the additional network's address

Bischoff:
New syntax:
```
additional_network = "192.168.5.0/24"
```

instead of:
```
additional_network = true
```


Bischoff:
Tested, works.

Please don't merge until it's ready on test suite side as well.

PR #464, opened 2019-01-11
Enable only eth0 for avahi

Bischoff:
On machines with several network cards, avahi listens on all. This patch enables avahi only on eth0.

This should prevent this problem encountered by @lkotek:
```
ebischoff@ix64ph1078:~> ping lko-proxy.tf.local    
PING lko-proxy.tf.local (192.168.8.254) 56(84) bytes of data.
```

The address `192.168.8.254` is on private network (`eth1`) and unreachable from controller and server.

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -1,36 +1,50 @@
 include:
   - default.hostname
 
-avahi:
+avahi_pkg:
   pkg.installed:
     {%- if grains['os_family'] == 'Debian' %}
     - name: avahi-daemon
     {% else %}
     - name: avahi
     {% endif %}
+
+avahi_change_domain:
   file.replace:
     - name: /etc/avahi/avahi-daemon.conf
     - pattern: "#domain-name=local"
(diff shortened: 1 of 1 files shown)


Bischoff:
@moio, if you want a simpler (minimal) PR, the only useful lines here are:
```
    - name: /etc/avahi/avahi-daemon.conf
    - pattern: "#allow-interfaces=eth0"
    - repl: "allow-interfaces=eth0"
```

The rest is just reorganization.

moio:
Thanks

Bischoff:
Sure. I'll test it again, just for safety (it was tested, but long ago...) before I merge.

Bischoff:
The `netstat`command shows:
```
udp        0      0 0.0.0.0:5353            0.0.0.0:*                           1705/avahi-daemon:  
udp        0      0 :::5353                 :::*                                1705/avahi-daemon:  
```
so at network level it still listens on all interfaces (`0.0.0.0` and `::`). But my understanding is that it listens on all interfaces but then rejects everything not coming from `eth0`, so it's correct.

PR #467, opened 2019-01-15
Disable private network interfaces

Bischoff:
On some systems like Ubuntu there is no `eth0` but instead `ens3`. Rather than enabling only `eth0`, disable `eth1`and `ens4`.

BTW, there are 5 other occurrences of  `eth0` in sumaform :cry: .

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -20,8 +20,8 @@ avahi_change_domain:
 avahi_restrict_interfaces:
   file.replace:
     - name: /etc/avahi/avahi-daemon.conf
-    - pattern: "#allow-interfaces=eth0"
-    - repl: "allow-interfaces=eth0"
+    - pattern: "#deny-interfaces=eth1"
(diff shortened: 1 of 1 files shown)


moio:
I am not really happy about this patch. Can't we do this with grains, dynamically compiling the black or whitelist?

Hardcoding interface names will always be error-prone.

Bischoff:
I do agree, hardcoding is wrong, and I'm not happy either with this emergency patch.

However, I don't see what the grain would be. Do you mean a new grain, or an existing one, and with what semantics?

I would create a sumaform issue once we have this information.

moio:
grains contain the list of network interfaces, see for example

```
# salt minion.tf.local grains.item ip_interfaces
minion.tf.local:
    ----------
    ip_interfaces:
        ----------
        eth0:
            - 192.168.127.187
            - fe80::3cda:fbff:feb9:a7ad
        lo:
            - 127.0.0.1
            - ::1
```

We could have logic in jinja that extracts the first non-lo interface alphabetically.

PR #476, opened 2019-02-10
Switch head to SUSE Manager 4.0 based on SLE15 SP1

mcalmer:
add the missing pieces to install SUSE Manager 4.0 (head) with SLE 15 SP1.

- add  image definition
- sumaform tools for SLE15 SP1
- Module repository definitions required for SUSE Manager
- change defaults for head
- fix firewall package

### DIFF
--- a/modules/libvirt/base/main.tf
+++ b/modules/libvirt/base/main.tf
@@ -23,6 +23,13 @@ resource "libvirt_volume" "sles15_volume" {
   pool = "${var.pool}"
 }
 
+resource "libvirt_volume" "sles15sp1_volume" {
+  name = "${var.name_prefix}sles15sp1"
+  source = "http://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles15sp1.x86_64.qcow2"
+  count = "${var.use_shared_resources ? 0 : (contains(var.images, "sles15sp1") ? 1 : 0)}"
+  pool = "${var.pool}"
+}
+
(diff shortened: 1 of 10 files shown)


modules/libvirt/base/main.tf
```diff
@@ -23,6 +23,13 @@ resource "libvirt_volume" "sles15_volume" {
   pool = "${var.pool}"
 }
 
+resource "libvirt_volume" "sles15sp1_volume" {
```
moio: Please propagate these changes to the `modules/libvirt` directory also to `libvirt/openstack`. Image is already built at IBS:

https://build.suse.de/project/monitor/Devel:Galaxy:Terraform:Images:OpenStack

At this point I am fine if OpenStack support is there, even if not tested.

salt/suse_manager_proxy/init.sls
```diff
@@ -13,7 +13,11 @@ proxy-packages:
       {% else %}
       - patterns-suma_proxy
       {% endif %}
+{% if grains.get('osmajorrelease', None)|int() == 15 %}
```
moio: Please propagate this change also to the `suse_manager_server` module.

mcalmer: This already exists for server (https://github.com/moio/sumaform/blob/master/salt/suse_manager_server/firewall.sls#L6)

modules/libvirt/base/main.tf
mcalmer: done

PR #481, opened 2019-02-21
Enable additional disk definitions on server too

Bischoff:
This PR makes it possible to use additional disks on the suse manager server too. Currently it's only possible on "hosts" (aka suse manager clients).

Syntax is strictly identical to the one for hosts, i.e. you pass a list of volume ids to the `suse_manager` module:
```
  additional_disk = [ {
    volume_id = "${element(libvirt_volume.packages_disk.*.id, 1)}"
  } ]
```
and you define the extra resources like this:
```
resource "libvirt_volume" "packages_disk" {
  name = "ebi2-packages-disk"
  base_volume_name = "empty_disk.qcow2"
  pool = "default"
  count = 1
}
```

NOTE: I wanted a more ambitious pull request, that would also do the following:
 * define the disk resource for the packages
 * modify `/etc/fstab` to define a mount point for `/var/spacewalk`
 * do the mount
 * possibly reduce the size of the server's main disk
   (200 GiB becomes too much once you have this)

HOWEVER I don't have the time to write it right now, and even with that more ambitious PR, the current PR could remain useful. It's simple and compatible with the rest of sumaform, so I suggest we start with this less ambitious PR.

### DIFF
--- a/modules/libvirt/suse_manager/main.tf
+++ b/modules/libvirt/suse_manager/main.tf
@@ -73,6 +73,7 @@ EOF
   vcpu = "${var.vcpu}"
   running = "${var.running}"
   mac = "${var.mac}"
+  additional_disk = "${var.additional_disk}"
 }
 
 output "configuration" {
--- a/modules/libvirt/suse_manager/variables.tf
+++ b/modules/libvirt/suse_manager/variables.tf
@@ -227,3 +227,8 @@ variable "mac" {
   description = "a MAC address in the form AA:BB:CC:11:22:22"
   default = ""
 }
+
+variable "additional_disk" {
+  description = "disk block definition(s) to be added to this host"
+  default = []
+}


modules/libvirt/suse_manager/main.tf
```diff
@@ -22,6 +22,7 @@ module "suse_manager" {
   use_unreleased_updates = "${var.use_unreleased_updates}"
   additional_repos = "${var.additional_repos}"
   additional_packages = "${var.additional_packages}"
+  additional_disk = "${var.additional_disk}"
```
moio: Please move this under `// Provider-specific variables`

Bischoff: sure.

Bischoff:
> This might be less ambitious, but it is also more flexible in that people might have different uses for it.

Yes.

> I of course welcome if ever you get the time to add to "complete" this (since I suppose you will need to implement partitioning and mounting anyway).

Yes, I'm currently doing this manually every time and that's not a solution.

> Inspiration:
> 
> https://github.com/moio/sumaform/blob/master/salt/mirror/init.sls#L65-L70 (partitioning)
> https://github.com/moio/sumaform/blob/master/salt/mirror/init.sls#L74-L90 (mounting)
> 
> I think you can really just copypaste the code and change the mount point

Thanks for that template!

I was not thinking at partitioning, but it could be an idea. Currently, partitioning is avoided because the packages disk is on top of an "empty-disk.qcow2" disk that is already partitioned.

I was thinking we could just put "empty-disk.qcow2" online, together with the other qcow2s.

PR #482, opened 2019-02-21
monitor the salt event log when running the testsuite

mcalmer:
When we monitor the whole event log we can easier see why errors happens in the testsuite.


mcalmer:
I still want to try out if this really works

mcalmer:
JFYI: the next run of the Head testsuite will use this branch to test if the logging is working.

mcalmer:
I have no idea what happens is salt-master is restarted.

mcalmer:
wow it works :-)

PR #483, opened 2019-02-22
Repository refactoring

moio:
This is to use Salt's `pkgrepo.repo` state instead of plain files for zypper repos.

Expectation is that maintainance and refactorings will be easier.

CC @Bischoff


moio:
@juliogonzalez can you have a look? This is best reviewed commit-by-commit.

Bischoff:
Wow! ...

PR #484, opened 2019-02-25
Install wget on minion

amannch:
In order for a new cucumber scenario to work properly, it would be nice to have `wget` installed at setup time.
new scenario: https://github.com/SUSE/spacewalk/pull/6906

### DIFF
--- a/salt/minion/testsuite.sls
+++ b/salt/minion/testsuite.sls
@@ -18,6 +18,7 @@ minion_cucumber_requisites:
       {% else %}
       - openscap-utils
       {% endif %}
+      - wget
     - require:
       - sls: default
 


moio:
If this is exclusively needed for the testsuite, please move the state to `minion/testsuite.sls`

mbologna:
Why not `curl | bash`?

mbologna:
Why not using `curl`?

amannch:
> If this is exclusively needed for the testsuite, please move the state to `minion/testsuite.sls`

Okay, will do this

> Why not using `curl` ?

screenshot_2019-02-25_15-51-11

also: there is already a given step definition using wget

amannch:
@moio I changed it so that it only gets installed for the testsuite. Can you please check if this can be merged now?

PR #488, opened 2019-03-05
Rename use_un/released_updates to use_sle_un/released_updates

srbarrios:
Just a renaming to clarify the parameter inside a terraform file.


modules/libvirt/client/main.tf
```diff
@@ -4,8 +4,8 @@ module "client" {
   base_configuration = "${var.base_configuration}"
   name = "${var.name}"
   count = "${var.count}"
-  use_released_updates = "${var.use_released_updates}"
```
moio: We have a problem here: this flag used to work for all distros, not only SLE (see `salt/default/init.sls`) so renaming this to a SLE-specific name is definitely wrong.

I think back then I chose to use `use_unreleased_updates` instead of a SLE-specific name to keep coherence with this name.

Bottom line: I am fine if you rename `use_unreleased_updates` to `use_sle_unreleased_updates` (although in future we might have to revert that, if ever we support something like Ubuntu Backports...), but `use_released_updates` must definitely stay as it is.

srbarrios:
That's the reason why I first asked you if it make sense that change, before start to work on it :) It's a pitty to has been waste some time on it.

I think that still not clear the use of that parameter in the context we usually do, as can be missinterpreted with the version of suse manager to use.

But is fine, I will close PR.

modules/libvirt/client/main.tf
srbarrios: So, if all the discussion is about SLE-specific, I will replace both for use_os_unreleased_updates and use_os_released_updates.

I don't like the idea of only change one of them, it will be even less understandable. I don't like the idea to keep it as it is, as it can be misinterpreted with updates on the suma product instead of operating system.

srbarrios: By the way, we should be coherent in the documentation, so in README_ADVANCE.md we are describing those parameters only for SLE, so for me it make sense to have use_sle_release_updates.
If we want to be more generic, then documentation should also describe it. Otherwise, I would just say YAGNI.

moio: I agree to change both variables to `use_os_unreleased_updates` and `use_os_released_updates` respectively, and document coherently.

"Document coherently" means the title should be generic, but in the paragraph we should also mention this is only available for SLE with the particular meaning it as for SLE at the moment. If one day we add support for, eg., Ubuntu proposed updates, we will add an explanation about the exact effect of the option there.

srbarrios: OK, let's do it in that way.
We are probably breaking YAGNI rule, but it will be at least more understandable ;)
I will add changes in few days.

PR #489, opened 2019-03-05
Remove /etc/apt/apt.conf.d/00InstallRecommends on Debian

mateiw:
Make sure `APT::Install-Recommends` is always `true` on Debian-like systems. The default and recommended behavior is to install `Recommends` on Debian.


mateiw:
Thanks for the suggestion. Moved it to the end of `salt/repos/default.sls`. Merging.

PR #490, opened 2019-03-07
Use the correct URL for SLE12 HEAD Client Tools

juliogonzalez:
We have a new URL as the product definition at HEAD was changed to match what we have at SUSE:SLE-12:Update:Products:ManagerToolsBeta

### DIFF
--- a/salt/mirror/minima.yaml
+++ b/salt/mirror/minima.yaml
@@ -150,7 +150,7 @@ http:
     archs: [x86_64]
 
   # SLE 12 (GA and all SPs) Manager Tools Head devel
-  - url: http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/Head:/SLE12-SUSE-Manager-Tools/images/repo/SLE-12-Manager-Tools-POOL-x86_64-Media1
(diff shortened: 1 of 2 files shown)


salt/mirror/minima.yaml
```diff
@@ -130,7 +130,7 @@ http:
     archs: [x86_64]
 
   # SLE 12 (GA and all SPs) Manager Tools 3.2 devel
-  - url: http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/3.2:/SLE12-SUSE-Manager-Tools/images/repo/SLE-12-Manager-Tools-POOL-x86_64-Media1
+  - url: http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/3.2:/SLE12-SUSE-Manager-Tools/images/repo/SLE-12-Manager-Tools-Beta-POOL-x86_64-Media1
```
mcalmer: Did you change it also for 3.2 ?

mcalmer: I think not as I get a 404 Not found

juliogonzalez: Nope. That's a mistake-

Good that you saw it, because I was testing sumaform locally, but without mirror.

Fixing it.

juliogonzalez: Fixed.

PR #492, opened 2019-03-15
Released versions for 3.1 and 3.2 are based only on SP3

ktsamis:
Released versions for 3.1 and 3.2 are based only on SP3 for 3.1-nightly it will never be SP4 it's EOL.

### DIFF
--- a/modules/libvirt/suse_manager/main.tf
+++ b/modules/libvirt/suse_manager/main.tf
@@ -2,9 +2,9 @@ variable "images" {
   default = {
     "3.0-released" = "sles12sp1"
     "3.0-nightly" = "sles12sp1"
-    "3.1-released" = "sles12sp4"
-    "3.1-nightly" = "sles12sp4"
-    "3.2-released" = "sles12sp4"
+    "3.1-released" = "sles12sp3"
+    "3.1-nightly" = "sles12sp3"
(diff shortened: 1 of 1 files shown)


moio:
OK for 3.1, but 3.2 will be supported for another year, so I expect it to be installable both on SP3 and SP4. Am I missing anything?

mcalmer:
Officially supported is sp4 for 3.1 and 3.2. Support for SP3 ends when the overlapping period ends.
Testing against SP4 sounds more correct.

moio:
@ktsamis FMPOV current code is correct, we do not really need this patch. Can you confirm and, if needed, close this PR?

ktsamis:
At the moment the ```3.1-released ``` can only be SP3. Since the next update will be the last one for 3.1 I think SP4 makes no sense to test. If there's an other agreement about it and we support 3.1 with SP4 I don't know about it. We can leave 3.1 with SP4 but since SP4 is not released still, for me makes no sense really.

For ```3.2-released ``` I was looking at it from the QAM point of view which says, SP4 is not released yet so we can switch it when it is, but we need to test it in QA before that point. 

So yes I can specify in my test runs which image to use instead of the default. We don't need this patch, I misunderstood. Closing now, thanks for the feedback :)

PR #493, opened 2019-03-16
Open ipmi for testsuite

mcalmer:
OpenIPMI package is required for our tests. In SLE12 it was a dependency of

  cobbler -> ipmitool -> OpenIPMI

and installed via cobbler.
In SLE15 ipmitool does not have this dependency anymore. This cause our IPMI tests to fail.

Additionally combine multiple pkg.installed into one to safe time


mcalmer:
CC @Bischoff

Bischoff:
OK. Thanks.

PR #496, opened 2019-03-20
[WIP] Feat: remove update-notifier-common for Ubuntu

mbologna:
To be tested.

### DIFF
--- a/salt/minion/testsuite.sls
+++ b/salt/minion/testsuite.sls
@@ -22,6 +22,14 @@ minion_cucumber_requisites:
     - require:
       - sls: default
 
+{% if grains['os_family'] == 'Debian' %}
+  {% if grains['os'] == 'Ubuntu' %}
+  update_notifier_not_present:
+    pkg.removed:
+      - update-notifier-common
+  {% endif %}
+{% endif %}
+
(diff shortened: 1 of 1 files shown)


Bischoff:
Currently testing. Missing dash before pkg.removed: ?

salt/minion/testsuite.sls
```diff
@@ -22,6 +22,14 @@ minion_cucumber_requisites:
     - require:
       - sls: default
 
+{% if grains['os_family'] == 'Debian' %}
```
moio: Please avoid nested/indented blocks if possible (YAML is picky on whitespace). In this case I would suggest a simple `and`.

Thanks!

Bischoff:
Retested, works (in the sense it removes updates-notifier-common package). Merging.

Bischoff:
No effect, I suggest reverting. See however #6267 where I might know what's happening.

moio:
Can you please revert this @mbologna? Feel free to just commit without a PR in this specific case.

mbologna:
> Can you please revert this @mbologna? Feel free to just commit without a PR in this specific case.

Done in https://github.com/moio/sumaform/pull/499

PR #497, opened 2019-03-24
Revert "use newer chromedriver version"

mcalmer:
The newer driver version 72 has memory leaks and is a lot slower
Use the old one as long as we do not need a newer driver because
of other reasons

This reverts commit f23a0763b2780ef5818273d89a13e030e20359cb.

### DIFF
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -60,12 +60,12 @@ cucumber_requisites:
 chromium_fixed_version:
   pkg.installed:
   - name: chromium
-  - version: 72.0.3626.96
+  - version: 64.0.3282.167
 
 chromedriver_fixed_version:
   pkg.installed:
   - name: chromedriver
-  - version: 72.0.3626.96
+  - version: 64.0.3282.167
 
 create_syslink_for_chromedriver:
   file.symlink:


Bischoff:
See other PR, version v73 is okay-ish.

moio:
@Bischoff can you revert this please? A direct commit to master is OK.

mcalmer:
@moio why? Version 72 is not working. We could go to 73 as @Bischoff said but i think this would be a new PR.

moio:
@mcalmer OK, I misunderstood, sorry for the noise.

mcalmer:
Btw. V64 seems to be faster. Current situation: 947 scenarions with 29 errors tested in about 6 hours.

Last week we were at 8 hours test time and the around 100 scenarios were just skipped because of the driver crash.
When testing v73 we should also have a look at the runtime.

Bischoff:
@mcalmer: I just verified in the log file: exact 7 hours.

But it was on x64ph1078.qa.suse.de, which might be slower than cokerunner (?). Hard to say without testing on same machine...

mcalmer:
@Bischoff  - I would suggest to prepare a PR with a switch to 73, push in the afternoon and try it during the night.
Then we can decide tomorrow morning if we rollback again or if we stay with it.

Bischoff:
Will do @mcalmer !

PR #513, opened 2019-04-12
Let hosts be able to use additional swap files

Bischoff:
This PR allows to use a swapfile to complement missing memory at the expense of some slowness.

Most interesting is the test suite controller (as an insurance policy against Selenium becoming crazy), but the others are allowed as well.

Not concerned: the server (SUSE manager will create a swapfile on its own), the mirror.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -102,7 +102,7 @@ By default, sumaform deploys hosts with a range of tweaked settings for convenie
    * `auto_register`: automatically registers clients to the SUSE Manager Server. Set to `false` for manual registration
  * `minion` module:
    * `auto_connect_to_master`: automatically connects to the Salt Master. Set to `false` to manually configure
- * `proxy` module:
+ * `suse_manager_proxy` module:
    * `minion`: whether to configure this Proxy as a Salt minion. Set to `false` to have the Proxy set up as a traditional client
    * `auto_connect_to_master`: automatically connects to the Salt Master. Set to `false` to manually configure. Requires `minion` to be `true`
    * `auto_register`: automatically registers the proxy to its upstream Server or Proxy. Defaults to `false`, requires `minion` to be `false`
@@ -333,7 +333,9 @@ Please note that `iss_master` is set from `master`'s module output variable `hos
 
 Also note that this requires `create_first_user` and `publish_private_ssl_key` settings to be true (they are by default).
 
+
 ## Performance testsuite
+
 It is possible to run the Performance testsuite for SUSE Manager by defining a "pts" module. This will create a test server, a locust load server, an minion instance with evil-minions running on it and (by default) a grafana host to monitor them.
 
 A libvirt example follows:
@@ -734,7 +736,7 @@ With the default configuration, whenever SUSE Manager server hosts are configure
 
 This setting can be overridden with a custom 'from' address by supplying the parameter: `from_email`. A libvirt example would be:
 
-```
+```hcl
 module "suma3pg" {
   source = "./modules/libvirt/suse_manager"
   base_configuration = "${module.base.configuration}"
@@ -749,7 +751,7 @@ module "suma3pg" {
 Internal Server Errors and relative stacktraces are sent via e-mail by default to `galaxy-noise@suse.de`.
 By suppling the parameter `traceback_email` you can override that address to have them in your inbox:
 
-```
+```hcl
 module "sumamail3" {
   source = "./modules/libvirt/suse_manager"
   base_configuration = "${module.base.configuration}"
@@ -760,3 +762,23 @@ module "sumamail3" {
   traceback_email = "michele.bologna@chameleon-mail.com"
 }
 ```
+
+
+# Swap file configuration
+
+You can add a swap file to most hosts, to prevent out-of-memory conditions.
+
+Please note that some systems already come with some swap file or swap partition of their own: Ubuntu and CentOS minions, and
+SUSE Manager server.
+
+A libvirt example is:
+
+```hcl
+module "minion" {
+   ...
+   swap_file_size = 2048 // in MiB
+   ...
+}
+```
+
+To disable the swap file, set its size to 0.
--- a/salt/default/init.sls
+++ b/salt/default/init.sls
@@ -36,6 +36,21 @@ update_packages:
       - sls: repos
 {% endif %}
 
+{% if grains.get('swap_file_size', "0")|int() > 0 %}
+file_swap:
+  cmd.run:
+    - name: |
+        {% if grains['os_family'] == 'RedHat' %}dd if=/dev/zero of=/extra_swapfile bs=1048576 count={{grains['swap_file_size']}}{% else %}fallocate --length {{grains['swap_file_size']}}MiB /extra_swapfile{% endif %}
+        chmod 0600 /extra_swapfile
+        mkswap /extra_swapfile
+    - creates: /extra_swapfile
+  mount.swap:
+    - name: /extra_swapfile
+    - persist: true
+    - require:
+      - cmd: file_swap
+{% endif %}
+
 {% if grains['authorized_keys'] %}
 authorized_keys:
   file.append:
--- a/modules/libvirt/client/main.tf
+++ b/modules/libvirt/client/main.tf
@@ -9,6 +9,7 @@ module "client" {
   additional_repos = "${var.additional_repos}"
   additional_packages = "${var.additional_packages}"
   gpg_keys = "${var.gpg_keys}"
+  swap_file_size = "${var.swap_file_size}"
   ssh_key_path = "${var.ssh_key_path}"
   connect_to_base_network = true
   connect_to_additional_network = true
--- a/modules/libvirt/client/variables.tf
+++ b/modules/libvirt/client/variables.tf
@@ -48,6 +48,11 @@ variable "count"  {
   default = 1
 }
 
+variable "swap_file_size" {
+  description = "Swap file size in MiB, or 0 for none"
+  default = 0
+}
+
 variable "ssh_key_path" {
   description = "path of additional pub ssh key you want to use to access VMs, see README_ADVANCED.md"
   default = "/dev/null"
--- a/modules/libvirt/controller/main.tf
+++ b/modules/libvirt/controller/main.tf
@@ -19,6 +19,7 @@ module "controller" {
   name = "${var.name}"
   additional_repos = "${var.additional_repos}"
   additional_packages = "${var.additional_packages}"
+  swap_file_size = "${var.swap_file_size}"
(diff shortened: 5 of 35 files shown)


Bischoff:
Currently testing...

Bischoff:
Tested, works.

README_ADVANCED.md
```diff
@@ -760,3 +762,24 @@ module "sumamail3" {
   traceback_email = "michele.bologna@chameleon-mail.com"
 }
 ```
+
+
+# Swap file configuration
+
+You can add a swap file to most hosts, in case you want the tests not to fail in the case memory gets exhausted.
```
moio:
```suggestion
You can add a swap file to most hosts, to prevent some out-of-memory conditions.
```

salt/default/init.sls
```diff
@@ -36,6 +36,22 @@ update_packages:
       - sls: repos
 {% endif %}
 
+{% if grains.get('swap_file_size', "0")|int() > 0 %}
+file_swap:
+  cmd.run:
+    - name: |
+        fallocate --length {{grains['swap_file_size']}}MiB /swapfile
```
moio: Was this tested on non-SUSE OSs? Does it work as intended there?

README_ADVANCED.md
Bischoff: Yup. Will do.

salt/default/init.sls
Bischoff: 1 - No, it was not tested on non-SUSE.

2 - It might also be it does not work on non-SUSE for another reason: CentOS and Ubuntu already have a swap file or partition. And even if it works, it is pretty useless...

3 - Maybe it's more generic: what if a SUSE system image already has a swapfile or partition?

I definitely need to investigate... Let's keep this open, I'll have a look when I have 5 minutes (ie almost never :smile_cat: ).

Bischoff: On CentOS, it fails, not because of CentOS itself, but because of XFS. It can be solved with:
```
dd if=/dev/zero of=/swapfile bs=1048576 count=(number)
```
but this is way slower than fallocate.

No problem with existing swap partition.

Bischoff: On ubuntu, it overrides the existing `/swapfile`.

Maybe that could be considered the expected behaviour.

Bischoff:
OK, latest version has been tested on both SLE, CentOS, and Ubuntu, and works.

I had to remove `swapon -a` command which was both not needed on any system, and failing on CentOS.

Can you please do last review @moio so I can merge?

moio:
> OK, latest version has been tested on both SLE, CentOS, and Ubuntu, and works.

Does that mean all OSs get the swapfile as configured regardless their original setting?

Bischoff:
> Does that mean all OSs get the swapfile as configured regardless their original setting?

Yes.

README_ADVANCED.md
```diff
@@ -760,3 +762,24 @@ module "sumamail3" {
   traceback_email = "michele.bologna@chameleon-mail.com"
 }
 ```
+
+
+# Swap file configuration
+
+You can add a swap file to most hosts, to prevent out-of-memory conditions.
+
+By default, swarm minions get 8 GB of swap file, and test suite controller gets 4 GB. Default for other hosts is no swap file.
```
moio:
```suggestion
```

I'd just omit this detail, as we omit documenting default values for vCPU, RAM, etc. We just try to pick sane defaults, and here in README_ADVANCED we document how to change that when needed.

README_ADVANCED.md
```diff
@@ -760,3 +762,24 @@ module "sumamail3" {
...
+You cannot set a swap file for the SUSE Manager server, because SUSE manager will create a swapfile on its own, nor for the mirror.
```
moio: What is the reason behind those two exclusions?

README_ADVANCED.md
```diff
@@ -760,3 +762,24 @@ module "sumamail3" {
...
+Please note that some operating systems already come with some swap file or swap partition of their own: Ubuntu and CentOS.
```
moio:
```suggestion
```

See above.

README_ADVANCED.md
```diff
@@ -760,3 +762,24 @@ module "sumamail3" {
...
+
+A libvirt example is:
+
+```hcl
+module "minion" {
+   (...)
```
moio:
```suggestion
   ...
```

README_ADVANCED.md
```diff
@@ -760,3 +762,24 @@ module "sumamail3" {
...
+   swap_file_size = 2048 // in MiB
+   (...)
```
moio:
```suggestion
   ...
```

README_ADVANCED.md
Bischoff: Well, my idea was it would not make sense on server since there was already one. But that was idiotic, we can enjoy two swap files, after all.

Same for the mirror, I did not see the point.

I will make them part of the mechanism and remove this remark.

README_ADVANCED.md
Bischoff: Makes sense.

README_ADVANCED.md
Bischoff: If you don't mind, I would keep this remark, and even mention the suse manager server.

README_ADVANCED.md
Bischoff: ok

Bischoff:
Changes done. However, a small problem remains @moio : Ubuntu images are using `/swapfile` as a file name just like the name Duncan picked up.

If we rename that file, then we will be able to use an additional swap file just like on all other systems.

Do you agree with the idea, and if yes, do you have any suggestions ? We can't use `/SWAPFILE` as it's the name already used by SUSE Manager server. What about `/Swapfile` or `/additional_swapfile` or `/sumaform_swapfile`?

Bischoff:
I just picked up `/extra_swapfile`. Merging.

moio:
I am not thrilled by the idea of more than one swapfile. Is there a reason not to override the existing one, if any? That's what I would expect from a Salt POV.

Bischoff:
@moio Er, yes, there is a reason : increase the existing swap space by the amount you specify.

That seems consistent to me with the semantics of `swap_file_size = nnn` (please add some more swap space for nnn megabytes).

Also, it would really be complicated to compute:

   `requested swap file size - sum of existing swap partitions sizes - sum of existing swap files sizes`

(and what if the result is negative? :smile_cat: )

moio:
What I would expect is that any existing swap file is removed, and only one swap file with the specified size is configured. IOW, the configuration one sets overrides any possible existing setting. Is this doable?

Bischoff:
> What I would expect is that any existing swap file is removed, and only one swap file with the specified size
> is configured. IOW, the configuration one sets overrides any possible existing setting.

Both strategies are possible (add a new swap file, or replace existing ones). I don't see one as more natural or having advantages on the other.

> Is this doable?

A swap partition can be disabled, but not removed (it boils down to reformatting the disk), so the theoretical answer is "no".

It should be possible to remove existing swap files, albeit it's probably complicated. As long as I don't see an obvious advantage of doing this, I don't think I will be spending more time on it.

I think the important thing is that the behaviour is consistent across all platforms. Currently, it is.

moio:
I request a fix from a design standpoint: both Salt states and Terraform files should define what is wanted (aka the "ought-ness") and get there whatever the starting condition is (aka the "is-ness"). Of course "any starting condition" is usually totally impractical, so we restrict that to the documented design assumptions of images (https://github.com/moio/sumaform/blob/master/DESIGN.md#requirements-for-base-disk-images).

Thus I see two possibilities:
1. swap files are disabled on all base images that currently have one and we document the design choice in DESIGN.md. The current code will then just add swap files assuming there is none, and it is a safe assumption because it's a documented design assumption
2. we adapt code added by this PR to delete any previously existing swap file and create the new one with the specified quantity


1. would be my favourite option and it's probably the easiest to implement.

Bischoff:
Well, 

1. you can define "what is wanted" as "i want additional 2 GB" :smile_cat: 

2. more seriously, you will never be able to get "what is wanted" because you have no control on what is previously on swap partitions

Remember there are not only swap files, you might also have pre-existing swap partitions - so deleting swap file has little point to get a desired amount of swap space.

moio:
We control base images and how they are partitioned.

Bischoff:
Then why can't you control initial swap files as well ?

Bischoff:
Don't misunderstand me - I get your point. It's just that I don't see a big emergency here.

Feel free to create a card where i would delete previous swapfiles - but i don't guarantee any delay.

moio:
> It's just that I don't see a big emergency here

True, there isn't any - this is about consistency of the software design and that aspect almost never qualifies as "urgent". Despite not being urgent, consistency is often important, especially long-term.

As the curator of this project, not breaking the design is pretty high on my wish list. My opinion is actually that sumaform has been more successful than other/previous projects because of many factors, the top two of which are:

1. a good level of participation from many people
1. a certain level of consistency in the design, kept in track mainly via asking for modifications in PR reviews

In the light of 1. I welcome people to proceed without me whenever I am not around for whatever reason and I trust them, _especially you_, to take the lead without explicit approval whenever appropriate.

But to be completely honest, I wouldn't have approved this PR in the form it was merged because of 2., hence I am requesting adjustments ex-post. It's just my compromise between the two conflicting principles above - world would be worse (FMPOV) both if there were blocking reviews for everything that gets in, and if the project grew with less order because "once it's merged, it's merged".

Of course I could fix that myself or possibly delay the resolution by just opening an issue, but I hope you will re-observe this matter from my same perspective, share my principles and act as a curator instead.

Bischoff:
I give up, you win :smiley_cat:. Please create a card for QA team, labels qa-squad and sumaform, assign it to me, and I'll try to sneak it in as soon as possible.

You're right, it "should not be too long" (every time I say this, four kittens die :skull_and_crossbones: :crying_cat_face: ...).

To be very honest too, your high requirement standards are all in your honour, but that's something that sometimes makes me hesitate contributing. For example in this PR, I hesitated a lot to start doing it, because I knew there would be follow-ups and discussions about how to make it better...

moio:
> To be very honest too, your high requirement standards are all in your honour, but that's something that sometimes makes me hesitate contributing.

That is an angle I only had intuitively in mind but never sat down and really thought through it.

You are right.

Of course I dislike being thought as a deterrent to contributions but I can see this happening in retrospect - I will have to think how to better balance being an enabler going forward. Thank you very much for bringing this to my attention.

Let's please have a chat whenever you have an idea in the future but hesistate - that will help me giving a concrete case to reason about.

:heart:

Bischoff:
Sure, I will.

I hope I was not too direct / to harsh... :blush: .

moio:
Absolutely not, no worries.

PR #515, opened 2019-04-16
WORKAROUND while our IPv6 infrastructure is broken

Bischoff:
Emergency fix as long as we can't reach outside SUSE with IPv6 from mgr.suse.de.

This touches only the test suite controller. Hopefully it's enough. If it's not enough, we will need a proper setting `ipv6 = no` at host level.

To be removed once the infra problems are solved.


Bischoff:
Ping @moio for awareness.

PR #523, opened 2019-04-25
Fix monitoring package and service file names

cavalheiro:
Requires that https://build.opensuse.org/request/show/697856 is accepted first.


moio:
Thanks!

PR #524, opened 2019-04-25
Use fqdn instead of 'localhost' as a target for reposync

aaannz:
- on SUMA4/Uyuni the SSL validation fails if the 'localhost' is
  used as a destination for reposync check script.
  Using machine fqdn grain to solve this problem.


aaannz:
Note I did not test this with avahi yet!

aaannz:
Retested with avahi and it works.

PR #527, opened 2019-05-03
Avoid machine name ending in -2 in case of an avahi reflector

Bischoff:
In case there is an avahi reflector on the network, the machine might end up with a name like `hmu-minion-2.tf.local`.

Issue is described here: https://github.com/lathiat/avahi/issues/117

Basically, the machine considers its own avahi probe as coming from a duplicate machine, and adds the `-2` to its own name.

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -25,6 +25,22 @@ avahi_restrict_interfaces:
     - require:
       - pkg: avahi_pkg
 
+# work around https://github.com/lathiat/avahi/issues/117 triggered by reflector
+avahi-avoid-name-conflicts:
+  file.replace:
+    - name: /etc/avahi/avahi-daemon.conf
+    - pattern: "use-ipv6=yes"
+    - repl: "use-ipv6=no"
+    - require:
+      - pkg: avahi_pkg
+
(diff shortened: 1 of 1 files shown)


Bischoff:
(tests in progress)

Bischoff:
@moio : forget it: it's even worse than that, the name changes over time! We start with `hmu-minion.tf.local`, and continue with `hmu-minion-2.tf.local`.

I will close this PR at the end, and try to figure out an alternate approach. In the meantime, I will leave it open, so you can read it and give your feedback.

Bischoff:
I have a working fix for avahi itself. Closing this as it becomes moot.

moio:
> I have a working fix for avahi itself.

Really?! Now you got me really curious, please share details!!!

Bischoff:
Here you are @moio:

https://github.com/lathiat/avahi/pull/224

It boiled down to removing a line which itself was a work around.

I'm not sure this will get accepted, and if it is, it might take ages until it is merged, and then until the package is available. Could we use a patched avahi in sumaform (on the minion only) ?

Bischoff:
It happens only in our avahi packages... I need a more accurate analysis.

PR #528, opened 2019-05-06
Add variable for http proxy

lkotek:
PR adds variable `http_proxy` to be used by testsuite. The variable specifies hostname and port for HTTP proxy on SUSE Manager. https://github.com/SUSE/spacewalk/issues/7726


lkotek:
Successfully tested. Sumaform environment works with the changes correctly.

lkotek:
> Hey @lkotek, thanks for this contribution!
> 
> Please also apply identical changes to the `openstack` module, even without explicitly testing them. We simply want to keep those aligned wherever possible.
> 
> I would also discourage having a default that points to SUSE internal architecture - sumaform is meant to be used by the community at large. Please change the default to no proxy and change our internal `main.tf` files in order to point to `galaxy-proxy.mgr.suse.de`.

@moio Thanks for review! Good point. I did changes you requested and prepared MR changing our internal main.tf files too.

lkotek:
@moio Perfect, thanks! Can you merge it, please? (I don't have write access.)

PR #529, opened 2019-05-07
[WIP] Fix: use bigger flavor for SLES15SP1

mbologna:
Without this patch you will not be able to spin up HEAD on SLES15SP1 as
it requires a bigger disk


mbologna:
Apparently, we just need a custom flavor just with only a bigger disk.

PR #530, opened 2019-05-07
[WIP] All clients as optional

srbarrios:
Not sure about the reason behind to have minion and traditional clients as mandatory, when you define a controller in our main.tf
As an idea, that PR makes all clients optional.

Here the card for test suite support: https://github.com/SUSE/spacewalk/issues/7786

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -58,6 +58,7 @@ module "srv" {
   ssh_key_path = "./salt/controller/id_rsa.pub"
 }
 
+# optional
 module "cli-sles12sp3" {
   source = "./modules/libvirt/client"
   base_configuration = "${module.base.configuration}"
@@ -69,6 +70,7 @@ module "cli-sles12sp3" {
   ssh_key_path = "./salt/controller/id_rsa.pub"
 }
 
+# optional
 module "min-sles12sp3" {
   source = "./modules/libvirt/minion"
   base_configuration = "${module.base.configuration}"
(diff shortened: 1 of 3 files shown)


moio:
I am fine with this change, but I guess more is needed before this works as intended.

Maybe @Bischoff can shed light.

srbarrios:
I don't recall in anything else we could need on sumaform side.

Bischoff:
> I don't recall in anything else we could need on sumaform side.

Generating environment variables in `.bashrc` on the controller :smile_cat: .

moio:
I marked the PR as [WIP]. Please keep it that way until code in the testsuite is also ready for the change - I would like both sides to be patched in sync.

srbarrios:
As discussed in IRC, that PR is ok, I don't think we need to split in several PRs. And we will have it pending, until our SUSE Manager TestSuite is updated to support it.
It doesn't mean that merging that PR something will break, as it will only fail on testsuite execution if we remove a minion or a traditional client in the main.tf of that testsuite.
As said, I plan to create a PR in spacewalk for it, I keep you in the loop. But I think this repository and this PR is not the best place to discuss about a different project.

Bischoff:
If both sides are in sync, why not.

moio:
This is now pretty different from `master` and would more or less need to be rewritten. I am closing this PR for the time being (while still accepting the spirit of it) - just open another one when the time comes.

Thanks for your efforts anyway!

Bischoff:
OK.

PR #532, opened 2019-05-09
Let a minion contain an avahi reflector if desired

Bischoff:
An Avahi reflector can forward mDNS packets between two network cards belonging to different networks.

This PR introduces a new setting, `avahi_reflector`,  that lets a minion act as a reflector. It can then for example forward packets between the minion's `eth0` card and the network interface of a Docker container on that minion. This is needed to run the container tests of the test suite when the VMs have no DNS entries.

This PR sometimes triggers a known Avahi bug, https://github.com/lathiat/avahi/issues/117.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -55,7 +55,8 @@ module 'srv' {
 }
 ```
 
-## Changing OSs
+
+## Changing Operating Systems
 
 You can specifiy a base OS in most modules specifying an `image` variable.
 
@@ -76,6 +77,7 @@ module "suma3pg" {
 }
 ```
 
+
 ## Multiple VMs of the same type
 
 Some modules, for example clients and minions, support a `count` variable that allows you to create several instances at once. For example:
@@ -93,6 +95,7 @@ module "minionsles12sp1" {
 ```
 
 This will create 10 minions connected to the `suma3pg` server.
+
 
 ## Turning convenience features off
 
@@ -125,6 +128,7 @@ By default, sumaform deploys hosts with a range of tweaked settings for convenie
    * `create_sample_bootstrap_script`: whether to create a sample bootstrap script for traditional clients. Requires `create_sample_activation_key`
    * `publish_private_ssl_key`: copies the private SSL key in /pub for Proxies to copy automatically. Set to `false` for manual distribution
 
+
 ## Adding channels to SUSE Manager Servers
 
 You can specifiy a set of SUSE official channels to be added at deploy time of a SUSE Manager Server. This operation is typically time-intensive, thus it is disabled by default. In order to add a channel, first get the label name from an existing SUSE Manager Server:
@@ -148,6 +152,7 @@ module "suma3pg" {
   channels = ["sles12-sp2-pool-x86_64"]
 }
 ```
+
 
 ## Cloning channels in SUSE Manager Servers upon deployment
 
@@ -175,6 +180,7 @@ At deploy time the `spacewalk-clone-by-date` will be used for each channel set. 
 
 Activation keys are also automatically created for each clone with the name `1-<CLONE_PREFIX>`.
 
+
 ## Shared resources, prefixing, sharing virtual hardware
 
 Whenever multiple sumaform users deploy to the same virtualization hardware (eg. libvirt host, OpenStack instance) it is recommended to set the `name_prefix` variable in the `base` module in order to have a unique per-user prefix for all resource names. This will prevent conflicting names.
@@ -184,23 +190,60 @@ Additionally, it is possible to have only one user to upload images and other sh
  * make sure there is exactly one user that does not have the variable set, make sure this user has no `name_prefix` set. This user will deploy shared infrastructure for all users
 
 
+## Disabling Avahi and Avahi reflectors
+
+SUSE Manager requires both direct and reverse domain names resolution. This can be provided by either DNS (client-server, unicast mode) or Avahi (peer-to-peer, multicast mode).
+
+Note that Avahi is not available in environments that disable multicast UDP, notably AWS, so the following is only relevant for the libvirt and OpenStack backends. Backends which support multicast UDP have Avahi enabled by default.
+
+Avahi can be disabled if it is not needed. A typical example is a libvirt environment in bridged networking mode where all VMs have static MAC addresses and DNS names known in advance. Avahi can be disabled via something like:
+
+```hcl
+module "base" {
+  ...
+  use_avahi = false
+  domain = "mgr.suse.de"
+  ...
+}
+
+module "suma3pg" {
+  ...
+  mac = "42:54:00:00:00:66"
+  ...
+}
+```
+
+If Avahi is enabled and you are running Docker on a minion, you will need an Avahi reflector on the minion to provide Avahi resolution inside of the containers. A typical example is the Cucumber testsuite which uses such a setup. An Avahi reflector can be enabled via:
+
+```hcl
+module "minion" {
+  ...
+  avahi_reflector = true
+  ...
+}
+```
+
+Beware this may trigger [a known Avahi bug](https://github.com/lathiat/avahi/issues/117). This bug causes wrong names to be assigned to hosts (with unexpected `-2`, `-3`, etc. suffixes) in many circumstances, in particular when more than one reflector is in use on the same network.
+
+
 ## Additional network and SUSE Manager for Retail
 
 You may get an additional, isolated, network, with neither DHCP nor DNS by specifying for example:
 
 ```hcl
 module "base" {
-  [...]
+  ...
   additional_network = "192.168.5.0/24"
-  [...]
+  ...
 }
 ```
 
-This will create a network named `private`, with your prefix in front of the name (e.g. `hmu-private`).
+This will create a network named `private`, with your prefix in front of the name (e.g. `prefix-private`).
 
 You may use that additional network to test SUSE Manager for Retail with the test suite or manually.
 
 For each VM, you can decide whether it connects to the base network and/or to the additional network by specifying:
+
 ```hcl
 connect_to_base_network = false
 connect_to_additional_network = true
@@ -218,15 +261,16 @@ If you want to use another key for all VMs, specify the path of the public key w
 
 ```hcl
 module "base" {
-  [...]
+  ...
   ssh_key_path = "~/.ssh/id_mbologna_terraform.pub"
-  [...]
+  ...
 }
 ```
 
 The `ssh_key_path` option can also be specified on a per-host basis. In this case, the key specified is treated as an additional key, copied to the machine as well as the `ssh_key_path` specified in the `base` section.
 
 If you don't want to copy any ssh key at all (and use passwords instead), just supply an empty file (eg. `ssh_key_path = "/dev/null"`).
+
 
 ## SSH access without specifying a username
 
@@ -238,6 +282,7 @@ StrictHostKeyChecking no
 UserKnownHostsFile=/dev/null
 User root
 ```
+
 
 ## Activation Keys for minions
 
@@ -254,6 +299,7 @@ module "min" {
   activation_key = "1-DEFAULT"
 }
 ```
+
 
 ## Proxies
 
@@ -304,6 +350,7 @@ module "proxy" {
   minion = false
 }
 ```
+
 
 ## Inter-Server Sync (ISS)
 
@@ -367,6 +414,7 @@ ssh server.tf.local run-pts --patching-only
 
 It is also possible to specify non-default hostnames and MAC addresses, see `pts/variables.tf`.
 
+
 ## Cucumber testsuite
 
 It is possible to run [the Cucumber testsuite for SUSE Manager and Uyuni](https://github.com/uyuni-project/uyuni/tree/master/testsuite) by using the `main.tf.libvirt-testsuite.example` file. This will create a test server, client and minion instances, plus a coordination node called a `controller` which runs the testsuite.
@@ -412,7 +460,7 @@ You can also select an alternative fork or branch where for the Cucumber testsui
 
 As an example:
 
-```
+```hcl
 module "controller" {
   source = "./modules/libvirt/controller"
   base_configuration = "${module.base.configuration}"
@@ -426,12 +474,13 @@ module "controller" {
 
 You can also use Docker and Kiwi profiles other than the ones embedded in the test suite:
 
-```
+```hcl
 module "controller" {
   git_profiles_repo = "https://github.com#mybranch:myprofiles"
 }
 
 ```
+
 
 ## Working on multiple configuration sets (workspaces) locally
 
@@ -449,10 +498,11 @@ local_workspaces/libvirt-testsuite
 local_workspaces/libvirt-testsuite/main.tf
 local_workspaces/libvirt-testsuite/terraform.tfstate
 ~/sumaform$ ls -l main.tf
-[...] main.tf -> local_workspaces/libvirt-testsuite/main.tf
+... main.tf -> local_workspaces/libvirt-testsuite/main.tf
 ~/sumaform$ ls -l terraform.tfstate
-[...] -> local_workspaces/libvirt-testsuite/terraform.tfstate
+... -> local_workspaces/libvirt-testsuite/terraform.tfstate
 ```
+
 
 ## Plain hosts
 
@@ -469,6 +519,7 @@ module "vanilla" {
   image = "sles12sp1"
 }
 ```
+
 
 ## PXE boot hosts
 
@@ -489,6 +540,7 @@ module "pxeboot"
 }
 ```
 
+
 ## `minionswarm` hosts
 
 It is possible to create large numbers of simulated minions using Salt's [minionswarm test script](https://docs.saltstack.com/en/latest/topics/releases/0.9.9.html#minionswarm).
@@ -508,6 +560,7 @@ module "minionswarm" {
 
 This will create 400 minions on 2 swarm hosts. Currently only SLES 12 SP1 with the released Salt version are supported.
 
+
 ## SMT
 
 You can configure SUSE Manager instances to download packages from an SMT server instead of SCC, in case a `mirror` is not used:
@@ -522,6 +575,7 @@ module "suma3pg" {
   smt = "http://smt.suse.de"
 }
 ```
+
 
 ## Add custom repos and packages
 
@@ -546,6 +600,7 @@ module "minsles12sp1" {
 }
 ```
 
+
 ## Add custom repo GPG keys
 
 If you need extra GPG keys to be installed for package installation, you can add them via the `gpg_keys` list variable to a module.
@@ -560,6 +615,7 @@ module "minssh-sles12sp2" {
   gpg_keys = ["default/gpg_keys/galaxy.key"]
 }
 ```
+
 
 ## Prometheus/Grafana monitoring
 
@@ -676,6 +732,7 @@ module "minion" {
 }
 ```
 
+
 ## Use Locust for http load testing
 
 You can deploy a locust host to test http performance of your SUSE Manager Server. An example would be:
@@ -715,6 +772,7 @@ module "locust" {
 }
 ```
 
+
 ## Use Operating System updates (released and unreleased)
 
 It is possible to run SUSE Manager servers, proxies, clients and minions with the latest packages of the operating system (for now, only SLE is supported) instead of outdated ones, including updates currently in QAM, that is, upcoming updates. This is useful to spot regressions early, and can be activated via the `use_os_released_updates` (respectively `use_os_unreleased_updates`) flag. Libvirt example:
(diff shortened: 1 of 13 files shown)


Bischoff:
@moio I think I have a minimal version now. All tests green so far, but I'll do some more today. Is a first review possible?

README_ADVANCED.md
```diff
@@ -55,7 +55,8 @@ module 'srv' {
 }
 ```
 
-## Changing OSs
+
+## Changing OS's
```
moio:
```suggestion
## Changing OSs
```

I am not an English native speaker, but this sounds like incorrect grammar to me - this is just a plural, not an https://en.wikipedia.org/wiki/English_possessive

README_ADVANCED.md
```diff
@@ -184,15 +190,40 @@ Additionally, it is possible to have only one user to upload images and other sh
  * make sure there is exactly one user that does not have the variable set, make sure this user has no `name_prefix` set. This user will deploy shared infrastructure for all users
 
 
+## Avahi and Avahi reflectors
```
moio: I think this should be moved/integrated with `modules/libvirt/README.md` where Avahi is introduced.

(As a side note FIY but totally unrelated to this PR: we should probably move that file here in the root directory of the project. People do not find it)

main.tf.libvirt-testsuite.example
```diff
@@ -13,6 +13,7 @@ module "base" {
   // pool = "default"
   // network_name = "default" // change to "" if you change bridge below
   // bridge = ""
+  // use_avahi = false
```
moio:
```suggestion
  // use_avahi = false // uncomment if avahi is not needed (you have proper DNS/DHCP configured in your network)
```

main.tf.libvirt-testsuite.example
```diff
@@ -78,6 +79,8 @@ module "min-sles12sp3" {
   server_configuration = { hostname = "srv.tf.local" } // make sure to prepend the name_prefix, if used
   auto_connect_to_master = false
   ssh_key_path = "./salt/controller/id_rsa.pub"
+  // if you use Avahi, and Docker containers on the minion
+  // avahi_reflector = true
```
moio: I would uncomment this as avahi is on by default (see base module).

salt/minion/reflector.sls
```diff
@@ -0,0 +1,28 @@
+{% if grains['use_avahi'] and grains.get('avahi_reflector') %}
```
moio: Small remark: if the reflector is only really supported on a given OS (family) at this time, please make this explicit here and in the documentation.

README_ADVANCED.md
Bischoff: It's not a possessive but a plural. I checked an online resource that said "add an apostrophe if it makes reading easier" but it seems non-consensual, see for example https://painintheenglish.com/case/333/ . I'll revert.

README_ADVANCED.md
Bischoff: At least I did not :smile_cat: . But there is a more important problem:
```
modules/openstack/host/main.tf:use_avahi: ${var.base_configuration["use_avahi"]}
modules/openstack/minion/main.tf:avahi_reflector: ${var.avahi_reflector}
modules/openstack/minion/variables.tf:variable "avahi_reflector" {
```

If we support Avahi in openstack module, then it's another reason to move it to root directory!

I think I will move it as part of this PR.

Speaking about architecture of the READMEs, what about splitting the `README_ADVANCED.md` that is growing pretty big into a proper `documentation/` directory, with one chapter (and file) for each topic? Maybe that one in a separate PR.

moio: > If we support Avahi in openstack module, then it's another reason to move it to root directory!

We do, and at the moment the Avahi part is copy-paste-adapted between the two. Maybe it's time for a README_AVAHI.md.

> Speaking about architecture of the READMEs, what about splitting the README_ADVANCED.md that is growing pretty big into a proper documentation/ directory, with one chapter (and file) for each topic? Maybe that one in a separate PR.

I totally agree it now needs splitting. Not sure if that fine-grained, possibly in a goal-oriented way. I am told by documentation professionals that this kind of material is best organized around per-topic "HOWTOs", so to speak. I agree this is best discussed separately.

main.tf.libvirt-testsuite.example
Bischoff: No. `false` is the default. The convention so far for the examples is to present the defaults commented.

If you want me to add a comment, then it should read:
```
// uncomment and change to true if avahi is needed (you do not have proper ...
```

main.tf.libvirt-testsuite.example
Bischoff: See above, I was thinking wrongly the default was on `false`. My bad.

salt/minion/reflector.sls
Bischoff: Again a good remark :smile_cat: . Well, according to my code it could work on CentOS too. The only potential problem I can see is different package names in the `pkg.latest`section.

I did not check because we don't test containers on CentOS. Again our f***ing absence of orthogonality in the test suite.

What should I do? Check it works even if it has no practical applications? Trust my code blindly and only make the `pkg.latest` code SUSE-specific?

README_ADVANCED.md
Bischoff: Yes, I was thinking at goal-oriented approach, and even naming this `HOWTO/` instead of `documentation`.

Feel free to throw a task card at me in `SUSE/spacewalk` issues for sumaform documentation reorganisation (and be reassured, I will not start the works before we exactly agree on what we want to do).

Bischoff: I reworked the 3 files `README_ADVANCED.md`,  `modules/libvirt/README.md`, and ` modules/openstack/README.md`, trying to sort the basic "how to connect" into the two module READMEs, and the more advanced HOWTO stuff into `README_ADVANCED.md`.

salt/minion/reflector.sls
Bischoff: It now works on all 3 platforms. I simply removed the update.

Bischoff:
@moio I have both two bad news and two good news.

The bad news are:
 - I can now reproduce the avahi bug systematically: it is enough to add several reflectors to the same test suite.
 - None of my workaround works, they just change a bit the probability of the bug to occur.

The good news are:
 - If you use reflectors reasonably, you will not fall into the bug. I have used several test suites in parallel the whole week without seeing it.
 - We do not need reflectors in production, as we have DHCP and DNS there.

I would therefore suggest to merge this PR, and provide a few words of advice to anyone who wants to use this. The benefit seems greater than the drawbacks.

If it really goes wrong, I would have a fix for Avahi, but it is a bit radical, and it would need to be installed on every client (not just only the minion): I could just comment out all Duplicate Address Detection code... Or we can revert.

What do you think?

moio:
Before formulating an opinion, I would like to learn more about:
 - what do you consider a "reasonable use of reflectors" to be, vs. an "unreasonable" one
 - what would be the side effects of disabling the duplicate address detection code

Bischoff:
Hi @moio,

> Before formulating an opinion, I would like to learn more about:
> 
> * what do you consider a "reasonable use of reflectors" to be, vs. an "unreasonable" one

I think it is reasonable if there is only one VM using it in a given test suite.

Since it is needed only on the container build host (in the normal container), that's not too hard a constraint.

Also, if people don't need to test containers, it would be a good idea to refrain from using them.

Here are the probabilities:
 * if we don't use reflectors, or use reflectors on different networks, then the bug can appear, several people reported it, but in the test suite context, we never saw it, so probability is next to 0.
 * if we use one reflector per suite (on the minion), and the suites run on the same level 2 network: then with the version upgrade contained in this PR, I did not reproduce it in one week, but I would say probability to trigger the bug exists.
 * if we use more than one reflector per suite, then the bug is next to certain to be reproduced.
 

> * what would be the side effects of disabling the duplicate address detection code

Not many, we are not supposed to use two machines with the same domain name.

The biggest problem with that is that I am new to Avahi source code, so it could take me time to figure out to do it properly, and the result could be clumsy.

main.tf.libvirt-testsuite.example
```diff
@@ -78,6 +79,8 @@ module "min-sles12sp3" {
   server_configuration = { hostname = "srv.tf.local" } // make sure to prepend the name_prefix, if used
   auto_connect_to_master = false
   ssh_key_path = "./salt/controller/id_rsa.pub"
+  // if you kept Avahi enabled, and use Docker containers on the minion
```
moio:
```suggestion
```

main.tf.libvirt-testsuite.example
```diff
@@ -78,6 +79,8 @@ module "min-sles12sp3" {
...
+  // avahi_reflector = true
```
moio:
```suggestion
  // avahi_reflector = true // set to false if you have use_avahi = true and do not need container tests
```

modules/libvirt/README.md
```diff
@@ -26,24 +26,27 @@
 
 ## Accessing VMs
 
-All machines come with avahi's mDNS configured by default on the `.tf.local` domain, and user `root` with password `linux` accessible via your SSH public key (by default `~/.ssh/id_rsa.pub`).
+All machines come with user `root` with password `linux`. They are also accessible via your SSH public key (by default `~/.ssh/id_rsa.pub`) if you have one.
 
-Thus if your host is on the same network segment of the virtual machines you can simply use:
+By default, the machines use Avah (mDNS), and are configured on the `.tf.local` domain. Thus if your host is on the same network segment of the virtual machines you can simply use:
```
moio:
```suggestion
By default, the machines use Avahi (mDNS), and are configured on the `.tf.local` domain. Thus if your host is on the same network segment of the virtual machines you can simply use:
```

modules/libvirt/README.md
```diff
@@ -26,24 +26,27 @@
...
 ```
-ssh root@moio-suma3pg.tf.local
+ssh root@hmu-suma3pg.tf.local
```
moio:
```suggestion
ssh root@suma3pg.tf.local
```

modules/libvirt/README.md
```diff
@@ -26,24 +26,27 @@
...
 
 Web access is on standard ports, so `firefox suma3pg.tf.local` will work as expected. SUSE Manager default user is `admin` with password `admin`.
 
-Avahi can be disabled if it is not needed (bridged networking mode, all VMs with static MAC addresses and names known in advance). In that case use the following variables in the `base` module:
-```hcl
-use_avahi = false
-domain = "mgr.suse.de"
+Finally, the images come with serial console support, so you can use
 ```
+virsh console hmu-suma3pg
```
moio:
```suggestion
virsh console suma3pg
```

modules/openstack/README.md
```diff
@@ -20,18 +20,23 @@
...
-Thus if your host is on the same network segment of the virtual machines you can simply use:
+By default, the machines use Avahi (mDNS), and are configured on the `.tf.local` domain. Thus if your host is on the same network segment of the virtual machines you can simply use:
 ```
 ssh root@susemanager-suma31pg.tf.local
 ```
 
-Otherwise, you can look in the OpenStack admin Web UI for the floating IPs.
+If you use Avahi and are on another network segment, you can only connect using an IP address, because mDNS packets do not cross network boundaries unless using reflectors.
+
+To determine the IP addresses, you can look in the OpenStack admin Web UI for the floating IPs.
```
moio:
```suggestion
IP addresses, called floating IPs, are outputted by `terraform apply` and can also be seen in the OpenStack admin Web UI.
```

modules/openstack/README.md
```diff
@@ -20,18 +20,23 @@
...
 
 If you want to use a different SSH key, or mDNS does not work out of the box, or if you don't want to use mDNS, please check the README_ADVANCED.md and TROUBLESHOOTING.md files.
 
-Web access is on standard ports, so `firefox <IP_HOST>` will work as expected. SUSE Manager default user is `admin` with password `admin`.
+Web access is on standard ports, so `firefox <FLOATING_IP_ADDRESS>` will work as expected. SUSE Manager default user is `admin` with password `admin`.
+
+Finally, you should be able to use the Openstack web console to access your VM.
```
moio:
```suggestion
Finally, you can use the OpenStack web console to access your VM.
```

salt/minion/reflector.sls
```diff
@@ -0,0 +1,38 @@
+{% if grains.get('use_avahi') and grains.get('avahi_reflector') %}
+
+# We upgrade to latest version because of a bug in Avahi
```
moio: Please add URL(s) here.

README_ADVANCED.md
```diff
@@ -184,15 +190,48 @@ Additionally, it is possible to have only one user to upload images and other sh
  * make sure there is exactly one user that does not have the variable set, make sure this user has no `name_prefix` set. This user will deploy shared infrastructure for all users
 
 
+## Disabling Avahi and Avahi reflectors
+
+SUSE Manager requires both direct and reverse domain names resolution. This can be provided by either DNS (client-server, unicast mode) or Avahi (peer-to-peer, multicast mode).
```
moio: This is not exactly true, as in some providers (notably AWS) multicast UDP is blocked, so Avahi can't possibly work.

So either we make it clear that this is only relevant for Libvirt and Openstack, or we duplicate this section in the respective READMEs.

In the past I chose the latter, as those are more of "first installation guides" while this file is more about "what can I do now that I have a setup". Normally one has to decide networking details first.

A compromise would be to add a reference from specific READMEs to this section (along the lines of "see "Disabling Avahi and Avahi reflectors" in README_ADVANCED.md for alternate networkig setups").

modules/libvirt/README.md
Bischoff: Thanks.

Bischoff:
> I agree with the substance of the PR, now started some nitpicking over form.

This morning I reproduced the Avahi duplicate names bug by using a reflector on two different test suites on two different servers of the same network. The difference with last week is that this time both suites were using head. Last week one was head and the other 3.2, and it was never reproduced.

So it looks like the operating system versions plays a role. Not sure exactly how. But the fact is, I trigger it only when I run the *same* version of the test suite.

I will not enter such details when I will send my email. I will just advertise this simplified limitation: "For the moment, please run only one reflector per local network. A good idea is to use reflectors only when you need to test containers locally."

BTW, this is a nice way to trigger denials of service on SLE systems. It's therefore a security issue ;-) . But I will wait until I understand it more before I file the SLE bug. Or?

README_ADVANCED.md
Bischoff: I will add a note about this being blocked in AWS.

Reason is that you declare in the provider-specific READMEs "for details on how to change this, please refer to README_ADVANCED.md" (or something along this line). I would like to stick to that redactional choice.

main.tf.libvirt-testsuite.example
Bischoff: Uh ? looks like a leftover. Thanks.

main.tf.libvirt-testsuite.example
Bischoff: OK

modules/libvirt/README.md
Bischoff: Yes, I removed moio because I think we should not use real names in examples.

I used the standard German "hmu-" (Hans Mustermann). It might be obscure to international readers.

Maybe we should still give a hint to use the prefix. I think I will change it to `ssh root@prefix-suma3pg.tf.local`...

modules/openstack/README.md
Bischoff: OK (but I'm not sure "outputted" exists. I think the irregular past of "put" is "put" :tongue: .

modules/openstack/README.md
Bischoff: Yup.

salt/minion/reflector.sls
Bischoff: You mean, add package numbers in the comment?

I'm not sure I understand.

Bischoff:
@moio nitpicks addressed, you may review again.

modules/libvirt/README.md
moio: To me we can also keep the examples minimal and avoid the prefix.

modules/openstack/README.md
moio: Whoops :man_facepalming:

main.tf.libvirt-testsuite.example
```diff
@@ -78,6 +79,7 @@ module "min-sles12sp3" {
   server_configuration = { hostname = "srv.tf.local" } // make sure to prepend the name_prefix, if used
   auto_connect_to_master = false
   ssh_key_path = "./salt/controller/id_rsa.pub"
+  // avahi_reflector = true  // uncomment if you kept Avahi enabled, and use Docker containers on the minion
```
moio:
```suggestion
  avahi_reflector = true  // comment if Avahi is not used, or Docker container tests are not run
```

Sorry, I think I got confused on this one again. If `use_avahi` is true by default and `avahi_reflector` is false by default, we should turn it on in order to have a turnkey working testsuite, right?

main.tf.openstack-testsuite.example
```diff
@@ -82,9 +82,10 @@ module "min-sles12sp3" {
   product_version = "3.1-nightly"
   name = "min-sles12sp3"
   image = "sles12sp3"
-  server_configuration = { hostname = "srv.tf.local" }
+  server_configuration = { hostname = "srv.tf.local" } // make sure to prepend the name_prefix, if used
   auto_connect_to_master = false
   ssh_key_path = "./salt/controller/id_rsa.pub"
+  // avahi_reflector = true // uncomment if you kept Avahi enabled, and use Docker containers on the minion
```
moio:
```suggestion
  avahi_reflector = true  // comment if Avahi is not used, or Docker container tests are not run
```

README_ADVANCED.md
```diff
@@ -184,15 +190,50 @@ Additionally, it is possible to have only one user to upload images and other sh
...
+
+Avahi is not available in environments that disable multicast UDP, notably, so sumaform implements it only for the libvirt and openstack providers.
```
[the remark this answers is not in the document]
Bischoff: OK.

README_ADVANCED.md
```diff
@@ -184,15 +190,50 @@ Additionally, it is possible to have only one user to upload images and other sh
...
+
+Avahi can be disabled if it is not needed (bridged networking mode, all VMs with static MAC addresses and names known in advance). In that case, use the following variables:
```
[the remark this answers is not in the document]
Bischoff: It's a bit losing track of what we want to say. I'll come up with a compromise.

README_ADVANCED.md
```diff
@@ -184,15 +190,50 @@ Additionally, it is possible to have only one user to upload images and other sh
...
+}
+
+module "suma3pg" {
+  ...
+  mac = "42:54:00:00:00:66"
+  ...
+}
+```
+
+You might need an Avahi reflector on the minion if you want to forward Avahi to Docker containers on that minion:
```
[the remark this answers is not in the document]
Bischoff: OK

README_ADVANCED.md
```diff
@@ -184,15 +190,50 @@ Additionally, it is possible to have only one user to upload images and other sh
...
+```hcl
+module "minion" {
+  ...
+  avahi_reflector = true
+  ...
+}
+```
+
+Beware this sometimes triggers a known upstream Avahi bug: https://github.com/lathiat/avahi/issues/117 .
```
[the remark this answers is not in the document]
Bischoff: OK

README_ADVANCED.md
```diff
@@ -184,15 +190,50 @@ Additionally, it is possible to have only one user to upload images and other sh
...
+To be on the safe side, avoid running too many reflectors on the same network.
```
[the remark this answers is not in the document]
Bischoff: As you wish

main.tf.libvirt-testsuite.example
Bischoff: ... and to be at risk of triggering the bug.

Therefore I would not advertise this in the example files. It's better to fail on container tests.

main.tf.openstack-testsuite.example
Bischoff: Same, no.

modules/libvirt/README.md
Bischoff: OK.

main.tf.libvirt-testsuite.example
moio: Doesn't this mean any newcomer will have broken tests by default, and it may take him/her significant time before discovering why? Or is the testsuite really skipping container tests if avahi_reflector is false here?

Bischoff: It indeeds mean that.

We could either document it in the test suite, or use the environment variables / tag mechanism to skip the container tests if we are using avahi without a reflector.

I would prefer a simple comment that would direct the newcomer to the issue. Rationale:

1 - if we skip those tests, we might miss silently container problems, so it's not really better than having them fail all the time ("false green" syndrom)
2 - this is not meant to stay forever so (or at least, I hope...), so it would be a bit of overkill

What do you think?

PR #533, opened 2019-05-11
[WIP] possibility to disable IPv6 in the network

mcalmer:
This variable provide the possibility to disable IPv6 on network interfaces.
The default is to have an unchanged system which means typically IPv6 enabled.

### DIFF
--- a/salt/default/init.sls
+++ b/salt/default/init.sls
@@ -80,7 +80,7 @@ enable_ipv6_all:
     - name: net.ipv6.conf.all.disable_ipv6
     - value: 0
 
-{% else % }
+{% else %}
 
 enable_ipv6_accept_ra:
   sysctl.present:
--- a/salt/default/init.sls
+++ b/salt/default/init.sls
@@ -69,14 +69,33 @@ disable_ipv6_accept_ra:
     - name: net.ipv6.conf.default.accept_ra
     - value: 0
 
+{# disable/enable is just a trick make accept_ra work #}
 disable_ipv6_all:
   sysctl.present:
     - name: net.ipv6.conf.all.disable_ipv6
     - value: 1
 
-disable_ipv6_default:
+enable_ipv6_all:
   sysctl.present:
-    - name: net.ipv6.conf.default.disable_ipv6
+    - name: net.ipv6.conf.all.disable_ipv6
+    - value: 0
+
+{% else % }
+
+enable_ipv6_accept_ra:
+  sysctl.present:
+    - name: net.ipv6.conf.default.accept_ra
     - value: 1
+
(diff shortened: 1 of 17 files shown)


mcalmer:
Currently I am running tests with this parameter.

mcalmer:
My problem was that just setting accespt_ra to 0 did not change anything. Does it need a reboot or a network restart to get active ?

Bischoff:
@mcalmer:
1. "accept_ra" won't work with "all", you have to use it with "eth0"
2. it will prevent new IPv6 global addresses and routes from appearing, but it will not remove existing ones.

During a boot, since the addresses and root do normally not exist yet, it should not be a problem.

Otherwise, either you remove them manually (`ip a del` ...) or you wait for them to be deprecated and removed (about one hour).

mcalmer:
@Bischoff - it works if you disable/enable IPv6. But now I am not happy about the name of the option.
It is not urgent. I think I need to do some more changes here when I have some more time.

salt/default/init.sls
```diff
@@ -61,3 +61,41 @@ authorized_keys:
 {% endfor %}
     - makedirs: True
 {% endif %}
+
+{% if not grains.get('ipv6', 1) %}
+
+disable_ipv6_accept_ra:
+  sysctl.present:
+    - name: net.ipv6.conf.default.accept_ra
```
Bischoff: I don't think this will work. `default` in `/proc/sys/net` means "for newly created network interfaces", if I recall correctly. `eth0` already exists when you do this.

Just use `eth0`.

salt/default/init.sls
```diff
@@ -61,3 +61,41 @@ authorized_keys:
...
+enable_ipv6_all:
+  sysctl.present:
+    - name: net.ipv6.conf.all.disable_ipv6
+    - value: 0
+
+{% else % }
+
+enable_ipv6_accept_ra:
+  sysctl.present:
+    - name: net.ipv6.conf.default.accept_ra
```
Bischoff: Same problem.

salt/default/init.sls
mcalmer: Well, I tried it and it worked. fe80: addresses were shown and hiding like expected.

Bischoff: Impossible, `fe80` addresses are not managed by Router Announcements, controlled by the ` accept_ra` setting above.

What you saw working was probably the global `disable_ipv6` setting.

Bischoff:
***CLOSING***

The situation has changed, we now need it in emergency, I will finish it in a branch of my own and close it. Sorry for the brutal method.

Bischoff:
Closing in favour of #550

PR #534, opened 2019-05-13
If "use_avahi = false", then disable Avahi

Bischoff:
If we set up
```
  use_avahi = false
```

then it's a good idea to disable Avahi:
1. it can only improve the performance :smile_cat: ;
2. it avoids triggering some product bugs, like the dependence of base Salt grains towards name resolution (https://bugzilla.suse.com/show_bug.cgi?id=1134860);
3. it seems coherent with the expected semantics.

Note: it is **not** a fix for the presence ping problem, only a workaround.

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -1,6 +1,7 @@
 include:
   - default.hostname
 
+{% if grains['use_avahi'] %}
 avahi_pkg:
   pkg.installed:
     {%- if grains['os_family'] == 'Debian' %}
@@ -38,7 +39,7 @@ nsswitch_enable_mdns:
     - pattern: "(hosts: .*?)mdns([46]?)_minimal(.*)"
     - repl: "\\1mdns\\2\\3"
 
-avahi_service:
+avahi_enable_service:
   service.running:
     - name: avahi-daemon
     - require:
@@ -46,3 +47,18 @@ avahi_service:
       - file: nsswitch_enable_mdns
     - watch:
       - file: /etc/avahi/avahi-daemon.conf
+
+{% else %}
+
+nsswitch_disable_mdns:
+  file.replace:
+    - name: /etc/nsswitch.conf
+    - pattern: "(hosts: .*?)mdns([46]?)_minimal \\[NOTFOUND=return\\](.*)"
+    - repl: "\\1\\3"
+
+avahi_disable_service:
+  service.dead:
+    - name: avahi-daemon
+    - enable: false
+
+{% endif %}
--- a/salt/default/minimal.sls
+++ b/salt/default/minimal.sls
@@ -3,9 +3,7 @@ include:
   {% if grains['hostname'] and grains['domain'] %}
   - default.hostname
   {% endif %}
-  {% if grains['use_avahi'] %}
   - default.avahi
-  {% endif %}
 
 minimal_package_update:
   pkg.latest:


Bischoff:
Tested with and without Avahi, seems to work as expected.

salt/default/avahi.sls
```diff
@@ -38,11 +39,26 @@ nsswitch_enable_mdns:
...
+avahi_enable_service:
   service.running:
     - name: avahi-daemon
     - require:
       - file: mdns_declare_domains
       - file: nsswitch_enable_mdns
     - watch:
       - file: /etc/avahi/avahi-daemon.conf
+
+{% else %}
```
moio: Is this really necessary? Base images should not have avahi on by default.

Bischoff: They have.

Bischoff:
> Sorry, I do not really see the point of this change. Likely I am missing something...

Well, I was surprised to see yesterday that our production test machines, with `use_avahi = false`:
* have avahi installed
* are running an `avahi-daemon` service
* are running an `avahi-daemon` process
* and are configured to use Avahi for name resolution.

This is what this PR resolved.

Avahi running on those machines was triggering the presence ping bug. Since this PR is in production, we don't have it anymore.

The problem of the presence ping is twice:

1. if Avahi is running, and if you have `2620:` global unicast IPv6 addresses, then it takes ~5 seconds to resolve each `fe80:` link-local unicast IPv6 address. If Avahi is not running, or if you do not have global unicast addresses, then the resolution of link-local addresses is immediate. I know it sounds crazy, but that's so.

2. we have `fqdns` in our base grains, but it is legitimate to take time to perform name resolution. We should not have such grains, especially that some processes like the presence ping requires immediate answer.

moio:
Thanks for clarifying, the PR is perfectly acceptable given this background.

Just out of curiosity, you mentioned:

> If Avahi is running, and if you have 2620: global unicast IPv6 addresses, then it takes ~5 seconds to resolve each fe80: link-local unicast IPv6 address. If Avahi is not running, or if you do not have global unicast addresses, then the resolution of link-local addresses is immediate. I know it sounds crazy, but that's so.

Do we have any idea about why this apparently-bizarre behaviour happens? Is it because of Avahi design, some likely implementation bug, the kernel...?

Bischoff:
> Just out of curiosity, you mentioned:
> 
> > If Avahi is running, and if you have 2620: global unicast IPv6 addresses, then it takes ~5 seconds to resolve each fe80: link-local unicast IPv6 address. If Avahi is not running, or if you do not have global unicast addresses, then the resolution of link-local addresses is immediate. I know it sounds crazy, but that's so.
> 
> Do we have any idea about why this apparently-bizarre behaviour happens? Is it because of Avahi design, some likely implementation bug, the kernel...?

I might have a possible explanation:

It looks like:
 * if you have only fe80 addresses, Avahi has no choice, so that's the ones avahi would associate with the domain name; so the inverse resolution succeeds
 * if you have 2620 addresses, it seems logical Avahi prefers these ones (they are global after all), and the domain name is associated to them. As a consequence, the reverse resolution for 2620 works, but the one for fe80 fails.

No proof of that, it's a mere theory.

Ping @meaksh .

lkotek:
> Well, I was surprised to see yesterday that, with use_avahi = false, our production test machines:
> 
>     have avahi installed
>     are running an avahi-daemon service
>     are running an avahi-daemon process
>     and are configured to use Avahi for name resolution.
> 
> This is what this PR resolved.

@Bischoff I took a short look at the PR, because I encountered something (maybe) related today. Don't we also need to disable `avahi-daemon.socket`? (This should avoid avahi to be triggered.)

Bischoff:
> @Bischoff I took a short look at the PR, because I encountered something (maybe) related today. 
>
> Don't we also need to disable `avahi-daemon.socket`? (This should avoid avahi to be triggered to 
> start.)

@moio Normally, a service marked as "disabled" should not start at reboot, see this `eanbaled: false` option that does exactly this:
```
  service.dead:
￼	    - name: avahi-daemon
￼	    - enable: false
```

But I'm not a systemd expert and I don't know what happens if a service is disabled, but triggered via its socket. I think it still should not be started, but I would not bet on that either.

lkotek:
> I don't know what happens if a service is disabled, but triggered via its socket. I think it still should not > be started, but I would not bet on that either.

I thought the same until I simulated this behaviour on sles12sp3 machine. So it is possible to trigger it to start even with service disabled.

moio:
That's surprising, and would welcome a patch to make sure it does not happen.

Bischoff:
@moio @lkotek well, triggering it via a socket sounds like something intentional, the same as typing `systemctl start avahi-daemon` on the command line.

Or is there something that might trigger it accidentally?

I'm wondering if we are not getting too careful here... Just calling `systemctl disable`is pretty standard.

PR #535, opened 2019-05-14
Reduce "tcp_keepalive_idle" value on proxy to workaround connections hangs

meaksh:
This is a workaround that reduce the value for `tcp_keepalive_idle` to 20 seconds instead of 300 (default).

There is currently an issue when the SUSE Manager proxy is configured and registered on SUSE Manager. The `state.apply` which is executed to configure the proxy is doing changes on the network and it's also enabling the firewall on the proxy. It seems this is producing the already established connection from the `salt-minion` process on the proxy and the `salt-master` hangs.

Before this PR, the TCP connection (4505) will get restrablised after 300 seconds, which is making some testsuite scenario to fail.

This PR workaround this issue by reducing the `tcp_keepalive_idle` to 20 seconds. Already tested.

**NOTE:** This issue only happening in our testsuite infrastructure and not locally.

### DIFF
--- a/salt/suse_manager_proxy/init.sls
+++ b/salt/suse_manager_proxy/init.sls
@@ -202,7 +202,7 @@ ca-configuration-checksum:
 
 {% endif %}
 
-reduce_tcp_keepalive_idle_on_proxy:
+TEMPORARY_WORKAROUND_reduce_tcp_keepalive_idle_on_proxy:
   file.managed:
     - name: /etc/salt/minion.d/keepalive.conf
     - makedirs: True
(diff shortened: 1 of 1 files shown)


moio:
Hey @meaksh, thanks a lot for this!

I would only ask you to kindly add a comment starting with `# HACK:`, a one or two line explanation and a reference to the issue that, when closed, signals the workaround can be reverted.

This is just standard practice in sumaform to keep track of workarounds.

Feel free to commit that directly, we do not need a PR for that. Thanks.

PR #536, opened 2019-05-15
Empty variable for http proxy should not be defined

lkotek:
Environment variable `SUMA_HTTP_PROXY` shouldn't be declared in `.bashrc` in case `http_proxy` is undefined or has an empty string as a value in `main.tf`.

### DIFF
--- a/salt/controller/bashrc
+++ b/salt/controller/bashrc
@@ -14,7 +14,7 @@ export VIRTHOST_KVM_URL="{{ grains.get('kvm_host') }}"
 export VIRTHOST_KVM_PASSWORD="linux"
 {% else %}# no KVM host defined {% endif %}
 export GITPROFILES="{{ grains.get('git_profiles_repo') }}"
-export SUMA_HTTP_PROXY="{{ grains.get('http_proxy') }}"
+{% if grains.get('server_http_proxy') | default(false, true) %}export SERVER_HTTP_PROXY="{{ grains.get('server_http_proxy') }}" {% else %}# no server HTTP proxy defined {% endif %}
 
 # Generate certificates for Google Chrome
 if [ ! -f /etc/pki/trust/anchors/$SERVER.cert ]; then
--- a/modules/libvirt/controller/main.tf
+++ b/modules/libvirt/controller/main.tf
@@ -41,7 +41,7 @@ pxeboot_mac: ${var.pxeboot_configuration["macaddr"]}
 role: controller
 branch: ${var.branch == "default" ? lookup(var.testsuite-branch, var.server_configuration["product_version"]) : var.branch}
 git_profiles_repo: ${var.git_profiles_repo == "default" ? "https://github.com/uyuni-project/uyuni.git#master:testsuite/features/profiles" : var.git_profiles_repo}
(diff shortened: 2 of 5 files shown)


salt/controller/bashrc
```diff
@@ -14,7 +14,7 @@ export VIRTHOST_KVM_URL="{{ grains.get('kvm_host') }}"
 export VIRTHOST_KVM_PASSWORD="linux"
 {% else %}# no KVM host defined {% endif %}
 export GITPROFILES="{{ grains.get('git_profiles_repo') }}"
-export SUMA_HTTP_PROXY="{{ grains.get('http_proxy') }}"
+{% if grains.get('http_proxy') | default(false, true) %}export SUMA_HTTP_PROXY="{{ grains.get('http_proxy') }}" {% else %}# no HTTP proxy defined {% endif %}
```
Bischoff: LGTM but what about the following: rename `SUMA_HTTP_PROXY` to `HTTP_PROXY` to spare typing and for consistency between the various variables ?

Bischoff: Maybe not in this PR...

lkotek: Thanks for review :-) Actually I used `HTTP_PROXY` name at first, but this environment variable name is already expected by utilities like `curl`, which may bring unexpected behavior.

lkotek:
@Bischoff  Successfully tested, working as expected.

salt/controller/bashrc
Bischoff: Makes sense. Then please ignore remark above.

moio: How about using SERVER_HTTP_PROXY or some other more product-neutral name? Idea would be to be more welcome to the Uyuni community by gradually removing SUMA-ness (or at least try not to introduce new SUMA-ness...)

Bischoff: Makes sense !

lkotek: Yes, I can use this opportunity to change it at all places mentioned ;-) I'll prepare changes.

lkotek: OK :-) It's necessary to align it with sumaform runner too.

lkotek: Done :-) Changes for testsuite + sumaform runner prepared too.

salt/controller/bashrc
```diff
@@ -14,7 +14,7 @@ export VIRTHOST_KVM_URL="{{ grains.get('kvm_host') }}"
 export VIRTHOST_KVM_PASSWORD="linux"
 {% else %}# no KVM host defined {% endif %}
 export GITPROFILES="{{ grains.get('git_profiles_repo') }}"
-export SUMA_HTTP_PROXY="{{ grains.get('http_proxy') }}"
+{% if grains.get('server_http_proxy') | default(false, true) %}export SERVER_HTTP_PROXY="{{ grains.get('server_http_proxy') }}" {% else %}# no SERVER HTTP proxy defined {% endif %}
```
Bischoff: "no SERVER HTTP proxy" -> "no server HTTP proxy " in the comment

lkotek: Modified :-)

PR #537, opened 2019-05-15
Enable "conntrack" kernal modules on the Proxy to avoid hanging connections

meaksh:
This PR removes the `tcp_keepalive_idle` workaround on the Proxy, and preload the `conntrack` kernel modules to actually avoid the connections to hangs when `SuSEfirewall2` is started.

### DIFF
--- a/salt/suse_manager_proxy/init.sls
+++ b/salt/suse_manager_proxy/init.sls
@@ -201,3 +201,20 @@ ca-configuration-checksum:
       - file: ca-configuration
 
 {% endif %}
+
+
+# HACK: This avoids already established connections to hang
+# when SuSEfirewall2 is started by the retail formula
+preload_conntrack_modules_and_enable_them_at_boottime:
+  file.managed:
+    - name: /etc/modules-load.d/nf_conntrack.conf
+    - content: |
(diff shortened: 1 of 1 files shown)


Bischoff:
> Two questions from my side:
> 
> * Should this only be enabled if this state is run in the context of a testsuite?

Maybe it should be applied to everyone, as the problem can be reproduced outside of the test suite.

> * Is this something real-life users could stumble in? How are we fixing it? Documentation?

Yes. But as @mcalmer said, they would stumble on it only once, when installing a branch server. And even so, it's self-healing after 5 minutes (if standard Salt setting for keepalive is kept).

We are still uncertain about how to "fix" it (post-install script, documentation, ...). We decided to wait  for the opinion of Retail team.

meaksh:
Hey @moio,
this workaround is applied for all proxy sumaform deployments (not only if testsuite is enabled).

About a general solution for this, we've created https://bugzilla.suse.com/show_bug.cgi?id=1135303

moio:
Thanks, I will just add the bug reference to the HACK comment.

Bischoff:
ok!

PR #540, opened 2019-05-17
Expose the testsuite directory to the http port 8080 with a service

ncounter:
Let the `/root/spacewalk/testsuite` directory be accessible via an `http` service.
The service is created and enabled as a systemd service, and it simply execute the python `SimpleHTTPServer` in the testsuite directory exposing the content at the port `8080`.

The aim of this PR is to be able to simply reach the `output.html` result of the testsuite from a remote browser directly  (e.g. for instance `http://ctl.tf.local:8080/output.html`).

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -436,6 +436,9 @@ Once all `core_` features have been executed once other non-core Cucumber featur
 ssh -t head-ctl.tf.local cucumber spacewalk/testsuite/features/my_feature.feature
 ```
 
+Read HTML results at:
+
+ `head-ctl.tf.local:8080/output.html`. There is an additional running service, enabled during the `highstate`, on the `controller` which is exposing the entire `/root/spacewalk/testsuite` folder: all testsuite files, including results saved under this folder, are readable through the `http` protocol at the port `8080`.
 
 Get HTML results with:
 ```


salt/controller/init.sls
```diff
@@ -154,3 +154,18 @@ google_cert_db:
    - name: certutil -d sql:/root/.pki/nssdb -N --empty-password
    - require:
      - file: chrome_certs
+
+http_testsuite:
```
moio:
```suggestion
http_testsuite_service_file:
```

salt/controller/init.sls
```diff
@@ -154,3 +154,18 @@ google_cert_db:
...
+    - user: root
+    - groop: root
+    - mode: 600
+
+http_testsuite_service:
+  service.running:
+    - name: http_testsuite
+    - enable: true
+    - require:
+      - file: http_testsuite
```
moio:
```suggestion
      - file: http_testsuite_service_file
```

salt/controller/http_testsuite.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Spacewalk Testsuite HTTP Service
```
moio:
```suggestion
Description=Testsuite HTTP Service
```

ncounter:
> Thanks! Feel free to squash and merge with the inline nitpicks.

All nitpicks addressed and squashed. Thanks.

> Please also add a README line so that users know about this feature.

Documented [here](https://github.com/moio/sumaform/pull/540/commits/f40483c5c10de596e4da88ad59c7bea80106b13b)

PR #542, opened 2019-05-23
Fix default system locale to "LANG=en_US.UTF-8" for all hosts

meaksh:
This PR sets the default locale to `LANG=en_US.UTF-8` in all sumaform hosts to workaround the inconsistent state where some hosts may not have any default locale assigned.

This situations leads into `UnicodeDecodeErrors` when running `salt-ssh` from a system which does not have any default locale set.

### DIFF
--- a/salt/default/locale.sls
+++ b/salt/default/locale.sls
@@ -1,0 +1,3 @@
+fix_en_US_UTF8_as_system_local:
+  cmd.run:
+    - name: localectl set-locale LANG=en_US.UTF-8
--- a/salt/default/init.sls
+++ b/salt/default/init.sls
@@ -1,4 +1,5 @@
 include:
+  - default.locale
   - default.minimal
   - default.pkgs
   {% if grains.get('reset_ids') | default(false, true) %}


salt/default/locale.sls
```diff
@@ -0,0 +1,3 @@
+fix_en_US_UTF8_as_system_local:
+  cmd.run:
+    - name: localectl set-locale LANG=en_US.UTF-8
```
moio: Do all OSs support `localectl`? I guess SLE 11 might not.

brejoc: I'll create a new issue to check that. Thanks, Moio!

meaksh: @moio - you're right! `localectl` is not available on SLE 11, I missed that! Sorry.

We applied this workaround for setting the locale to all images to keep consistency but this locale issue is actually only affecting the images where we install a SUSE Manager server which deals with `salt-ssh`.

Would you be OK if we apply this locale fix to only those images (SLE12-15 and OpenSUSE 15.0/1)? Or maybe just exclude SLE11.

moio: I would recommend to exclude SLE 11 or, even better, do the equivalent there.

meaksh: Fixed on https://github.com/moio/sumaform/pull/546

PR #546, opened 2019-05-28
Do not use "localectl" on SLE11 systems

meaksh:
This PR configures `en_US.UTF-8` as default locale on SLE11 where `localectl` command is not available.


meaksh:
Hmmm, that's a good point ;-) I'm doing it right now

PR #550, opened 2019-06-01
Make it possible to disable IPv6 (fully or only RAs)

Bischoff:
**Work in progress - not ready yet** 

This PR is written on top of @mcalmer 's #533 PR

The `ipv6` variable is becoming a map:
```
  type = "map"
  default = {
    enable = true
    accept_ra = true
  }
```
I hesitated with a tristate, but the map is more open to future extensions (yes, I know, YAGNI...). Feel free to disagree and find the tristate simpler, it's easy to change.

As those settings are not really provider-specific, I ported them to openstack as well.

The results go to `/etc/sysctl.d/99-salt.conf`:
```
net.ipv6.conf.lo.accept_ra = 0
net.ipv6.conf.eth0.accept_ra = 0
net.ipv6.conf.eth1.accept_ra = 0
net.ipv6.conf.all.disable_ipv6 = 1
```

When disabling RAs:
 - existing dynamic addresses are removed (otherwise, they time out only after a delay that is too long);
 - we also make sure that the network manager does not destroy our efforts: wicked on SLE, networkd on Ubuntu.

**The documentation and examples are still missing**

### DIFF
--- a/modules/libvirt/client/main.tf
+++ b/modules/libvirt/client/main.tf
@@ -11,6 +11,7 @@ module "client" {
   gpg_keys = "${var.gpg_keys}"
   swap_file_size = "${var.swap_file_size}"
   ssh_key_path = "${var.ssh_key_path}"
+  ipv6 = "${var.ipv6}"
   connect_to_base_network = true
   connect_to_additional_network = true
   grains = <<EOF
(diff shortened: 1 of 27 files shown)


Bischoff:
It looks like Travis CI is still testing the initial push...

Bischoff:
Works in production for some time now, merging

PR #563, opened 2019-07-04
HACK: workaround a not yet released bugfix of cucumber-html-reporter

ncounter:
This is a workaround to inject [this bugfix](https://github.com/gkushang/cucumber-html-reporter/pull/181) until it is release into a newer version of the `cucumber-html-reporter` module.


srbarrios:
Very nice approach. 
Let's merge it, and after merged, I remove the hack from sumaform-test-runner

PR #568, opened 2019-07-29
Fix grains variable to work for host

ktsamis:
Before it wasn't working because the definition wasn't there. I tested with a custom grain now and it is applied. It's still a string only, I will see if I can make it a list later.

### DIFF
--- a/modules/libvirt/host/main.tf
+++ b/modules/libvirt/host/main.tf
@@ -99,7 +99,7 @@ connect_to_base_network: ${var.connect_to_base_network}
 connect_to_additional_network: ${var.connect_to_additional_network}
 reset_ids: true
 ipv6: {${join(", ", formatlist("'%s': '%s'", keys(var.ipv6), values(var.ipv6)))}}
-${var.grains}
+grains: ${var.grains}
 
 EOF
 


moio:
var.grains is supposed to contain the grain definition, of any type.

This will break many use cases as:

https://github.com/uyuni-project/sumaform/blob/master/modules/libvirt/controller/main.tf#L29

Can you share the main.tf you are using? I bet the problem is actually there.

ktsamis:
I was using something like:
```
module "12sp4" {
  source = "./modules/libvirt/host"
  count = 3
  name = "12sp4"
  image = "sles12sp4"
  grains = qam
}
```
I wanted to have some custom salt states triggered according to the grain. It probably was the wrong way to define the grain. Can you give me an example for the right one? We had some errors related to it and with changing the definition like the one in the PR it worked. But if it's deliberate (and since you say it is I believe it) then we were using it wrong.

btw this is for a SLE project for QAM that I was helping with, not SUMA

ktsamis:
I will close this actually and we can discuss the proper way to use the custom grains in the host module, but the PR isn't needed

moio:
grains = "grainname: grainvalue"

ktsamis:
Thanks very much, didn't realise it was a dict. I will pass it on to QAM and we see how it will work for them.

PR #569, opened 2019-07-31
Fix baseproduct for SUSE Manager 4

moio:
The `/etc/products.d/baseproduct` symlink was not set correctly.

@mcalmer this is also true for Uyuni, right?


mantel:
> The `/etc/products.d/baseproduct` symlink was not set correctly.
> 
> @mcalmer this is also true for Uyuni, right?

No, this is running on openSUSE. No changes needed there.

PR #576, opened 2019-08-14
Change default RAM size for host, client, and minion

Bischoff:
This PR changes the default memory size from `512 MB` to `768 MB` for generic host, client, and minion in the libvirt module.

Rationale: I noticed very bad responsiveness at the end of SLE 12 SP4 minion at end of 3.2 testsuite. Error count was 92 with 0.5 GB, mostly due to timeouts. I retested exactly same setup with 1.5 GB (yeah, not 0.75 GB like here) and saw the failure count get down to 30.

The change in this PR is rather conservative: production test suites already put 1 GB on all clients, excepted PXE boot minion.

This PR leaves unchanged the explicit setting of 512 MB in `main.tf` for the mirror.

### DIFF
--- a/main.tf.libvirt-testsuite.example
+++ b/main.tf.libvirt-testsuite.example
@@ -65,6 +65,7 @@ module "cli-sles12sp3" {
   product_version = "3.1-nightly"
   name = "cli-sles12sp3"
   image = "sles12sp3"
+  memory = 768
   server_configuration = { hostname = "srv.tf.local" } // make sure to prepend the name_prefix, if used
   auto_register = false
   ssh_key_path = "./salt/controller/id_rsa.pub"
@@ -76,6 +77,7 @@ module "min-sles12sp3" {
   product_version = "3.1-nightly"
   name = "min-sles12sp3"
   image = "sles12sp3"
+  memory = 768
   server_configuration = { hostname = "srv.tf.local" } // make sure to prepend the name_prefix, if used
   auto_connect_to_master = false
   ssh_key_path = "./salt/controller/id_rsa.pub"
@@ -88,6 +90,7 @@ module "minssh-sles12sp3" {
   base_configuration = "${module.base.configuration}"
   name = "minssh-sles12sp3"
   image = "sles12sp3"
+  memory = 768
   ssh_key_path = "./salt/controller/id_rsa.pub"
   gpg_keys = ["default/gpg_keys/galaxy.key"]
 }
@@ -99,6 +102,7 @@ module "min-centos7" {
   product_version = "3.1-nightly"
   name = "min-centos7"
   image = "centos7"
+  memory = 768
   server_configuration = { hostname = "srv.tf.local" } // make sure to prepend the name_prefix, if used
   auto_connect_to_master = false
   ssh_key_path = "./salt/controller/id_rsa.pub"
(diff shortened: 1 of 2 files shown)


Bischoff:
> LGTM. Is a conservative value, I don't think it will hurt to local runs.

It will :cry: , but not as bad as not enough memory... It's a tradeoff.

> Which is weird is that the new Client Tools, or the interaction with server now, it has increased that much.

It could come from `salt-minion` (not sure you consider it as a part of the client tools), I saw each process eats quite a few RAM, and there are several spawn.

Or it could just be operating systems, including SLE 12, eating more and more memory...

I'm not sure about the exact cause.

> It might be a flag to alert the team about this increase of needs.
> Good to have @moio on that PR :)

yes, I think he has the final decision ;-) .

moio:
I would like to have a cause pinned down before proceeding. SLES is supported at 512 MB, and every extra resource will hurt people just trying something minimal in their laptop. I do not mind if testsuites require more - in that case the change should be in the example testsuite files.

Moreover if we really have a reason, that should be reflected across backends, not just libvirt.

Bischoff:
> I would like to have a cause pinned down before proceeding. SLES is supported at 512 MB, and every extra resource will hurt people just trying something minimal in their laptop. I do not mind if testsuites require more - in that case the change should be in the example testsuite files.
> 
> Moreover if we really have a reason, that should be reflected across backends, not just libvirt.

OK. Makes sense. Will redo the PR.

Bischoff:
I did as you suggested @moio, please check.

I'm still not fully convinced. OK, advertised minimum for SLES is 512 MB, but what about SLES + salt-minion? There are quite a few reports out there for high memory consumption on salt minions: https://www.google.de/search?q=salt+minion+memory+usage .

Bischoff:
@moio, this study analyzes https://www.diva-portal.org/smash/get/diva2:1117279/FULLTEXT01.pdf to > 250 MB.

A quick look on 4.0 reference host's minion:
```
suma-ref40-min-sles12sp4:~ # ps aux | grep salt-mi
root      1008  0.0  0.1  10564  1556 pts/0    S+   13:06   0:00 grep --color=auto salt-mi
root      8102  0.0  2.6 166224 26788 ?        Ss   11:39   0:00 /usr/bin/python /usr/bin/salt-minion
root      8107  0.3  6.8 814588 68080 ?        Sl   11:39   0:16 /usr/bin/python /usr/bin/salt-minion
root      8110  0.0  2.3 259164 23620 ?        S    11:39   0:00 /usr/bin/python /usr/bin/salt-minion
suma-ref40-min-sles12sp4:~ # for i in 8102 8107 8110; do pmap $i | tail -n 1; done
18128K writable-private, 148072K readonly-private, 28K shared, and 26416K referenced
117764K writable-private, 696788K readonly-private, 40K shared, and 66624K referenced
30564K writable-private, 228564K readonly-private, 40K shared, and 4752K referenced
```

On same machine, `top` shows me regularly `salt-minion` among the most RAM-intensive processes.

Shouldn't we close this PR and instead open a research card to investigate minion RAM footprint?

I searched in SUSE Manager's documentation and did not find requirements for the client side. Maybe I missed it. Ping @Loquacity ?

Bischoff:
I will change this into an issue. Closing.

PR #578, opened 2019-08-14
Use "owners" field for existing images lookup

Bischoff:
This PR specifies the value of mandatory field `"owners"` when querying available images in the AWS example.

I used value `"amazon"`, which is okay for the community AMIs that I am using for SUSE BYOS images. I am not sure this value is okay for the sumaform images @moio has in mind.

### DIFF
--- a/main.tf.aws.example
+++ b/main.tf.aws.example
@@ -35,6 +35,9 @@ data "aws_ebs_snapshot" "data_disk_snapshot" {
 
 data "aws_ami" "opensuse422" {
   most_recent = true
+  owners = [
+    // add "self" or your AWS account ID to restrict search
+  ]
 
   filter {
     name   = "name"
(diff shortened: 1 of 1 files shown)


moio:
Please make this a parameter, or current instructions will not work as they assume the user uploads the custom image.

Bischoff:
Er, this is already a main.tf parameter, the change is in the main.tf example :smile_cat: .

@moio, you mean I should add a comment, maybe?

(I'm lacking context to write this comment: if the user uploads the image, what should be the value of this `owner` parameter?)

moio:
> Er, `owner` is already a main.tf parameter, the PR changes the main.tf _example_ smile_cat .

Whoops, I need more coffee.

> @moio, you mean I should add a comment, maybe?
> 
> (I'm lacking context to write this comment: if the user uploads the image, what should be the value of this `owner` parameter?)

It should be his/her user id IIRC.

> I think something has changed in AWS, your example was probably working in the past without specifying the owner, but now this parameter is marked as mandatory.

Most probably in Terraform, yes.

Bischoff:
OK, please check new version @moio.

I did my best, using reference documentation at https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-images.html .

It is possible that leaving the list empty just works, I did not check.

PR #579, opened 2019-08-19
Support for QAM clients

lkotek:
This PR adds variables for clients needed by QAM.

### DIFF
--- a/modules/libvirt/controller/main.tf
+++ b/modules/libvirt/controller/main.tf
@@ -46,6 +46,29 @@ branch: ${var.branch == "default" ? lookup(var.testsuite-branch, var.server_conf
 git_profiles_repo: ${var.git_profiles_repo == "default" ? "https://github.com/uyuni-project/uyuni.git#:testsuite/features/profiles" : var.git_profiles_repo}
 server_http_proxy: ${var.server_http_proxy}
 
+sle11sp4_minion: ${var.sle11sp4_minion_configuration["hostname"]}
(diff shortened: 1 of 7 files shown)


srbarrios:
I have the feeling that we need to somehow do it generic to accept other clients in the future.
For now, in my opinion, it will better to have it in your fork or a branch until we succeed in having the whole workflow of a QAM test suite running.

lkotek:
> For now, in my opinion, it will better to have it in your fork or a branch until we succeed in having the whole workflow of a QAM test suite running.

Agreed, I changed target branch to point to `qam-testsuite`.

Bischoff:
> I have the feeling that we need to somehow do it generic to accept other clients in the future. (...)

Yes, not sure how. Maybe put them all in some configuration file. Not sure whether it's a good idea...

srbarrios:
I think we could miss something on those changes. I can see the changes applied in the sumaform used in the jenkins slave, but going to the bashrc of the controller I can't even see the commented lines in case of non existance of those grains.

Follow progress here: 
- https://ci.suse.de/view/Manager/view/Manager-4.0/job/manager-40-cucumber-qam-pipeline/
- CONTROLLER: qam-pip-40-ctl.prv.suse.net 
- https://ci.suse.de/view/Manager/view/Manager-4.0/job/manager-40-cucumber-qam-pipeline/44/execution/node/4/ws/sumaform/salt/controller/bashrc/*view*/

PR #581, opened 2019-08-26
Update run-testsuite to use the new rake tasks

srbarrios:
After the improvements made in the test suite to support parallel cucumber features, that script was outdated.
In fact, I'm not sure if that code was never used before :)

### DIFF
--- a/salt/controller/run-testsuite.sh
+++ b/salt/controller/run-testsuite.sh
@@ -9,7 +9,9 @@ if [ -z "$1" ]; then
 fi
 # this to run cucumber features in parallel
 if [[ $1 = "parallel" ]]; then
-   rake core
-   rake secondary_parallel
-   rake post_run
+   rake cucumber:core
+   rake parallel:init_clients
+   rake cucumber:secondary
+   rake parallel:secondary_parallelizable
+   rake cucumber:finishing
 fi


Bischoff:
> After the improvements made in the test suite to support parallel cucumber features, that script was outdated.

Yes, I noticed the problem, but was waiting until the end of your big restructuring to raise it.

> In fact, I'm not sure if that code was never used before :)

I only tried it once, and noticed the problem, but you had enough to do already and wanted to spare you. Normally, I indeed don't use the `-parallel` flag.

srbarrios:
@Bischoff It might make sense for you if we just remove the flag "parallel", having as default the content currently inside parallel section?

Bischoff:
@srbarrios if it runs in parallel by default, no, as I still run sequentially locally.

Unless you provide a `-sequential`flag of course -- I can survive typing one more flag.

srbarrios:
Ok, I fixed the parallel flag, but in fact, sequentially, will run the old testsuite.yml, if I'm not wrong.
So we might want to fix it, to trigger:

```
   rake cucumber:core
   rake cucumber:init_clients
   rake cucumber:secondary
   rake cucumber:secondary_parallelizable
   rake cucumber:finishing
```

Which will run all the stages sequentially (the namespace :cucumber or :parallel is what makes the difference)

PR #583, opened 2019-08-30
Support openSUSE clients for Uyuni

Bischoff:
* add client tools for openSUSE
* use `mgr-cfg-actions` instead of `rhncfg-actions` for Uyuni client tools
* server is now based on 15.1 based, not on 42.3 anymore

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -24,6 +24,23 @@ os_update_repo:
     - name: /etc/zypp/repos.d/openSUSE-Leap-{{ grains['osrelease'] }}-Update.repo
     - source: salt://repos/repos.d/openSUSE-Leap-Update.repo
     - template: jinja
+
+{% if not grains.get('role') or not grains.get('role').startswith('suse_manager') %}
+{% if 'uyuni-master' in grains.get('product_version') | default('', true) %}
+tools_pool_repo:
+  file.managed:
+    - name: /etc/zypp/repos.d/systemsmanagement_Uyuni_Master_openSUSE_Leap_15-Uyuni-Client-Tools.repo
+    - source: salt://repos/repos.d/systemsmanagement_Uyuni_Master_openSUSE_Leap_15-Uyuni-Client-Tools.repo
+    - template: jinja
+{% elif 'uyuni-released' in grains.get('product_version') | default('', true) %}
+tools_pool_repo:
+  file.managed:
+    - name: /etc/zypp/repos.d/systemsmanagement_Uyuni_Stable_openSUSE_Leap_15-Uyuni-Client-Tools.repo
+    - source: salt://repos/repos.d/systemsmanagement_Uyuni_Stable_openSUSE_Leap_15-Uyuni-Client-Tools.repo
+    - template: jinja
+{% endif %}
+{% endif %}
+
 {% elif grains['osfullname'] == 'SLES' %}
 
 {% if grains['osrelease'] == '11.4' %}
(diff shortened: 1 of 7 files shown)


mcalmer:
Why do we need openSUSE 42? It is unmaintained now.

Bischoff:
> Why do we need openSUSE 42? It is unmaintained now.

I currently need it for my tests because it allows me to avoid the problem of Kiwi-ng not supported that we have on *15 series.

We can remove 42 everywhere (and not only for Uyuni !) after that in a separate PR.

Bischoff:
Deployment passes, now running a first test suite.

salt/repos/default.sls
```diff
@@ -24,6 +24,23 @@ os_update_repo:
     - name: /etc/zypp/repos.d/openSUSE-Leap-{{ grains['osrelease'] }}-Update.repo
     - source: salt://repos/repos.d/openSUSE-Leap-Update.repo
     - template: jinja
+
+{% if not grains.get('role') or not grains.get('role').startswith('suse_manager') %}
+{% if 'uyuni-master' in grains.get('product_version') | default('', true) %}
+tools_pool_repo:
+  file.managed:
+    - name: /etc/zypp/repos.d/systemsmanagement_Uyuni_Master_openSUSE_Leap_{{ grains['osrelease'] }}-Uyuni-Client-Tools.repo
```
moio: @juliogonzalez do we have / will we have a new repo for minor Leap versions, eg. 15.1?

juliogonzalez: No, with Leap it is going to happen the same as for SLE.

15.0 repository will be valid for all 15.X versions (in the same way SLE12 is valid for all SLESPX versions).

moio: Then

```suggestion
    - name: /etc/zypp/repos.d/systemsmanagement_Uyuni_Master_openSUSE_Leap_15-Uyuni-Client-Tools.repo
```

salt/repos/default.sls
```diff
@@ -24,6 +24,23 @@ os_update_repo:
...
+    - source: salt://repos/repos.d/systemsmanagement_Uyuni_Master_openSUSE_Leap_{{ grains['osrelease'] }}-Uyuni-Client-Tools.repo
```
moio:
```suggestion
    - source: salt://repos/repos.d/systemsmanagement_Uyuni_Master_openSUSE_Leap_15-Uyuni-Client-Tools.repo
```

salt/repos/default.sls
```diff
@@ -24,6 +24,23 @@ os_update_repo:
...
+    - template: jinja
+{% elif 'uyuni-released' in grains.get('product_version') | default('', true) %}
+tools_pool_repo:
+  file.managed:
+    - name: /etc/zypp/repos.d/systemsmanagement_Uyuni_Stable_openSUSE_Leap_{{ grains['osrelease'] }}-Uyuni-Client-Tools.repo
```
moio:
```suggestion
    - name: /etc/zypp/repos.d/systemsmanagement_Uyuni_Stable_openSUSE_Leap_15-Uyuni-Client-Tools.repo
```

salt/repos/default.sls
Bischoff: @moio's solution will not work for Leap 16. But we could live with it in the meantime.

Will do.

salt/repos/default.sls
Bischoff: @moio no, it's rather "source" to modify, not "name" ?

PR #585, opened 2019-09-03
Support for QAM clients

lkotek:
This PR adds variables for clients needed for QAM.

### DIFF
--- a/modules/libvirt/controller/main.tf
+++ b/modules/libvirt/controller/main.tf
@@ -42,6 +42,29 @@ branch: ${var.branch == "default" ? lookup(var.testsuite-branch, var.server_conf
 git_profiles_repo: ${var.git_profiles_repo == "default" ? "https://github.com/uyuni-project/uyuni.git#:testsuite/features/profiles" : var.git_profiles_repo}
 server_http_proxy: ${var.server_http_proxy}
 
+sle11sp4_minion: ${var.sle11sp4_minion_configuration["hostname"]}
+sle11sp4_sshminion: ${var.sle11sp4_sshminion_configuration["hostname"]}
+sle11sp4_client: ${var.sle11sp4_client_configuration["hostname"]}
+sle12sp4_minion: ${var.sle12sp4_minion_configuration["hostname"]}
+sle12sp4_sshminion: ${var.sle12sp4_sshminion_configuration["hostname"]}
+sle12sp4_client: ${var.sle12sp4_client_configuration["hostname"]}
+sle15_minion: ${var.sle15_minion_configuration["hostname"]}
+sle15_sshminion: ${var.sle15_sshminion_configuration["hostname"]}
+sle15_client: ${var.sle15_client_configuration["hostname"]}
+sle15sp1_minion: ${var.sle15sp1_minion_configuration["hostname"]}
+sle15sp1_sshminion: ${var.sle15sp1_sshminion_configuration["hostname"]}
+sle15sp1_client: ${var.sle15sp1_client_configuration["hostname"]}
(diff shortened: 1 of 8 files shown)


salt/client/testsuite.sls
```diff
@@ -10,9 +10,14 @@ client_cucumber_requisites:
       - spacewalk-client-setup
       - spacewalk-check
       - spacewalk-oscap
-      - rhncfg-actions
       - openscap-utils
+      {% if grains['os'] == 'CentOS' %}
+      - mgr-cfg-actions
```
srbarrios: I'm confused with that condition, wasn't mgr-cfg-actions the new package that we need to use everywhere?
It was any reason to use only in CentOS?

lkotek: It is a good point. I think you are right. The reason why I modified it this way has just the opposite cause ;-) There was `rhncfg-actions` package missing for CentOS, so I replaced it by its "successor" - mgr-cfg-actions. But what you said is valid, we can probably remove `rhncfg-action` at all and replace it by `mgr-cfg-actions` for all clients. So inside the condition will stay only the `man` and `man-db` package.

lkotek: Changed :-)

lkotek:
> LGTM, but you advice in the "variables" files to look at the test suite example... but I don't see an example in your PR.

You're absolutely right. Let's fix it ;-)

modules/libvirt/controller/main.tf
```diff
@@ -46,6 +46,29 @@ branch: ${var.branch == "default" ? lookup(var.testsuite-branch, var.server_conf
 git_profiles_repo: ${var.git_profiles_repo == "default" ? "https://github.com/uyuni-project/uyuni.git#:testsuite/features/profiles" : var.git_profiles_repo}
 server_http_proxy: ${var.server_http_proxy}
 
+sle11sp4_minion: ${var.sle11sp4_minion_configuration["hostname"]}
```
moio: I don't mind this very verbose style, but I can't help thinking if this would be better implemented via an array or map. Maybe this is something to refactor after Terraform 0.12 is supported.

salt/repos/default.sls
```diff
@@ -401,11 +401,15 @@ tools_update_repo:
     - file: /etc/apt/sources.list.d/Ubuntu1804-Client-Tools.list
 {% if 'head' in grains.get('product_version') | default('', true) %}
 {% set tools_repo_url = 'http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/Head:/Ubuntu18.04-SUSE-Manager-Tools/xUbuntu_18.04' %}
-{% elif '4.0' in grains.get('product_version') | default('', true) %}
+{% elif '4.0-master' in grains.get('product_version') | default('', true) %}
```
moio: Maybe my memory is failing again, but what is this product version?

At least it is not mentioned in https://github.com/uyuni-project/sumaform/blob/master/README_ADVANCED.md#changing-product-versions

lkotek: Very good catch, thanks :-) There should be nightly instead of master, I will correct it immediately.

modules/libvirt/controller/main.tf
lkotek: Yes, I agree. This was easy, but a bit repetitive way how to do it.
>  Maybe this is something to refactor after Terraform 0.12 is supported.

Yep, I am volunteering ;-)

lkotek:
> LGTM, but you advice in the "variables" files to look at the test suite example... but I don't see an example in your PR.

Examples added in separate file as we have a lot of new clients.

PR #588, opened 2019-09-04
Change default RAM size for clients

Bischoff:
It appears that it's impossible to use SUSE systems < 1 GiB with updates repository enabled because then zypper would generate an OOM condition.

This is therefore PR #576 reloaded :smile_cat: .

There is no change in the openstack module because `m1.smaller` is already at 1 GiB.


Bischoff:
@moio, do you agree?

PR #592, opened 2019-09-09
Forgot the openstack module when upgrading the controller

Bischoff:
As discussed by email.

My bad, I forgot the openstack module when upgrading the controler from 42.3 to 15.0.


Bischoff:
I will merge now.

Michele checked and the image seem to be in engineering cloud, however it's been a while we have not tested the openstack flavor, so I will try a deployment after the merge. I expect surprises.

PR #599, opened 2019-09-19
Forgot the GPG keys for the new repos

Bischoff:
## What does this PR change?

dummy packages are currently failing because of missing GPG key. This PR adds it.

it also fixes an error of `uyuni-stable` that should be `uyuni-released`.


moio:
Thanks for fixing this!

Bischoff:
> Thanks for fixing this!

I should have tested my rebase in the first place :disappointed: .

PR #600, opened 2019-09-20
Allow hosts to have more than one role

moio:
This is expected to be an internal-only change.

At the moment all it does is to open the door to overlay roles, which will be needed to implement Servers registering to other Servers.

@cbosdo I took the liberty of using `virthost` as the first example on how this refactoring could be useful.

I am pretty certain more cases will come up later.

### DIFF
--- a/modules/aws/mirror/main.tf
+++ b/modules/aws/mirror/main.tf
@@ -72,7 +72,7 @@ resource "null_resource" "mirror_salt_configuration" {
 hostname: ${replace("${aws_instance.instance.private_dns}", ".${var.region == "us-east-1" ? "ec2.internal" : "${var.region}.compute.internal"}", "")}
 domain: ${var.region == "us-east-1" ? "ec2.internal" : "${var.region}.compute.internal"}
 use_avahi: False
-role: mirror
+roles: [mirror]
 cc_username: ${var.cc_username}
 cc_password: ${var.cc_password}
 data_disk_device: xvdf
--- a/modules/libvirt/minionssh/main.tf
+++ b/modules/libvirt/minionssh/main.tf
@@ -14,10 +14,10 @@ module "minionssh" {
   ipv6 = "${var.ipv6}"
   connect_to_base_network = true
   connect_to_additional_network = true
+  roles = ["minionssh"]
   grains = <<EOF
 
 product_version: ${var.product_version}
-role: minionssh
 
 EOF
 
--- a/modules/libvirt/virthost/main.tf
+++ b/modules/libvirt/virthost/main.tf
@@ -15,11 +15,11 @@ module "virthost" {
   gpg_keys = "${var.gpg_keys}"
   ssh_key_path = "${var.ssh_key_path}"
   ipv6 = "${var.ipv6}"
+  roles = ["minion", "virthost"]
   additional_grains = <<EOF
 
 hvm_disk_image: "${var.hvm_disk_image}"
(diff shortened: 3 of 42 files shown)


modules/aws/mirror/main.tf
```diff
@@ -72,7 +72,7 @@ resource "null_resource" "mirror_salt_configuration" {
 hostname: ${replace("${aws_instance.instance.private_dns}", ".${var.region == "us-east-1" ? "ec2.internal" : "${var.region}.compute.internal"}", "")}
 domain: ${var.region == "us-east-1" ? "ec2.internal" : "${var.region}.compute.internal"}
 use_avahi: False
-role: mirror
+roles: [mirror]
```
Bischoff: Some of the main.tf files have:

roles: [mirror]

others have:

  roles = ["client"]

I understand one is a Salt grain, and the other one is a HCL variable, but it would be easy to do a mistake here and for me not to spot it. So I apologize if there's an error in those lines I miss.

modules/libvirt/minionssh/main.tf
```diff
@@ -14,10 +14,10 @@ module "minionssh" {
   ipv6 = "${var.ipv6}"
   connect_to_base_network = true
   connect_to_additional_network = true
+  roles = ["minionssh"]
```
Bischoff: speaking of which, we should have picked "sshminion" which is a more correct construct in English, and would be in line with "virthost" for example. But that's a different problem. Ping @juliogonzalez for awareness.

modules/libvirt/virthost/main.tf
```diff
@@ -15,11 +15,11 @@ module "virthost" {
   gpg_keys = "${var.gpg_keys}"
   ssh_key_path = "${var.ssh_key_path}"
   ipv6 = "${var.ipv6}"
+  roles = ["minion", "virthost"]
```
Bischoff: Yay!

salt/default/testsuite.sls
```diff
@@ -1,5 +1,5 @@
 {% if grains.get('testsuite') | default(false, true) %}
-{% if grains.get('role') in ['client', 'minion', 'minionssh'] %}
+{% if 'client' in grains.get('roles') or 'minion' in grains.get('roles') or 'minionssh' in grains.get('roles') %}
```
Bischoff: * What is the cost of `grains.get('roles')` ? If high, it would make sense to put it into some intermediate variable.
* What happens if there is no "roles"  grain, by some accident for example? (if it returns some empty list it should stil work)

modules/libvirt/minionssh/main.tf
juliogonzalez: Well, you reviewed my PR :-D

FMPOV it's not a big deal and this way you quickly identify it's a minion, but feel free to change it (if so, remember about the test-runner as well, and we'll need to recreate environments)

Bischoff: > Well, you reviewed my PR :-D

Yeah, I'm sorry I missed that. :crying_cat_face:

moio: We will have a number of breaking changes in 0.12 which will basically require rewriting of all `main.tf`s anyway. I plan to batch breaking changes so that we do them only once. This could be a candidate.

salt/default/testsuite.sls
moio: Cost is basically none.

At this point in time roles is expected to always be present, `[]` being a valid and default value. I tested `vanilla`.

PR #606, opened 2019-09-26
Mirror and testsuite assorted fixes

mcalmer:
## What does this PR change?

Mirror also images. They apear under the same path as the original.

### DIFF
--- a/modules/libvirt/base/main.tf
+++ b/modules/libvirt/base/main.tf
@@ -4,21 +4,21 @@ terraform {
 
 resource "libvirt_volume" "centos7_volume" {
   name = "${var.name_prefix}centos7"
-  source = "${var.mirror != "" ? "http://"var.mirror : "https://github.com"}/moio/sumaform-images/releases/download/4.3.0/centos7.qcow2"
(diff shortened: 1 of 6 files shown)


mcalmer:
@moio I am not sure if the syntax in base/main.tf is correct and works. I did not test it yet. Please have an extra look.

mcalmer:
Hmm, good question. I didn't thought about this problem as I installed a static mirror not setup via sumaform :-)
There is a long weekend comming

mcalmer:
> How is the mirror going to be deployed, since it depends on an image that lives on the mirror?

I tried, but I do not get it done.

moio:
Idea could be to mirror all images but opensuse. That will always be needed for the mirror

mcalmer:
@moio I changed to not use a mirror or openSUSE 15.1 . AFAIK this is the image used for minima.
I fixed also another script which was not working with the mirror. Are we now good to merge ?

moio:
I will add a README note and merge ASAP

mcalmer:
@moio thanks

PR #607, opened 2019-09-27
Disable download token checking by default

moio:
## What does this PR change?

It disables token checking by default on Servers, which is convenient to consume repos from other sumaform hosts.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -114,7 +114,7 @@ By default, sumaform deploys hosts with a range of tweaked settings for convenie
    * `create_sample_activation_key`: whether to create a sample activation key. Requires `create_first_user`
    * `create_sample_bootstrap_script`: whether to create a sample bootstrap script for traditional clients. Requires `create_sample_activation_key`
    * `publish_private_ssl_key`: copies the private SSL key in /pub for Proxies to copy automatically. Set to `false` for manual distribution
-   * `disable_download_tokens`: disable package token download checks. Set to `true` to enable checking
+   * `disable_download_tokens`: disable package token download checks. Set to `false` to enable checking
 
 
 ## Adding channels to SUSE Manager Servers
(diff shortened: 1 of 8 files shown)


README_ADVANCED.md
```diff
@@ -114,6 +114,7 @@ By default, sumaform deploys hosts with a range of tweaked settings for convenie
    * `create_sample_activation_key`: whether to create a sample activation key. Requires `create_first_user`
    * `create_sample_bootstrap_script`: whether to create a sample bootstrap script for traditional clients. Requires `create_sample_activation_key`
    * `publish_private_ssl_key`: copies the private SSL key in /pub for Proxies to copy automatically. Set to `false` for manual distribution
+   * `disable_download_tokens`: disable package token download checks. Set to `true` to enable checking
```
srbarrios:
```suggestion
   * `disable_download_tokens`: disable package token download checks. Set to `true` to disable checking
```

srbarrios:
Actually, if we will change all our main.tf to set as false this new parameter. Even in the examples.
It could be better to set as default to false?

README_ADVANCED.md
moio: :man_facepalming:

...embarassing!

moio:
Well, the defaults apply to developers/debuggers/users that want to demo SUSE Manager/Uyuni and are optimized towards convenience. In this case, default is to disable token checking as that's more convenient than enabling it (which is the SUSE Manager default).

In testsuites this is different: we want to test the software in the way it's configured by default, unchanged. That's why the testsuite examples have `false`.

srbarrios:
> Well, the defaults apply to developers/debuggers/users that want to demo SUSE Manager/Uyuni and are optimized towards convenience. In this case, default is to disable token checking as that's more convenient than enabling it (which is the SUSE Manager default).
> 
> In testsuites this is different: we want to test the software in the way it's configured by default, unchanged. That's why the testsuite examples have `false`.

Fair enough then :)

PR #608, opened 2019-09-29
enable debug for salt master and api

mcalmer:
## Warning

## What does this PR change?

Enable debug for salt master and api to debug connection problems when bootstrapping clients

### DIFF
--- a/salt/suse_manager_server/master-custom.conf
+++ b/salt/suse_manager_server/master-custom.conf
@@ -4,3 +4,4 @@ auto_accept: True
 ssh_minion_opts:
     log_file: ../../../../../var/log/salt-ssh.log
     log_level: debug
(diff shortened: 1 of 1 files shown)


mcalmer:
CC @Bischoff @srbarrios JFYI

Bischoff:
OK.

moio:
@mcalmer how about making this optional behind a flag?

https://github.com/uyuni-project/sumaform/blob/master/README_ADVANCED.md#turning-convenience-features-off

mcalmer:
For now it was meant only for debugging a special case. When we want to intoroduce debug tuning I would go for a bigger solution. We have a lot of them.
So this one can be rolled back when we know why the testsuite failed.

PR #615, opened 2019-10-03
Support for additional disk in server & proxy mounting in /var/spacewalk

srbarrios:
Related task: https://github.com/SUSE/spacewalk/issues/9486

> Note: I had issues testing from master, so my only choice was to move to this temporal branch, where I succeded with the contribution. 

Adding support for an additional disk mounting in /var/spacewalk
On susemanager_server and susemanager_proxy on libvirt provider.

Example main.tf
```
provider "libvirt" {
  uri = "qemu:///system"
}

module "base" {
  source = "sumaform/modules/libvirt/base"

  cc_username = "xxxxx"
  cc_password = "xxxxx"

  pool = "obarrios_disks"
  additional_network = "192.168.6.0/24"
  name_prefix = "oba-"
  images = ["sles15sp1"]
}

module "server" {
  source = "sumaform/modules/libvirt/suse_manager"
  product_version = "4.0-nightly"
  name = "server"  
  repository_disk_size = 109951162777

  auto_accept = false
  disable_firewall = false
  allow_postgres_connections = false
  skip_changelog_import = false
  browser_side_less = false
  create_first_user = false
  mgr_sync_autologin = false
  create_sample_channel = false
  create_sample_activation_key = false
  create_sample_bootstrap_script = false
  publish_private_ssl_key = false

  ssh_key_path = "./salt/controller/id_rsa.pub"

}
```

### DIFF
--- a/modules/libvirt/suse_manager_proxy/main.tf
+++ b/modules/libvirt/suse_manager_proxy/main.tf
@@ -10,6 +10,13 @@ variable "images" {
   }
 }
 
+resource "libvirt_volume" "server_data_disk" {
+  name = "${var.base_configuration["name_prefix"]}${var.name}-server-data-disk"
+  size = "${var.repository_disk_size}"
+  pool = "${var.data_pool}"
+  count = "${var.repository_disk_size > 0 ? 1 : 0}"
+}
+
 module "suse_manager_proxy" {
   source = "../host"
 
@@ -42,6 +49,7 @@ server_password: ${var.server_configuration["password"]}
 generate_bootstrap_script: ${var.generate_bootstrap_script}
 publish_private_ssl_key: ${var.publish_private_ssl_key}
 apparmor: ${var.apparmor}
+repository_disk_size: ${var.repository_disk_size}
 
 EOF
 
@@ -51,6 +59,15 @@ EOF
   vcpu = "${var.vcpu}"
   running = "${var.running}"
   mac = "${var.mac}"
+
+  // HACK: Terraform 0.11 ternary operator is not short-circuiting
+  // https://github.com/hashicorp/terraform/issues/11566
+  // example by https://github.com/coreos/tectonic-installer/commit/0f209b31b169ce129ba457096b23792f6601d66e#diff-17708827fa84c637698cd840c01710c9R16
+  additional_disk = "${slice(
+    list(map("volume_id", join("",libvirt_volume.server_data_disk.*.id))),
+    0,
+    var.repository_disk_size > 0 ? 1 : 0
+  )}"
 }
 
 output "configuration" {
--- a/salt/suse_manager_proxy/additional_disk.sls
+++ b/salt/suse_manager_proxy/additional_disk.sls
@@ -1,0 +1,31 @@
+include:
+  - default
+
+{% if grains.get('repository_disk_size') > 0 %}
+
+parted:
+  pkg.installed
+
+spacewalk_partition:
+  cmd.run:
+    - name: /usr/sbin/parted -s /dev/vdb mklabel gpt && /usr/sbin/parted -s /dev/vdb mkpart primary 0% 100% && sleep 1 && /sbin/mkfs.ext4 /dev/vdb1
+    - unless: ls /dev/vdb1
+    - require:
+      - pkg: parted
+
+spacewalk_directory:
+  file.directory:
+    - name: /var/spacewalk
+    - makedirs: True
+  mount.mounted:
+    - name: /var/spacewalk
+    - device: /dev/vdb1
+    - fstype: ext4
+    - mkmnt: True
+    - persist: True
+    - opts:
+      - defaults
+    - require:
+      - cmd: spacewalk_partition
+
+{% endif %}
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -797,3 +797,19 @@ module "minion" {
 ```
 
 To disable the swap file, set its size to 0.
+
+# Additional disk on SUSE Manager server or proxy
+
+In case the default disk size for those machines is not enough for the amount of products you want to synchronize. You can add an additional disk which will mount the first volume in /var/spacewalk with size `repository_disk_size`. This additional disk will be created in the pool specified by `data_pool`
+
+A libvirt example is:
+
+```hcl
+module "server" {
+  source = "sumaform/modules/libvirt/suse_manager"
+  base_configuration = "${module.base.configuration}"
+  product_version = "4.0-nightly"
+  name = "server"
+  repository_disk_size = 536870912000
+  data_pool = "default"
+```
--- a/modules/libvirt/suse_manager/main.tf
+++ b/modules/libvirt/suse_manager/main.tf
@@ -11,6 +11,13 @@ variable "images" {
   }
 }
 
+resource "libvirt_volume" "server_data_disk" {
+  name = "${var.base_configuration["name_prefix"]}${var.name}-server-data-disk"
+  size = "${var.repository_disk_size}"
+  pool = "${var.data_pool}"
+  count = "${var.repository_disk_size > 0 ? 1 : 0}"
+}
+
 module "suse_manager" {
   source = "../host"
 
@@ -72,6 +79,7 @@ apparmor: ${var.apparmor}
 from_email: ${var.from_email}
 traceback_email: ${var.traceback_email}
 saltapi_tcpdump: ${var.saltapi_tcpdump}
+repository_disk_size: ${var.repository_disk_size}
 
 EOF
 
@@ -81,7 +89,15 @@ EOF
   vcpu = "${var.vcpu}"
   running = "${var.running}"
   mac = "${var.mac}"
-  additional_disk = "${var.additional_disk}"
+
+  // HACK: Terraform 0.11 ternary operator is not short-circuiting
+  // https://github.com/hashicorp/terraform/issues/11566
+  // example by https://github.com/coreos/tectonic-installer/commit/0f209b31b169ce129ba457096b23792f6601d66e#diff-17708827fa84c637698cd840c01710c9R16
+  additional_disk = "${slice(
+    list(map("volume_id", join("",libvirt_volume.server_data_disk.*.id))),
+    0,
+    var.repository_disk_size > 0 ? 1 : 0
+  )}"
 }
 
 output "configuration" {
--- a/modules/libvirt/suse_manager/variables.tf
+++ b/modules/libvirt/suse_manager/variables.tf
@@ -261,9 +261,14 @@ variable "mac" {
   default = ""
 }
 
-variable "additional_disk" {
-  description = "disk block definition(s) to be added to this host"
-  default = []
+variable "repository_disk_size" {
+  description = "Size of an aditional disk for /var/spacewalk partition"
+  default = 0
+}
+
+variable "data_pool" {
+  description = "libvirt storage pool name for this host's data disk"
+  default = "default"
 }
 
 variable "cpu_model" {
--- a/modules/libvirt/suse_manager_proxy/variables.tf
+++ b/modules/libvirt/suse_manager_proxy/variables.tf
@@ -135,6 +135,16 @@ variable "mac" {
   default = ""
 }
 
+variable "repository_disk_size" {
+  description = "Size of an aditional disk for /var/spacewalk partition"
+  default = 0
+}
+
+variable "data_pool" {
+  description = "libvirt storage pool name for this host's data disk"
+  default = "default"
+}
+
 variable "cpu_model" {
   description = "Define what CPU model the guest is getting (host-model, host-passthrough or the default)."
   default = ""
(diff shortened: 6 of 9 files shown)


srbarrios:
Tested in local with an additional disk of 10GB.
This is the output:

```
oba-server:~ # parted
GNU Parted 3.2
Using /dev/vdb
Welcome to GNU Parted! Type 'help' to view a list of commands.
(parted) print devices                                                    
/dev/vdb (10.0GB)
/dev/vda (215GB)
(parted) print all                                                        
Model: Virtio Block Device (virtblk)
Disk /dev/vdb: 10.0GB
Sector size (logical/physical): 512B/512B
Partition Table: gpt
Disk Flags: 

Number  Start   End     Size    File system  Name     Flags
 1      2048MB  9999MB  7951MB  ext4         primary


Model: Virtio Block Device (virtblk)
Disk /dev/vda: 215GB
Sector size (logical/physical): 512B/512B
Partition Table: msdos
Disk Flags: 

Number  Start   End    Size   Type     File system  Flags
 1      1049kB  215GB  215GB  primary  ext4         boot, type=83


(parted) quit                                                             
oba-server:~ # mount
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
devtmpfs on /dev type devtmpfs (rw,nosuid,size=2000244k,nr_inodes=500061,mode=755)
securityfs on /sys/kernel/security type securityfs (rw,nosuid,nodev,noexec,relatime)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev)
devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000)
tmpfs on /run type tmpfs (rw,nosuid,nodev,mode=755)
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,mode=755)
cgroup on /sys/fs/cgroup/unified type cgroup2 (rw,nosuid,nodev,noexec,relatime)
cgroup on /sys/fs/cgroup/systemd type cgroup (rw,nosuid,nodev,noexec,relatime,xattr,name=systemd)
pstore on /sys/fs/pstore type pstore (rw,nosuid,nodev,noexec,relatime)
cgroup on /sys/fs/cgroup/cpu,cpuacct type cgroup (rw,nosuid,nodev,noexec,relatime,cpu,cpuacct)
cgroup on /sys/fs/cgroup/net_cls,net_prio type cgroup (rw,nosuid,nodev,noexec,relatime,net_cls,net_prio)
cgroup on /sys/fs/cgroup/cpuset type cgroup (rw,nosuid,nodev,noexec,relatime,cpuset)
cgroup on /sys/fs/cgroup/devices type cgroup (rw,nosuid,nodev,noexec,relatime,devices)
cgroup on /sys/fs/cgroup/hugetlb type cgroup (rw,nosuid,nodev,noexec,relatime,hugetlb)
cgroup on /sys/fs/cgroup/rdma type cgroup (rw,nosuid,nodev,noexec,relatime,rdma)
cgroup on /sys/fs/cgroup/perf_event type cgroup (rw,nosuid,nodev,noexec,relatime,perf_event)
cgroup on /sys/fs/cgroup/pids type cgroup (rw,nosuid,nodev,noexec,relatime,pids)
cgroup on /sys/fs/cgroup/blkio type cgroup (rw,nosuid,nodev,noexec,relatime,blkio)
cgroup on /sys/fs/cgroup/memory type cgroup (rw,nosuid,nodev,noexec,relatime,memory)
cgroup on /sys/fs/cgroup/freezer type cgroup (rw,nosuid,nodev,noexec,relatime,freezer)
/dev/vda1 on / type ext4 (rw,relatime,data=ordered)
rpc_pipefs on /var/lib/nfs/rpc_pipefs type rpc_pipefs (rw,relatime)
systemd-1 on /proc/sys/fs/binfmt_misc type autofs (rw,relatime,fd=28,pgrp=1,timeout=0,minproto=5,maxproto=5,direct,pipe_ino=14564)
hugetlbfs on /dev/hugepages type hugetlbfs (rw,relatime)
debugfs on /sys/kernel/debug type debugfs (rw,relatime)
mqueue on /dev/mqueue type mqueue (rw,relatime)
tmpfs on /run/user/0 type tmpfs (rw,nosuid,nodev,relatime,size=403084k,mode=700)
tracefs on /sys/kernel/debug/tracing type tracefs (rw,relatime)
/dev/vdb1 on /var/spacewalk type ext4 (rw,relatime,data=ordered)
oba-server:~ # df -h
Filesystem      Size  Used Avail Use% Mounted on
devtmpfs        2.0G  8.0K  2.0G   1% /dev
tmpfs           2.0G   52K  2.0G   1% /dev/shm
tmpfs           2.0G  8.7M  2.0G   1% /run
tmpfs           2.0G     0  2.0G   0% /sys/fs/cgroup
/dev/vda1       196G  4.7G  182G   3% /
tmpfs           394M     0  394M   0% /run/user/0
/dev/vdb1       7.3G   34M  6.9G   1% /var/spacewalk
```

moio:
Will be looking at this ASAP, sorry for the delay in review

srbarrios:
> For the other "additional" resources (eg. `additional_network`) we have only 1 variable.
> 
> Why is the `enable_` variable needed? The list of additional disks should be enough, i.e. if empty or absent, then it's disabled.

Totally agreed!
That was my first approach but I couldn't make it work, I really spend time trying to get it. It will be awesome if you or Silvio can help to avoid that extra variable.

My issue is that if I want to use additional_disk inside my salt file to do `{% if grains.get('additional_disk') %}`, I first need to add this variable as a grain in the main.tf of the server/proxy, and when I do that I have a problem of types that I was not able to solve.

I was adding it like that `additional_disk: ${var.additional_disk}`

Bischoff:
OK, here is how I would try do it if I were you: instead of introducing new variables, I would simply use the already existing variable `additional_disk`.

I would use it from the final `main.tf` and/or from the salt states.

It could be you need a new variable after all to describe the partitioning (not the disk).

You can see information about existing `additional_disk` variable in `modules/libvirt/ host/main.tf` line 31 and following.

Hope that helps.

srbarrios:
> I can see two ways out of it:
> 
> 1. in the states on the server, you test if a grain `additional_disk` exists, and if it does, you do the partitioning, etc. This grain is computed by testing if `additional_disk` is empty or not. Or you can put the number of disks in the list in the grain with `.count`, instead of a boolean. Or you can marshall the list into a string with `join()` .
> 
> or
> 
> 1. you indeed use a new variable, but don't call it `enable`. Creating the disk enables it already smile . Maybe create a variable that describes more accurately what you want to do, like `partition_additional_disk`.

Thanks for the help Eric. Looks now work and no need of a second variable in variables.tf, but only a variable as grain, but nothing need from user side :)

srbarrios:
@moio could you also take a look? Thanks

modules/libvirt/suse_manager_proxy/main.tf
```diff
@@ -42,6 +42,7 @@ server_password: ${var.server_configuration["password"]}
 generate_bootstrap_script: ${var.generate_bootstrap_script}
 publish_private_ssl_key: ${var.publish_private_ssl_key}
 apparmor: ${var.apparmor}
+additional_disk_count: ${length(var.additional_disk)}
```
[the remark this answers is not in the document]
Bischoff: As discussed, it can also stay there...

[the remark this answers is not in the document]
srbarrios: Discussed. Closing.

salt/suse_manager_proxy/additional_disk.sls
```diff
@@ -0,0 +1,31 @@
+include:
+  - default
+
+{% if grains.get('additional_disk_count') > 0 %}
```
moio: Is this feature about adding one disk for `/var/spacewalk` only or is it a general feature to add arbitrary disks?

I would separate those concerns. I think the objective here is the former, as it's actually the most frequent use case.

So I'd add a variable in the `suse_manager` Terraform module to optionally set the size (eg. `"repository_disk_size"`, 0 by default). At that point, if the size is set, create such a disk inside the `suse_manager` module and pass the device name of the disk (`vdb`) via grains to Salt.

Look at the `mirror` module, it does the same.

This will make it trivial to port to backends other than libvirt.

srbarrios: Yep, maybe I lost perspective here. I started looking at mirror in fact :)
But was my first time playing with salt and grains, now I think I understand how it works.
I will do changes as suggested. Thanks.

srbarrios: Ok @moio we have a problem, it differs from the mirror that we need to be able to create or not the additional disk depending on the variable, so in the mirror we do:
```
resource "libvirt_volume" "data_disk" {
  name = "${var.base_configuration["name_prefix"]}mirror-data-disk"
  size = 1099511627776 # 1 TiB
  pool = "${var.data_pool}"
  lifecycle {
    prevent_destroy = true
  }
}
```
On the main.tf, because we always create the disk. But in our case, how we can do, to create only if the repository_disk_size is bigger than 0?

srbarrios: I will use count for that purpose:
```
resource "libvirt_volume" "server_data_disk" {
  name = "${var.base_configuration["name_prefix"]}-server-data-disk"
  size = "${var.repository_disk_size}"
  pool = "${var.data_pool}"
  count = "${var.repository_disk_size > 0 ? 1 : 0}"
}
```

Bischoff: nice idea

srbarrios: @moio Ok Silvio, I spend a lot of effort and days to make it work, my pleasure to learn hcl, but arriving at that point it doesn't make sense that if you can do it faster, as you have good knowledge of Terraform language. Really I have more interesting tasks to do, and that's one of those tasks we discussed were other squads can really help us.
Current problem is that I want to set an additional_disk variable to an empty array if we don't have a volume created, otherwise return the array with the volume_id. No way to accomplish it.

Please, could you finish this PR?

moio:
I cannot finish the PR immediately but I did rebase it on top of current `master`, and added fixes for the particular HCL problem you were stuck on and another small detail. That should definitely unblock you on the HCL side.

Congrats, you stumbled upon a pretty peculiar HCL limitation that requires a circus-level spectacular hack to work around. It's only because I faced similar issues in the past that I could find the right reference to someone who worked around it. I am very glad this was solved in a nice-to-look-at way in Terraform 0.12, and moving forward with that migration to avoid us similar headaches in the future. I can understand the frustration, this is one of the very irritating things about 0.11.

In another piece of good news for this PR, I'll be dropping OpenStack support soon due to a shift in priorities. That means you can focus on the libvirt part and not care at all about OpenStack to get this merged.

Hope this is enough to at least relieve you a bit.

Thanks for the efforts so far

srbarrios:
> I cannot finish the PR immediately but I did rebase it on top of current `master`, and added fixes for the particular HCL problem you were stuck on and another small detail. That should definitely unblock you on the HCL side.
> 
> Congrats, you stumbled upon a pretty peculiar HCL limitation that requires a circus-level spectacular hack to work around. It's only because I faced similar issues in the past that I could find the right reference to someone who worked around it. I am very glad this was solved in a nice-to-look-at way in Terraform 0.12, and moving forward with that migration to avoid us similar headaches in the future. I can understand the frustration, this is one of the very irritating things about 0.11.
> 
> In another piece of good news for this PR, I'll be dropping OpenStack support soon due to a shift in priorities. That means you can focus on the libvirt part and not care at all about OpenStack to get this merged.
> 
> Hope this is enough to at least relieve you a bit.
> 
> Thanks for the efforts so far

Thanks Silvio, I'm going to continue from that point, see if we can finish it for that milestone :)

srbarrios:
Looks good now, it works.

srbarrios:
Let's merge it so I can continue our qam development.
Please @moio if you see something tomorrow, let me know and I can add a new commit for it. Thanks.

salt/suse_manager_proxy/additional_disk.sls
```diff
@@ -0,0 +1,31 @@
...
+  - default
+
+{% if grains.get('repository_disk_size') > 0 %}
+
+parted:
+  pkg.installed
+
+spacewalk_partition:
+  cmd.run:
+    - name: /usr/sbin/parted -s /dev/vdb mklabel gpt && /usr/sbin/parted -s /dev/vdb mkpart primary 0% 100% && sleep 1 && /sbin/mkfs.ext4 /dev/vdb1
```
moio: The device name should be passed via grains, as you can't assume it will be `/dev/vdb` for other backends. Please correct this in a follow up PR and ping me before merging, thanks.

srbarrios: https://github.com/uyuni-project/sumaform/pull/639

PR #616, opened 2019-10-07
DRY for SLES and Galaxy paths

a-kpappas:
## What does this PR change?

Rework the repository saltstates so that they are generated by jinja instead of being hard coded.

This applies to the base repos and Galaxy repos. I did not touch the SLE11 repos since I have not used sumaform to generate such machines and their paths are dissimilar enough to not fit in the same template.

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -93,200 +93,71 @@ tools_additional_repo:
 {% endif %} {# grains['osrelease'] == '11.4' #}
 
 
-{% if '12' in grains['osrelease'] %}
-{% if grains['osrelease'] == '12' %}
+{% if grains['osrelease_info'][0] >= 12 %}
+
+{% set version = grains['osrelease']|replace('.','-SP') %}
+{% set major = grains['osrelease_info'][0]|string %}
+{% if grains['osrelease_info'][0] == 12  %}
+{% set base_name = "SLE-SERVER" %}
+{% elif grains['osrelease_info'][0] == 15%}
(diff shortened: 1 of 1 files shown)


moio:
@a-kpappas one alternative I was contemplating for this particular file was to use Python directly instead of YAML + Jinja.

For you to consider:

https://docs.saltstack.com/en/latest/ref/renderers/all/salt.renderers.pyobjects.html
https://docs.saltstack.com/en/latest/ref/renderers/all/salt.renderers.py.html

a-kpappas:
But anyway I am not really sure I can rework it to the pyobjects renderer, since I am not familiar with it at all and don't want to introduce bugs. However this simplification is useful for now and for when this is going to be refactored to py/pyobjects.

moio:
Then just ping me when this is ready to review and the WIP tag is removed and I'll take a look.

I do not plan to move away from Jinja any time soon in general, but a few very complicated states might see that happening. repos/default.sls is the prime candidate

a-kpappas:
@moio Could you have a look?

moio:
If the answer is yes this can be merged.

a-kpappas:
I tested with the main.tf.libvirt.example ONLY on 12. I will test with both. Is there a better configuration 
to test with?

moio:
One could probably invent several different combinations, we do not have examples for all. That's the whole point of sumaform after all: flexibility.

I am fine when you tested it with a few combinations covering at least the main cases. Thanks!

moio:
Whoops sorry I forgot to merge this. Apologies.

PR #617, opened 2019-10-09
Add missing "pkgrepo.managed" to "tools_additional_repo"

meaksh:
## What does this PR change?

This PR add a missing `pkgrepo.managed` reference when rendering the `tools_additional_repo` state in order to avoid a failure that is caused because a malformed state.


moio:
Whoops, thanks!

PR #618, opened 2019-10-09
Temporarily don't use mirrors for fake packages

Bischoff:
## What does this PR change?

Temporary workaround, `testsuite-stable-with-new-fake-repos` branch only.

Don't use arbitrary mirrors for fake packages. Use the one in the Netherlands.

Reason: we had to keep same version and release numbers when redoing the fake packages, and this puzzles some mirrors, and others not.

Long term solution is to fix problem at mirrors.


srbarrios:
OK.

PR #619, opened 2019-10-09
[WIP] configure loglevel for salt and python

mcalmer:
## Warning

## What does this PR change?

Configure logging and loglevel via main.tf .

As of now for salt and python backend.

### DIFF
--- a/modules/libvirt/suse_manager/main.tf
+++ b/modules/libvirt/suse_manager/main.tf
@@ -52,6 +52,8 @@ disable_firewall: ${var.disable_firewall}
 allow_postgres_connections: ${var.allow_postgres_connections}
 unsafe_postgres: ${var.unsafe_postgres}
 java_debugging: ${var.java_debugging}
+salt_logging: ${var.salt_logging}
+python_logging: ${var.python_logging}
 skip_changelog_import: ${var.skip_changelog_import}
 browser_side_less: ${var.browser_side_less}
 create_first_user: ${var.create_first_user}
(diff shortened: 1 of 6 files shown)


mcalmer:
Still need a good idea how to configure log4j.properties for tomcat and taskomatic

moio:
I think best would be to allow for a new parameter which is a map from log level to class or package names. It can be passed as-is via grains, then salt should pick it up and calculate lines to add to `log4j.conf`.

Example:

```hcl
java_logging = {
 "DEBUG" : ["com.suse.manager.tasks, com.suse.manager.reactor.PGEventListener"],
 "INFO" : ["com.redhat.rhn.frontend.action.LoginAction"],
}
```

Should result in:

```properties
log4j.logger.com.redhat.rhn.frontend.action.LoginAction=INFO
log4j.logger.com.suse.manager.tasks=DEBUG
log4j.logger.com.suse.manager.reactor.PGEventListener=DEBUG 
```

(`log4j.logger.` is prepended and `=LEVEL` is appended)

Note that the `log4.conf` file should be extracted from `rhn.jar` (by default it does not exist) and then new lines should be added. The result should be copied in `/srv/tomcat/webapps/rhn/WEB-INF/classes/log4j.properties` (for Tomcat), `/usr/share/rhn/classes/log4j.properties` (for Taskomatic) and `usr/share/rhn/search/classes/log4j.properties` (for Search). Parent directories might be missing.


Any plans to finish this PR? :-)

PR #622, opened 2019-10-10
Install the GPG key of download.opensuse.org

Bischoff:
This PR installs the GPG key of download.opensuse.org when we run the test suite.

If we are running the testsuite, we need the fake packages,
and those are now on download.opensuse.org.

Note: This key is also installed during completly different circumstances (on RedHat client, when we are running uyuni)

### DIFF
--- a/salt/mirror/minima.yaml
+++ b/salt/mirror/minima.yaml
@@ -217,13 +217,13 @@ http:
     archs: [x86_64]
 
   # Testsuite dummy packages for testing repo
-  - url: http://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Pool/standard_rpm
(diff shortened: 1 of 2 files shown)


Bischoff:
> LGTM. Not sure if this watch needs some timeout, but looks ok.

To be very honest, I'm not sure why we need "wait" at all here. It's copy and paste from Silvio.

moio:
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.cmd.html#should-i-use-cmd-run-or-cmd-wait

Bischoff:
> https://docs.saltstack.com/en/latest/ref/states/all/salt.states.cmd.html#should-i-use-cmd-run-or-cmd-wait

Yes, I know this piece of documentation, but it's clear as mud...

moio:
What is not clear about that particular paragraph?

PR #624, opened 2019-10-11
Remove hack for cucumber-html-reporter

srbarrios:
## What does this PR change?

Removing a hack about cucumber-html-reporter to add a hotfix which is already merged in 5.0.2
https://www.npmjs.com/package/cucumber-html-reporter


moio:
Thanks @srbarrios

PR #627, opened 2019-10-17
Update AWS templates

jgleissner:
This is an attempt to bring the AWS templates into a state where they can be used to spin up a cluster using the official released SUSE Manger 4.0 Server image in which the spacewalk test suite can be executed.

It adds a controller resource and makes some adjustments to the other ones, and also makes some changes to the salt templates that were required to make the test suite pass. Those changes should be non intrusive.

It probably needs a bit more polishing but I thought this should work as an initial base for discussion.

### DIFF
--- a/main.tf.aws-create-mirror-snapshot.example
+++ b/main.tf.aws-create-mirror-snapshot.example
@@ -1,0 +1,82 @@
+// This template can be used to create a mirror hosts data snapshot.
+// To do so, copy this template to main.tf, fill in your SCC and AWS
+// credentials and the regions and AMI information.
+
+// Create a mirror instance using 'terraform apply', and if successful,
+// run 'modules/aws/create_aws_mirror_snapshot' to sync
+// channels and create a snapshot from the resulting data volume.
+
+// If you want to base the snapshot on an existing snapshot, either set
+// data_volume_snapshot_id to the ID of the snapshot to use, or set it
+// to "${data.aws_ebs_snapshot.data_disk_snapshot.id}" to look up the
+// latest snapshot that uses the same $name_prefix.
+
+locals {
+  cc_username       = "" // SCC or NCC username
+  cc_password       = "" // SCC password
+  key_name          = "" // SSH key name
+  key_file          = "" // SSH key file
+  name_prefix       = "mirrorsnap"  // change to a prefix if you do not have an exclusive VPC
+  region            = "eu-west-2"
+  availability_zone = "${local.region}a"
+}
+
+provider "aws" {
+  region     = "${local.region}"
+  access_key = "" // add access key as a string here
+  secret_key = "" // add secret key as a string here
+}
+
+module "aws_network" {
+  source            = "./modules/aws/network"
+  region            = "${local.region}"
+  availability_zone = "${local.availability_zone}"
+
+  ssh_allowed_ips = [
+    "1.2.3.4",
+  ]
+
+  name_prefix = "${local.name_prefix}"
+}
+
+data "aws_ebs_snapshot" "data_disk_snapshot" {
+  most_recent = true
+
+  filter {
+    name   = "tag:Name"
+    values = ["${local.name_prefix}-mirror-data-volume-snapshot"]
+  }
+}
+
+module "aws_mirror" {
+  source                   = "./modules/aws/mirror"
+  region                   = "${local.region}"
+  availability_zone        = "${local.availability_zone}"
+  ami                      = "ami-00aa3f926488064bf"
+  key_name                 = "${local.key_name}"
+  key_file                 = "${local.key_file}"
+  public_subnet_id         = "${module.aws_network.public_subnet_id}"
+  public_security_group_id = "${module.aws_network.public_security_group_id}"
+  cc_username              = "${local.cc_username}"
+  cc_password              = "${local.cc_password}"
+  name_prefix              = "${local.name_prefix}"
+  minima_config            = "minima-small.yaml"
+  data_volume_snapshot_id  = "" // use existing snapshot as starting point
+                                // leave empty to create from scratch
+}
+
+output "mirror_host_name" {
+  value = "${module.aws_mirror.public_name}"
+}
+
+output "ebs_volume_id" {
+  value = "${module.aws_mirror.data_volume_id}"
+}
+
+output "region" {
+  value = "${local.region}"
+}
+
+output "snapshot_name" {
+  value = "${local.name_prefix}-mirror-data-volume-snapshot"
+}
--- a/main.tf.aws-testsuite.example
+++ b/main.tf.aws-testsuite.example
@@ -212,11 +212,23 @@ resource "null_resource" "update_server_conf" {
   }
 }
 
-output "mirror host public name" {
+output "mirror_host_name" {
   value = "${module.aws_mirror.public_name}"
 }
 
-output "controller host public name" {
+output "ebs_volume_id" {
+  value = "${module.aws_mirror.data_volume_id}"
+}
+
+output "region" {
+  value = "${local.region}"
+}
+
+output "snapshot_name" {
+  value = "${local.name_prefix}-mirror-data-volume-snapshot"
+}
+
+output "controller_host_public_name" {
   value = "${module.testcontroller.public_name}"
 }
 
--- a/main.tf.aws-testsuite.example
+++ b/main.tf.aws-testsuite.example
@@ -1,0 +1,222 @@
+locals {
+  cc_username = "CC_USER" // SCC or NCC username
+  cc_password = "CC_PASS"   //SCC or NCC password
+  key_name = "AWS_KEY"      // SSH key name
+  key_file = "/path/to/priave_ssh_key"      // SSH key file
+  name_prefix = "sumatest"   // change to a prefix if you do not have an exclusive VPC
+  prod_version = "4.0"
+  region = "eu-central-1"
+  availability_zone = "${local.region}a"
+}
+
+provider "aws" {
+  region = "${local.region}"
+  access_key = "MY_AWS_ACCESS_KEY" // add access key as a string here
+  secret_key = "MY_AWS_SECRET_KEY" // add secret key as a string here
+}
+
+
+module "aws_network" {
+  source = "./modules/aws/network"
+  region = "${local.region}"
+  availability_zone = "${local.availability_zone}"
+  ssh_allowed_ips = [
+    "1.2.3.4"
+  ]
+  name_prefix = "${local.name_prefix}"
+}
+
+data "aws_ebs_snapshot" "data_disk_snapshot" {
+  most_recent = true
+
+  filter {
+    name   = "tag:Name"
+    values = ["${local.name_prefix}-mirror-data-volume-snapshot"]
+  }
+}
+
+module "aws_mirror" {
+  source = "./modules/aws/mirror"
+  region = "${local.region}"
+  availability_zone = "${local.availability_zone}"
+  ami = "" // AMI ID of image to use as mirror. openSUSE Leap works fine. Image needs to have salt-minion installed.
+  key_name = "${local.key_name}"
+  key_file = "${local.key_file}"
+  data_volume_snapshot_id = "${data.aws_ebs_snapshot.data_disk_snapshot.id}" // see top comment in modules/aws/mirror/main.tf
+  public_subnet_id = "${module.aws_network.public_subnet_id}"
+  public_security_group_id = "${module.aws_network.public_security_group_id}"
+  cc_username = "${local.cc_username}"
+  cc_password = "${local.cc_password}"
+  name_prefix = "${local.name_prefix}"
+  additional_repos {
+    tools_repo = "http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools/openSUSE_Leap_15.0/"
+  }
+}
+
+module "aws_server" {
+  dependencies = [
+    "${module.aws_mirror.depended_on}"
+  ]
+  source = "./modules/aws/host"
+  roles = ["suse_manager_server"]
+  name = "server"
+  region = "${local.region}"
+  availability_zone = "${local.availability_zone}"
+  ami = "" // look up ID with 'pint amazon images --active --region my-region --filter name~manager-4-0-server'
+  instance_type = "t2.large"
+  volume_size = 100 // GiB
+  key_name = "${local.key_name}"
+  key_file = "${local.key_file}"
+  monitored = false
+  private_subnet_id = "${module.aws_network.private_subnet_id}"
+  private_security_group_id = "${module.aws_network.private_security_group_id}"
+  name_prefix = "${local.name_prefix}"
+  authorized_keys = ["ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDOSLK0kJAulwO9Qwm/DVKLsGzv8Fh+ZcGe5x5d+VzO4/+xKnxjOPvGVngm9Hki8N436bCCvIRqwzstDUyJnTHhEe3qzsl02QGqIIfVjzK3BEp4sA8XcmByBy3pxqbTjAjrW8a+n43DBRMx4J86LgywX3fFt8ceEym3rFer/IzQD9QvKRocycNCwhpmGKEDh4zdl9PdyU5RcUpiA01AD/8rdJMNYnUxIsaNElcgP2LSzROjNdyu11cq9h3/cb3b7oRRef718GOkIkzqL/szo4Mm/j4KfyG0SBWyPxlJktFADD3lcQIRrGakFhiZOFbg2HZci26q+l/hUfLMbzcBrWlc+gzY4z4LMF8YBy1x/x+yR4j1SzZuqempSRu7rDKevcIGBF/fALPyiXH2EzYKsVcN9nv5mFCjmwVK7WTjw1SFzufKwdmnvOYgyeYuMaOJv3LEBY1jJg8r4xqzuQx32KNK/iJz1E+njptV3ZhOYDiHBmieykP08nWxdS79N1DbljxRzS54T7CArfGo23nYTNlMewPeKFx7ZqtUVv03g8i1Hn8WdArz2TYciWWrPErBEALBpXsu7Oq5AT97wAhcAIf3B3dV5p/bXgnhALA/SodNBf/wr0U2f5VzTDyJ9eNZURt16jSxDgXkuKKY/rimLqqt2FC1SeicmTql1nVyHG3j9Q== root@controller"]
+
+  mirror_public_name = "${module.aws_mirror.public_name}"
+  mirror_private_name = "${module.aws_mirror.private_name}"
+  product_version = "${local.prod_version}"
+  cc_username = "${local.cc_username}"
+  cc_password = "${local.cc_password}"
+  testsuite = true
+  no_install = true
+}
+
+module "aws_minion" {
+  dependencies = [
+    "${module.aws_mirror.depended_on}"
+  ]
+  source = "./modules/aws/host"
+  roles = ["minion"]
+  name = "sle-minion"
+  count = 1
+  region = "${local.region}"
+  availability_zone = "${local.availability_zone}"
+  ami = "ami-014799f67e8f28c77" // SLES12 SP4 image (SLES12 minion currently required for testsuite)
+                                // look up with 'pint amazon images --active --region my-region --filter name~sles-12-sp4-byos'
+  instance_type = "t2.nano"
+  volume_size = 10 // GiB
+  key_name = "${local.key_name}"
+  key_file = "${local.key_file}"
+  private_subnet_id = "${module.aws_network.private_subnet_id}"
+  private_security_group_id = "${module.aws_network.private_security_group_id}"
+  name_prefix = "${local.name_prefix}"
+  authorized_keys = ["ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDOSLK0kJAulwO9Qwm/DVKLsGzv8Fh+ZcGe5x5d+VzO4/+xKnxjOPvGVngm9Hki8N436bCCvIRqwzstDUyJnTHhEe3qzsl02QGqIIfVjzK3BEp4sA8XcmByBy3pxqbTjAjrW8a+n43DBRMx4J86LgywX3fFt8ceEym3rFer/IzQD9QvKRocycNCwhpmGKEDh4zdl9PdyU5RcUpiA01AD/8rdJMNYnUxIsaNElcgP2LSzROjNdyu11cq9h3/cb3b7oRRef718GOkIkzqL/szo4Mm/j4KfyG0SBWyPxlJktFADD3lcQIRrGakFhiZOFbg2HZci26q+l/hUfLMbzcBrWlc+gzY4z4LMF8YBy1x/x+yR4j1SzZuqempSRu7rDKevcIGBF/fALPyiXH2EzYKsVcN9nv5mFCjmwVK7WTjw1SFzufKwdmnvOYgyeYuMaOJv3LEBY1jJg8r4xqzuQx32KNK/iJz1E+njptV3ZhOYDiHBmieykP08nWxdS79N1DbljxRzS54T7CArfGo23nYTNlMewPeKFx7ZqtUVv03g8i1Hn8WdArz2TYciWWrPErBEALBpXsu7Oq5AT97wAhcAIf3B3dV5p/bXgnhALA/SodNBf/wr0U2f5VzTDyJ9eNZURt16jSxDgXkuKKY/rimLqqt2FC1SeicmTql1nVyHG3j9Q== root@controller"]
+  server = "${module.aws_server.private_names[0]}"
+  mirror_public_name = "${module.aws_mirror.public_name}"
+  mirror_private_name = "${module.aws_mirror.private_name}"
+  testsuite = true
+}
+
+module "aws_client" {
+  dependencies = [
+    "${module.aws_mirror.depended_on}"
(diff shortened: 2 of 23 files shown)


moio:
Hey @jgleissner, thanks for the metric ton of work here!

Yeah, the AWS provider code was in a really sorry state, and a lot is needed to bring it back on-par with the other providers. I have that on my to-do list.

With Terraform 0.11, one could not really create "cross-provider" modules - eg. a `suse_manager` module that could be deployed either to AWS or libvirt depending on parameters. This created the situation that currently, most code in those providers is actually copied and pasted, and we have to make sure changes are ported across providers.

Historically this led to the eventual bitrotting of the AWS provider code, which you heroically brought back from the grave here. :+1:!


My long-term idea to fix this was the following:

1- make sure provider-related HCL code is relatively well separated from provider-unrelated parts. Some files are totally bound to a provider, such as:

https://github.com/uyuni-project/sumaform/blob/master/modules/libvirt/host/main.tf
https://github.com/uyuni-project/sumaform/blob/master/modules/openstack/host/main.tf

But others, only have a few bits of provider-related code, with the majority being really completely copy and paste. In those cases, keep the provider-related code at the bottom of the file, after a comment, and literally copypaste everything else:

https://github.com/uyuni-project/sumaform/blob/master/modules/libvirt/host/variables.tf#L91
https://github.com/uyuni-project/sumaform/blob/master/modules/openstack/host/variables.tf#L81

I have had mixed success at keeping this enforced between OpenStack and libvirt backends. I never really tried to make it for the AWS backend due to lack of time.

2- after migration to 0.12 is done (currently it's [in progress](https://github.com/uyuni-project/sumaform/pull/629)), try to unify all copypasted code between OpenStack and libvirt. Hopefully that should be easier given copypasted parts are clearly separated from others.

3- at this point it should be relatively easy to bring back the AWS backend.



May I ask you how the plan above sounds like? What are your goals exactly, what are your timelines? Do you plan to cover Azure/GCP or any other Public Cloud on top of AWS?


Once again thanks a lot, I will have a deep look as soon as possible.

jgleissner:
Hi @moio , basically our goal is to have automated testing for the SUSE Manager images we produce for the public clouds. To this date, testing those images is a pretty manual process. There isn't really a strict timeline, but the idea is to have something in place before the next Manager release. Azure and GCP should be covered too. I started with AWS as there were templates already, and obviously I wanted to get some feedback on the AWS approach.

There are obviously commonalities between the different platforms, so if terraform 0.12 can help to avoid code duplication that would be great. I'm relatively new to terraform and haven't looked into 0.12 yet, which I should probably do. I also did a couple of things that feel a little hackey, like creating dependencies between the resource as a ran into timing issues. I'm not entirely sure whether that's needed, and perhaps 0.12 brings improvements in that regards, so any feedback you would have would be appreciated.

Anyhow, I was able to run the spacewalk testsuite successfully (to some degree) in a cluster set up with these templates, with some restrictions. I used specifically built images for controller and mirror hosts, both based on released openSUSE images, but with required packages, most importantly salt-minion, pre-installed. The mirror storage snapshot I used is somewhat manually set up, but I'm working on a script to set up a suitable snapshot from scratch. And I cloned the test suite repository, as I had to disable the very first test, which creates the admin user. The SUSE manager images we have in the public clouds have a hook script that creates the admin user on setup, so that step will fail. With that disabled, I could run the test suite with all primary tests passing. So not quite out-of-the-box yet, but I hope it's a useful start.

salt/mirror/mirror.sh
```diff
@@ -8,6 +8,24 @@ minima -c /root/minima.yaml sync 2>&1 | tee /var/log/minima.log
 
 apt-mirror
 
+# check for nvidia repo and create links, if necessary
```
moio: I'd recommend moving this in salt/mirror/init.sls, if at all needed. When are those nVIDIA repos used?

salt/mirror/minima-pubcloud.yaml
```diff
@@ -0,0 +1,152 @@
+storage:
+  type: file
+  path: /srv/mirror
+
+scc:
```
moio: I understand that for the Public Cloud team current use case there is no need for mirroring IBS repos.

Questions are: are you actually sure about that? In any case, what would be your recommendation in case they are actually needed (we will likely be using public clouds for development more in the future, and I always thought that the current rsync-based approach is pretty hacky)?

salt/mirror/mirror.sh
jgleissner: The test suite adds SLED channels, and the SLED product requires nvidia repos to be available, otherwise it won't be visible.

salt/mirror/minima-pubcloud.yaml
jgleissner: Since the dummy test packages used by the test suite were moved to OBS, there's nothing really required from IBS to run the test suite in the public clouds, fortunately. The SUSE Manager images already have the SUSE Manager packages installed, obviously, so there shouldn't be the need to install anything from IBS, and in fact this would actually be unwanted, as we want to test the image as-is.

SUSE Manager images could be built from devel repositories too to create instances in the public cloud and run the test suite against them, so in theory this approach should also work for devel snapshots, as long as you don't need to make something unreleased available to the clients (like Manager-Tools packages, not sure whether that could be a problem).

jgleissner:
Hi @moio, any updates, esp. the port to terraform 0.12?

moio:
I am really sorry I could not keep up with this. The 0.12 port looks nice, with the libvirt part working as expected and now a big decision to drop the openstack backend altogether.

So my next step is to try and make it work with AWS.

I will dedicate time in this direction tomorrow and hopefully be back to you soon.

Thanks for your patience and once again apologies for taking this long

modules/aws/evil_minions/main.tf
```diff
@@ -66,7 +66,7 @@ EOF
 
   provisioner "remote-exec" {
     inline = [
-      "sh /root/salt/first_deployment_highstate.sh"
+      "bash /root/salt/first_deployment_highstate.sh"
```
moio: Was this module tested at all? Does it work? It's not really needed for the basic testsuite.

modules/aws/mirror/main.tf
```diff
@@ -33,37 +33,24 @@ resource "aws_instance" "instance" {
   tags {
     Name = "${var.name_prefix}-mirror"
   }
-}
-
-resource "aws_ebs_volume" "data_disk" {
```
moio: What is the reason to remove this resource?

modules/aws/mirror/main.tf
```diff
@@ -33,37 +33,24 @@ resource "aws_instance" "instance" {
...
-}
 
-resource "aws_volume_attachment" "data_disk_attachment" {
-  device_name = "/dev/xvdf"
-  volume_id = "${aws_ebs_volume.data_disk.id}"
-  instance_id = "${aws_instance.instance.id}"
 }
 
 resource "null_resource" "mirror_salt_configuration" {
-  triggers {
```
moio: Was this `triggers` block not necessary or wrong in some way?

modules/aws/mirror/main.tf
```diff
@@ -33,37 +33,24 @@ resource "aws_instance" "instance" {
...
+  connection {
+    host = "${aws_instance.instance.public_dns}"
+    private_key = "${file(var.key_file)}"
+    user = "${var.ssh_user}"
   }
-
   connection {
     host = "${aws_instance.instance.public_dns}"
     private_key = "${file(var.key_file)}"
+    user = "${var.ssh_user}"
```
moio: If I read the diff correctly, this seems duplicated now. Any particular reason for that, am I missing anything?

modules/aws/mirror/main.tf
```diff
@@ -33,37 +33,24 @@ resource "aws_instance" "instance" {
...
   }
 
   provisioner "file" {
     source = "salt"
-    destination = "/root"
```
moio: In the libvirt backend we deploy salt files to `/root` and I would like to keep that uniform in the light of an upcoming unification of the backends. Is there a particular reason to prefer `/tmp` here?

salt/mirror/init.sls
```diff
@@ -117,7 +123,7 @@ mirror_directory:
   mount.mounted:
     - name: /srv/mirror
     - device: /dev/{{grains['data_disk_device']}}1
-    - fstype: ext4
+    - fstype: {{fstype}}
```
moio: While I appreciate being flexible is good, I wonder if and why a particular filesystem choice is important here. What is the motivation to use `xfs`, is this particular to AWS in any way?

salt/mirror/mirror.sh
moio: I see that we do have both a `SLE-15-GA-Desktop-NVIDIA-Driver` repo coming from SCC and a download.nvidia.com one. Are both needed? What is the reason why we cannot get everything we need from SCC?

Maybe I am missing something, I believed we had those packages in our CDN, is that incorrect?

salt/mirror/mirror.sh
```diff
@@ -8,6 +8,24 @@ minima -c /root/minima.yaml sync 2>&1 | tee /var/log/minima.log
...
+    if [ ! -e ${full_path} ]; then
+      ln -s ../suse/$dir ${full_path}
+    fi
+  done
+fi
+
 jdupes --linkhard -r -s /srv/mirror/
 
-chmod -R 777 .
+#chmod -R 777 .
```
moio: Why was this removed? I remember seeing errors (403) in the past due to incorrect permissions.

salt/repos/default.sls
```diff
@@ -170,10 +170,12 @@ test_update_repo:
 os_pool_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-SERVER/12-SP4/x86_64/product/
+    - name: SLE-12-SP4-x86_64-Pool
```
moio: Is this because of failures in the testsuite?  I think we had a follow-up patch to fix those recently.

modules/aws/controller/main.tf
```diff
@@ -0,0 +1,108 @@
...
+      "sudo mv /tmp/github_deploy_key /root/.ssh/",
+      "sudo chown root:root /root/.ssh/github_deploy_key",
+      "sudo chmod 400 /root/.ssh/github_deploy_key",
+      "sudo mv /tmp/ssh_config /root/.ssh/config",
+      "sudo chown root:root /root/.ssh/config",
+      "sudo chmod 400 /root/.ssh/config",
+      "sudo bash /root/salt/first_deployment_highstate.sh",
+      "sudo gem install parallel parallel_tests minitest",
+      "sudo zypper in -y npm8",
+      "sudo npm install cucumber-html-reporter"
```
moio: These should definitely be moved into the controller's Salt states. Why are these only needed for AWS and not in libvirt?

modules/aws/controller/main.tf
```diff
@@ -0,0 +1,108 @@
...
+ StrictHostKeyChecking no
+EOF
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mv /tmp/grains /etc/salt/grains",
+      "sudo mv /tmp/salt /root/salt",
+      "sudo mv /tmp/github_deploy_key /root/.ssh/",
+      "sudo chown root:root /root/.ssh/github_deploy_key",
```
moio: Why do we need a specific mechanism for the AWS backend here?

main.tf.aws-create-mirror-snapshot.example
```diff
@@ -0,0 +1,82 @@
+// This template can be used to create a mirror hosts data snapshot.
```
moio: What is the reason to keep this file separate from `main.tf.aws-testsuite.example`?

Is the mirror even optional in AWS?

modules/aws/create_aws_mirror_snapshot
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/python3
+
+# Copyright 2019 SUSE Linux Solutions GmbH
+#
+# create_aws_mirror_snapshot is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation, either version 3 of the License,
+# or (at your option) any later version.
```
moio: Would it be an issue to keep this file aligned with the general license of this project, ie. BSD?

modules/aws/create_aws_mirror_snapshot
```diff
@@ -0,0 +1,134 @@
...
+params = get_terraform_output()
+
+mirror_host = params["mirror_host_name"]["value"]
+print("Connecting to mirror host '{}'".format(mirror_host))
+  with SSHConnection(mirror_host, "ec2-user", creds["key_file"]) as conn:
+      print("Starting mirror sync (this will take a while)")
+      conn.exec_command("sudo bash /root/mirror.sh")
+      conn.exec_command("sudo systemctl stop nfs-server")
+      conn.exec_command("sudo systemctl stop apache2")
+      conn.exec_command("sudo umount /srv/mirror")
```
moio: Is it really necessary to do the unmounting? Practical experience I had is that live snapshots worked just fine. Was it just luck? :wink:

modules/aws/create_aws_mirror_snapshot
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/python3
```
moio: Is there any particular reason why we coudn't / shoudn't use Terraform for this?

Back in the day when the original code was written, the AWS provider didn't support snapshot creation, but apparently now it does:

https://www.terraform.io/docs/providers/aws/r/ebs_snapshot.html

What would be pros and cons from your perspective?

modules/aws/host/main.tf
```diff
@@ -2,7 +2,16 @@ terraform {
     required_version = "~> 0.11.7"
 }
 
+resource "null_resource" "dependency_getter" {
+  provisioner "local-exec" {
+    command = "echo ${length(var.dependencies)}"
```
moio: Is this some kind of a hack to force dependencies, or what is the purpose for this `echo` command?

main.tf.aws-testsuite.example
```diff
@@ -0,0 +1,234 @@
...
+  key_name = "${local.key_name}"
+  key_file = "${local.key_file}"
+  server = "${module.aws_server.private_names[0]}"
+  client = "${module.aws_client.private_names[0]}"
+  minion = "${module.aws_minion.private_names[0]}"
+  git_username = "" // put github credentials here if necessary
+  git_password = "" // alternatively store a private deploy key in 'github_deploy_key' file
+  git_repo = "ssh://git@github.com/SUSE/spacewalk"
+  git_branch = "Manager-${local.prod_version}"
+  repos_to_sync = "sles12-sp4-pool-x86_64 sles12-sp4-updates-x86_64 sle-manager-tools12-pool-x86_64-sp4 sle-manager-tools12-updates-x86_64-sp4" // repos required for testsuite 'init_clients' set
```
moio: I am a bit uncertain about this new mechanism. Why is this not necessary in the `libvirt` version of the testsuite?

salt/default/init.sls
```diff
@@ -1,7 +1,9 @@
 include:
   - default.locale
   - default.minimal
+  {% if not grains.get('no_install') | default(false) %}
```
moio: What exactly do we want to prevent to install here, and what is the reason for that?

salt/first_deployment_highstate.sh
```diff
@@ -3,6 +3,7 @@
 # Applies the "minimal" state in isolation to set the hostname and update Salt itself
 # then applies the highstate
 
+exec &> >(tee /var/log/highstate.log)
```
moio: Please pardon my ignorance, I'm pretty bad at bash. What is the meaning of this command?

salt/mirror/init.sls
```diff
@@ -2,9 +2,14 @@ include:
   - default
 
 system_update:
+{% if not grains.get('no_update') | default(false) %}
```
moio: Why should we not want to have an up-to-date mirror? IIRC it's used as bastion host in AWS, so the last thing one wants is to have outdated software on it. Am I missing anything?

salt/repos/suse_manager_server.sls
```diff
@@ -1,6 +1,6 @@
 {% if 'suse_manager_server' in grains.get('roles') %}
 
-{% if '3.2' in grains['product_version'] %}
+{% if '3.2' in grains['product_version']|string %}
```
moio: What is the reason for this extra filter?

modules/aws/evil_minions/main.tf
jgleissner: No, but I changed first_deployment_highstate.sh to redirect output into a log file for better debugging, and since that requires bash I've change the invocation. I just realise that I should have changed it in the libvirt and openstack modules too.

Could be reverted to plain sh of course, if logging the output of the salt orchestration is deemed unnecessary.

modules/aws/mirror/main.tf
jgleissner: It's not removed, just moved down a little in the file. Sorry, the patch could probably be a bit smaller and less confusing.

modules/aws/mirror/main.tf
jgleissner: Actually, not sure why I removed that. Might have been an accident actually. I'll double-check.

modules/aws/mirror/main.tf
jgleissner: Yes, that's duplicate. May have even something to do with the removed `triggers` block. Possibly a cut&paste accident. I'll fix it.

modules/aws/mirror/main.tf
jgleissner: Yes. Root login is disabled in the public cloud images, in AWS you need to log in as `ec2-user` and then sudo. This also means you cannot provision files directly to `/root`. The remote execution block in the `mirror_salt_configuration` resource moves it to `/root`.

salt/mirror/init.sls
jgleissner: My personal motivation was that I started with an xfs mirror volume and then realised it's actually hardcoded to ext4 in sumaform and I didn't want to re-sync everything, so I thought I make it configurable to be able to use my xfs volume without breaking anything. It's not important or in any way particular to AWS, other than the fact that our images use xfs root volumes.

salt/mirror/mirror.sh
jgleissner: Commenting `chmod -R 777 .` was probably a knee-jerk reaction to a command that makes the whole tree world-writable. ;)  `chmod -R a+r` should suffice.

As for the nvidia packages, syncing the SCC nvidia repo with minima doesn't actually download the packages, hence the http download and the fiddling with the directories in the mirror tree.

salt/repos/default.sls
jgleissner: Yes, it resulted in testsuite failures when not named that way. I'll check the latest code base.

modules/aws/controller/main.tf
jgleissner: I wanted to touch as little as possible outside the AWS modules, but I strongly assume the packages will be necessary in libvirt too to run the test suite, so I agree this should go into the salt states.

modules/aws/controller/main.tf
jgleissner: As mentioned, root login is disabled in the public clouds, so we'll have to go the sudo route.

main.tf.aws-create-mirror-snapshot.example
jgleissner: This template only creates a mirror instance, intended to be used together with `modules/aws/create_aws_mirror_snapshot` to create a mirror data snapshot from scratch. I thought this may be handy.

modules/aws/create_aws_mirror_snapshot
jgleissner: Of course, sorry.

modules/aws/create_aws_mirror_snapshot
jgleissner: I'd say it's definitely safer, as a mounted filesystem is not guaranteed to be fully up-to-date. After the snapshot is created, the instance is most probably destined to be disposed of anyway.

modules/aws/create_aws_mirror_snapshot
jgleissner: I suppose it's possible. My terraform knowledge is still somewhat limited I'm afraid. I guess the sync script could be run as a remote exec provisioner. The snapshot creation would need to be dependent on the provisioner. Dependencies seem to be a bit of a tricky thing in terraform though. And the snapshot should obviously not be deleted on `terraform destroy` which I guess is possible somehow.

One potential issue comes to mind. I've ran into issues with long running scripts (salt orchestrations in this case) where the ssh connection dropped and terraform didn't realise and basically waited forever for the remote execution to finish. This kind of situation is easier to handle in a bespoke script.

modules/aws/host/main.tf
jgleissner: Yes, I stole, I mean recycled, a dependency hack I found for terraform because I ran into issues. For instance, if the mirror isn't ready yet the client orchestrations might fail. It's not really pretty though.

main.tf.aws-testsuite.example
jgleissner: I've never tried the libvirt version, so I can't comment whether it's necessary or not. However, the test suite in SUSE/spacewalk (Manager-4.0 branch) does abort repository syncing as the last step in the `reposync` run set, but the next run set `init_clients` requires the SLE 12 repos to be available. I would suspect that is the same in libvirt. I really feels like a test suite issue to me, which I'm trying to work around with this hack.

salt/default/init.sls
jgleissner: The idea is that we want to test the SUSE Manager image we produced for the public clouds as-is. If we install packages we potentially modify the installation and may mask errors in the image build.

salt/first_deployment_highstate.sh
jgleissner: This copies the output of the script to a file, which I though could be useful for debugging. May be reverted if you don't think it's necessary. I found it useful to have a log of the salt orchestration on the system.

salt/mirror/init.sls
jgleissner: Updating the system can take a while and this was one of the situations where I ran into ssh connection drops which resulted in provisioning failing. Also, a full system update may require a reboot which is not handled. The cloud paradigm would be to keep the mirror host image (reasonably) up to date rather than going through a lengthy update process at instance deployment time.

salt/repos/suse_manager_server.sls
jgleissner: Not sure why it apparently doesn't happen for you, but for me the orchestration fails otherwise. `grains['product_version']` expands to 4.0 and is considered a float, and then compared to a string which results in a type error. Explicitly converting the product version to a string fixes this.

salt/mirror/mirror.sh
moio: I'd suggest to revert the change here and instead using `chmod -R a+r` via a separate PR to be independently merged.

salt/first_deployment_highstate.sh
moio: I can agree with this change. I would ask you to have this change in a separate PR, to be merged autonomously (including the sh -> bash change), with a note in our README.

salt/repos/suse_manager_server.sls
moio: What version of Salt, from what repo, is running when you face that error?

jgleissner: The SUSE Manager 4.0 images are based on SLE 15 SP1, so the salt package is from the SLE 15 SP1 update repository. The latest I used was salt-2019.2.0-6.3.5 (not the very latest update, the image is a couple months old; can try a newer one if you wish).

rjmateus:
Hey @jgleissner 

We are working now on porting AWS support for the new sumaform structure. 
It can be seen here: https://github.com/uyuni-project/sumaform/pull/687

Thanks

salt/mirror/mirror.sh
juliogonzalez: @moio see https://github.com/uyuni-project/minima/issues/18 I was able to reproduce it.

Minima is not able to download such channels when defined.

moio:
Hey @jgleissner, with #687 finally merged, I think it is now time to close this PR and reset any work around this effort, as we changed the architecture so much that rebasing does not make a lot of sense.

I know it's not nice to have to close a body of work like the one you submitted here, and feel sorry about the fact that timing between the migration to Terraform 0.12 and the re-architecture of backends was really bad. I am also noting that keeping "soon-to-be-rewritten" code in a repo might create the wrong impression for contributors - who might start building more on top. I will try not to repeat the mistake going forward.

On the positive side, work here was definitely not wasted. Even if commits were never merged as-is, Ricardo and I took a lot of ideas and snippets from your branch when preparing #687, and hopefully also maintained all essential features your use case required.

I am also confident that running SUSE Manager/Uyuni tests in AWS and, in future, other Public Clouds will be much easier with the new structure in place.

So it is really my hope this is a "see you later" and not a "goodbye"!

Once again thanks,

moio

juliogonzalez:
Hey, and I used this PR as well, to add the support for nvidia repos no matter the SLE version or SP version (https://github.com/uyuni-project/sumaform/pull/697and https://github.com/uyuni-project/sumaform/pull/698).

So I agree, working on this PR was not a waste of time at all, and helped moving sumaform forward!

Thanks a lot @jgleissner!

rjmateus:
Hey. Thanks for this massive work here. I used several pieces from this PR in the on #687 , so this was very helpful work.
This work allows us to have a head start on the aws refactoring for terraform independents modules.

Thanks a lot for the contribution @jgleissner

jgleissner:
I'm glad I could contribute. :)

PR #630, opened 2019-10-19
Fix mirror config

mcalmer:
## Warning

## What does this PR change?

Fixes 2 issues:

1. testsuite repos must be synced with all archs incl. source packages
2. add Ubuntu devel repos


moio:
Thanks

PR #629, opened 2019-10-18
Port to Terraform 0.12

moio:
## What does this PR change?

It adapts code to run with the newest Terraform version.

Fixes #547

PR #631, opened 2019-10-25
Drop OpenStack backend

moio:
## What does this PR change?

This PR drops the whole OpenStack backend, which is now getting de-prioritized.


.travis.yml
```diff
@@ -1,4 +1,4 @@
-dist: xenial 
+dist: bionic
```
srbarrios: Are the changes in that file related to the topic of the PR?

moio: Yes, in an indirect way. Basically the Travis tests we had were exclusively on OpenStack, although they were executed on all PRs.

Dropping OpenStack meant some rework so that those really use the libvirt backend at this point.

pgonin:
Hi Silvio,

SUSE OpenStack Cloud 9 still has a 3 years lifetime.
ECP will be supported for at least that time.
Internal scalability testing capability will likely be available on ECP
There are also investments to build an Engineering Cloud in Nuremberg.

It may be wise to keep OpenStack support in sumaform for a little bit longer.

moio:
Paul, thanks for contacting me, I was aware of the facts you pointed out.

FMPOV, I am [in the middle of several refactoring steps](https://github.com/SUSE/spacewalk/projects/45), most important of which is the port of the project to Terraform 0.12.

Biggest feature of this new version are language changes. Which for code bases like sumaform, it means extensive changes. For example [this PR, which takes care of the libvirt backend](https://github.com/uyuni-project/sumaform/pull/629/files) it's more than 1000 changed lines over more than 70 files. Porting OpenStack would presently entail a similar effort.

Now there is only limited use of this backend in our Team - [I even asked in our internal mailing list to speak up if anybody is impacted but received no response in 10 days](http://mailman.suse.de/mlarch/SuSE/galaxy-devel/2019/galaxy-devel.2019.10/msg00185.html). Thus there seem to be no short-term need to keep this maintained.

Being low on available time I think I will still merge this PR to move the whole project to 0.12 as fast as possible - knowing as usual that any decision and any PR can always be reverted at any later point in time if ever there is a new need, or I get capacity, or someone volunteers to do the port and maintain it.

I also [already tagged the latest version](https://github.com/uyuni-project/sumaform/releases/tag/last-sumaform-with-openstack-support) that contains OpenStack support, so anything in the current version of sumaform will continue to work for the foreseeable future - on Terraform 0.11 - simply by checking out that tag instead of `master`, which as always contains the latest changes.

Please let me know in case I am overlooking something else.

dirkmueller:
@moio so looking at that PR, most of those changes seem to be mechanical, so 1000s of hcanged lines doesn't seem like linear effort to me. 

Ultimatively its your decision of course, but the likelyhood of being able to do similar scale of testing that you can currently do inside VMs  (either in provo or in nbg) on plain hardware is very small. 

if you don't care about many multi node deployments being testable in parallel (which you'd need when you do significant testing on a e.g. per PR testing) then dropping that should be fine.

srbarrios:
Hi @moio , I understand your point of view and the worries of Paul, my opinion about it is that we might contact the people using the OpenStack support and propose them to maintain and migrate to Terraform 0.12 that part.
If no one has capacity to maintain it, I still understand that our team can't afford to keep maintaining that code on new versions, but maybe is a decission that must be agreed with our managers, at the end is a trade-off of capacity to work on that as far as I understood.

moio:
> @moio so looking at that PR, most of those changes seem to be mechanical, so 1000s of hcanged lines doesn't seem like linear effort to me.

@dirkmueller, you are right in pointing out that many of the changes are mechanical - that is a direct consequence of the fact many of the underlying language changes are syntactical. I did not intend to imply that the porting effort is in any way similar to writing those lines from scratch.

The point I was trying to make is that the effort of porting the `libvirt` backend is roughly equivalent to the effort of porting the `openstack` backend. 

One thing that might not be immediately obvious is that those two backends are very similar in structure/size/complexity - indeed a very significant part of the code should in principle be in common. The code bases are actually separate (meaning copy-paste-adapted, with all the usual perils of divergence) because of limitations in earlier language versions. We strived [to keep them aligned](https://github.com/uyuni-project/sumaform/blob/master/.github/PULL_REQUEST_TEMPLATE.md#warning) with mixed results - ultimately for a change like this it means one has to go through the (partly mechanical) process just twice. Consider also that one important chunk of that work is manual testing, and that just has to be repeated.

> Ultimatively its your decision of course, but the likelyhood of being able to do similar scale of testing that you can currently do inside VMs (either in provo or in nbg) on plain hardware is very small.
> 
> if you don't care about many multi node deployments being testable in parallel (which you'd need when you do significant testing on a e.g. per PR testing) then dropping that should be fine.

There are organization-specific considerations to be made in this regard, and I prefer addressing them via internal channels rather than this public PR, so I will follow up with more details via email if you don't mind.

Anyway the gist of it is that I am aware of the internal hardware situation and my Team is actually exploring ideas to use more Public Clouds for future projects, hence the decision of reducing efforts in the `openstack` backend to have more time for that instead.

In any case, thanks for your interest and for taking the time to take a look at our work here!

pgonin:
I know that the switch from 0.11 to 0.12 is painful despite being a minor version. 
There is the same issue currently on CaaSP but the choice as been to remain on 0.11 so far as the benefits of switching to 0.12 were unclear.

Sure it is your choice and I understand it is based on efforts needed to adapt the code.

My point is that currently and upcoming available capacity is and will be on ECP and Engineering Cloud in Nuremberg.
AWS might become an alternative but I am not sure how much will be possible to spend although I requested to clarify what are the budget possibilities for the various directions (but investing on dedicated hardware in Nuremberg and Provo will be limited, and painful to achieve)

moio:
As neither current use cases, nor plans for future use cases, nor voluntaries for maintainership have been found so far I am proceeding to merge this PR.

About the concerns that have been expressed, I can only once again highlight that this does not represent any long-term, set-in-stone decision for the project and that both me and my Team remain completely open to discuss reversing this choice if and when conditions change.

Thanks for the discussion

PR #632, opened 2019-10-28
Use new Uyuni POOL repositories

juliogonzalez:
## What does this PR change?

As part of the new Uyuni product definitions, we are removing the version from the POOL repositories. And we decided to remove the distribution name from the URL as well (so it's now `images` so we don't need to change it when we change the base OS.

This means than in the future no sumaform changes will be needed for new Uyuni versions or base OS changes (also good for development)

Requires https://github.com/uyuni-project/uyuni/pull/1535


juliogonzalez:
```
Apply complete! Resources: 4 added, 0 changed, 0 destroyed
```
Merging :-)

juliogonzalez:
The Uyuni PR is not required for sumaform, as it worked.

PR #636, opened 2019-11-13
refacto in host module to return list of ID

rjmateus:

## What does this PR change?

Refactor in the host module to return a list of ID's instead of the ID of the first domain created.
Adapt modules that use this field to behave according.


modules/libvirt/controller/variables.tf
```diff
@@ -30,7 +30,9 @@ variable "branch" {
 
 variable "server_configuration" {
   description = "use module.<SERVER_NAME>.configuration, see main.tf.libvirt-testsuite.example"
-  type        = map(string)
+  default = {
```
moio: Hmm, no, I would not do this, as it would make the server configuration optional. This is actually a mandatory parameter without which the controller can't be configured.

What is the reason to set this default?

modules/libvirt/controller/variables.tf
```diff
@@ -43,12 +45,16 @@ variable "proxy_configuration" {
 
 variable "client_configuration" {
   description = "use module.<CLIENT_NAME>.configuration, see main.tf.libvirt-testsuite.example"
-  type        = map(string)
+  default = {
```
moio: Same as above

modules/libvirt/controller/variables.tf
```diff
@@ -43,12 +45,16 @@ variable "proxy_configuration" {
...
+    hostname = "null"
+  }
 }
 
 variable "minion_configuration" {
   description = "use module.<MINION_NAME>.configuration, see main.tf.libvirt-testsuite.example"
-  type        = map(string)
+  default = {
```
moio: Same as above

modules/libvirt/controller/variables.tf
rjmateus: I cannot pass a map since we have fields of two different types. This was a change to allow it. Probably need to change to object, or revert it when changing the hostname.

PR #634, opened 2019-11-11
Base Head on 4.1, add suport for SLE15SP2

juliogonzalez:
## What does this PR change?

- Base Head on 4.1
- Add suport for SLE15SP2
- Preadd support for future 4.1 stable

Fixes https://github.com/SUSE/spacewalk/issues/10009

PR #638, opened 2019-11-13
Basesystem repos are part of default.sls

juliogonzalez:
## What does this PR change?

This explains https://github.com/uyuni-project/sumaform/pull/634#discussion_r345767666

Only basesystem is part of `default.sls`, all other modules for server and proxy are part of `suse_manager_server.sls` and `suse_manager_proxy.sls`.

So if we add a new OS for new SUSE Manager or Uyuni versions, we also need to change `default.sls`

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -288,6 +288,16 @@ os_update_repo:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Basesystem/15-SP1/x86_64/update/
 {% endif %} {# '15.1' == grains['osrelease'] #}
 
+{% if '15.2' == grains['osrelease'] %}
+os_pool_repo:
+  pkgrepo.managed:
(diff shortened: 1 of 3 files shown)


salt/repos/default.sls
```diff
@@ -288,6 +288,16 @@ os_update_repo:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Basesystem/15-SP1/x86_64/update/
 {% endif %} {# '15.1' == grains['osrelease'] #}
 
+{% if '15.2' == grains['osrelease'] %}
+os_pool_repo:
```
juliogonzalez: IMHO we should not call this `os_pool_repo` but  `module_basesystem_pool_repo`, as it is what it is. Same for the `update` repository.

juliogonzalez:
Head and 4.0 tests passed.

salt/repos/default.sls
moio: That would require changes in the testsuite code. Doable but has to be orchestrated

moio:
Thanks for fixing!

salt/repos/default.sls
srbarrios: @juliogonzalez if you need help on the testsuite part let's us know, you can create a card address it when possible. Thank you!

PR #641, opened 2019-11-15
0.12.additional repos for each server

rjmateus:
## What does this PR change?

- Refactor of the available hosts with the option to define additional repositories for each host.
- Remove variable proxy and use available hosts instead

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -20,6 +20,16 @@ module "base" {
   network_name       = lookup(var.provider_settings["libvirt"], "network_name", "default")
   bridge             = lookup(var.provider_settings["libvirt"], "bridge", "")
   additional_network = lookup(var.provider_settings["libvirt"], "additional_network", "")
+
+}
+
+locals {
+  server_full_name = "${var.name_prefix}srv.${var.domain}"
+  hosts            = keys(var.host_settings)
+  macs = { for host_key in local.hosts :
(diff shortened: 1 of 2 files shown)


rjmateus:
We now can have a host in the cucumber test module without any additional configuration using null or empty object.
Example:
`
hosts = {
    ctl = {
      mac = "AA:B2:93:00:00:71"
    }
    srv = { }
    pxy = null
    }

`

modules/cucumber_testsuite/variables.tf
```diff
@@ -45,14 +45,16 @@ variable "use_shared_resources" {
...
 // cucumber_testsuite
-variable "proxy" {
-  description = "whether or not to include a Proxy in the test"
-  default     = true
-}
-
-variable "optional_clients" {
-  description = "list of optional clients to test. Defaults to all"
-  default     = ["min-centos7", "min-ubuntu1804", "minssh-sles12sp4", "min-pxeboot", "min-kvm"]
+variable "hosts" {
```
moio:
```suggestion
variable "host_settings" {
```

modules/cucumber_testsuite/main.tf
```diff
@@ -53,13 +53,15 @@ module "srv" {
 
 locals {
   server_full_name = "${var.name_prefix}srv.${var.domain}"
-  macs             = lookup(var.provider_settings["libvirt"], "macs", {})
+  available_host = keys(var.hosts)
```
moio:
```suggestion
  hosts = keys(var.host_settings)
```

PR #642, opened 2019-11-19
Add per-host image setting

moio:
## What does this PR change?

Adds a setting to change the image per host (on hosts for which it makes sense to configure an image)

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -72,7 +72,7 @@ module "pxy" {
 
   base_configuration = module.base.configuration
   product_version    = var.product_version
-  image              = lookup(local.images, "pxy", null)
+  image              = lookup(local.images, "pxy", "default")
   name               = "pxy"
 
   server_configuration      = { hostname = local.server_full_name, username = "admin", password = "admin" }
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -20,7 +20,6 @@ module "base" {
   network_name       = lookup(var.provider_settings["libvirt"], "network_name", "default")
   bridge             = lookup(var.provider_settings["libvirt"], "bridge", "")
   additional_network = lookup(var.provider_settings["libvirt"], "additional_network", "")
-
 }
 
 locals {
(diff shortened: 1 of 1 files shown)


modules/cucumber_testsuite/main.tf
```diff
@@ -30,12 +29,15 @@ locals {
...
   host_key => lookup(var.host_settings[host_key], "additional_repos", {}) if var.host_settings[host_key] != null }
+  images = { for host_key in local.hosts :
+  host_key => lookup(var.host_settings[host_key], "image", null) if var.host_settings[host_key] != null }
 }
 
 module "srv" {
   source                         = "../libvirt/suse_manager"
   base_configuration             = module.base.configuration
   product_version                = var.product_version
+  image                          = lookup(local.images, "srv", null)
```
rjmateus: If the image is not defined we should use the default value for the variable which is: "default".
```suggestion
  image                          = lookup(local.images, "srv", "default")
```

modules/cucumber_testsuite/main.tf
```diff
@@ -70,6 +72,7 @@ module "pxy" {
 
   base_configuration = module.base.configuration
   product_version    = var.product_version
+  image              = lookup(local.images, "pxy", null)
```
rjmateus: If the image is not defined we should use the default value for the variable which is: "default".
```suggestion
  image              = lookup(local.images, "pxy", "default")
```

modules/cucumber_testsuite/main.tf
```diff
@@ -30,12 +29,15 @@ locals {
   host_key => lookup(var.host_settings[host_key], "mac", null) if var.host_settings[host_key] != null }
   additional_repos = { for host_key in local.hosts :
   host_key => lookup(var.host_settings[host_key], "additional_repos", {}) if var.host_settings[host_key] != null }
+  images = { for host_key in local.hosts :
+  host_key => lookup(var.host_settings[host_key], "image", null) if var.host_settings[host_key] != null }
```
rjmateus:
```suggestion
  host_key => lookup(var.host_settings[host_key], "image", "default") if var.host_settings[host_key] != null }
```
If the image value is not defined for the host, which default should we use? At least suse manager and suse manager proxy use "default" string as default value.

moio: Others will use different defaults though. I'd leave it as it is for now

PR #645, opened 2019-11-19
Cucumber testsuite module

moio:
## What does this PR change?

This PR builds on top of #629 and adds a `cucumber_testsuite` module that represents the whole testsuite infrastructure (Server, optional Proxy, Controller and clients).


moio:
@Bischoff plan is to merge this immediately after merging #629. At this point I do not think we need a complete review, but feel free to take a look.

Bischoff:
@moio Congrats for the new "clients list" approach, it's way more elegant.

I did not really read this PR because TL:DR, but thanks for the long "testing" documentation.

As I already discussed in private, my only concern with the new setup is how I can now do client-specific settings, for example disable IPv6 on SSH minion and not all others. I understand this will be addressed later, right?

moio:
It is already addressed with this version: we do not have a client list any more, we have a host map. Any attribute can be added and applied on a per-host basis, see this example where we override `mac` and `additional_repos` for `srv` only:

https://github.com/uyuni-project/sumaform/pull/645/files#diff-07a607a47e37334a1189dd4da02df479R109-R114

The logic can be extended to any attributes you might need. So I think we are covered.

Bischoff:
> It is already addressed with this version: we do not have a client list any more, we have a host map.

OK

> Any attribute can be added and applied on a per-host basis, see this example where we override
>  `mac` and `additional_repos` for `srv` only:
> 
> https://github.com/uyuni-project/sumaform/pull/645/files#diff-07a607a47e37334a1189dd4da02df479R109-R114

OK. Now I get it, thanks.

> The logic can be extended to any attributes you might need. So I think we are covered.

OK. @moio Did you remove the hardcoded logic for the avahi refector that would be run only at the normal minion? I would like to do it like the rest in your new hosts map.

moio:
> OK. @moio Did you remove the hardcoded logic for the avahi refector that would be run only at the normal minion? I would like to do it like the rest in your new hosts map.

Yeah. The reflector is used (only on that minion) only if explicitly defined (default is not to use it).

PR #646, opened 2019-11-22
Refactoring: use new features in Terraform 0.12 to remove hacks

rjmateus:
## What does this PR change?
The main goal of this PR is to use objects to define host grains and serialized them using terraform native methods. 
This allows removing a set of hacks that were needed since we are using string templates to generate the file content.

This rector acts in two action items:
- remove null string as default value in vars
- pass grains as objects instead of strings
- Remove type check for base configuration and server configuration

Jenkins Run: 
- https://ci.suse.de/view/Manager/view/Manager-3.2/job/manager-3.2-cucumber-pipeline-NUE/669
- https://ci.suse.de/view/Manager/view/Manager-3.2/job/manager-3.2-cucumber-pipeline-NUE/677/

I need to remove the type check since it was defined as map and all fields were converted to string. 
To solve this we have two approaches:
1 - Do not specify the variable type and allow any type as input.
2- Define as object, and specify each field in the object. This needs to be propagated in several modules, and if we need to add another field, we should synchronize in all modules

In this case, since it is for internal use only, I think it is safe to not specify the variable type, and use solution 1.

**Some more detailed information from documentation:**
If we specify the base input as map, it should have all the same type and  If we define as map of any, it will be converted to string.
"_The keyword map is a shorthand for map(any), which accepts any element type as long as every element is the same type._ " (https://www.terraform.io/docs/configuration/types.html#collection-types)

We are not obligated to specify the variable type, and in that case, it can have any type:
"_The type argument in a variable block allows you to restrict the type of value that will be accepted as the value for a variable. If no type constraint is set then a value of any type is accepted._

_While type constraints are optional, we recommend specifying them; they serve as easy reminders for users of the module, and allow Terraform to return a helpful error message if the wrong type is used._"  (https://www.terraform.io/docs/configuration/variables.html#type-constraints)

### DIFF
--- a/modules/libvirt/client/variables.tf
+++ b/modules/libvirt/client/variables.tf
@@ -98,5 +98,5 @@ variable "mac" {
 
 variable "cpu_model" {
   description = "Define what CPU model the guest is getting (host-model, host-passthrough or the default)."
-  default     = ""
+  default     = null
 }
--- a/modules/libvirt/client/variables.tf
+++ b/modules/libvirt/client/variables.tf
@@ -53,8 +53,7 @@ variable "swap_file_size" {
 
 variable "ssh_key_path" {
   description = "path of additional pub ssh key you want to use to access VMs, see README_ADVANCED.md"
-  default     = "/dev/null"
-  # HACK: "" cannot be used as a default because of https://github.com/hashicorp/hil/issues/50
+  default     = null
 }
 
 variable "gpg_keys" {
--- a/modules/libvirt/client/variables.tf
+++ b/modules/libvirt/client/variables.tf
@@ -93,7 +93,7 @@ variable "running" {
 
 variable "mac" {
   description = "a MAC address in the form AA:BB:CC:11:22:22"
-  default     = ""
+  default     = null
 }
 
 variable "cpu_model" {
--- a/modules/libvirt/client/variables.tf
+++ b/modules/libvirt/client/variables.tf
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
 }
 
 variable "name" {
@@ -15,7 +14,6 @@ variable "product_version" {
 
 variable "server_configuration" {
   description = "use module.<SERVER_NAME>.configuration, see the main.tf example file"
-  type        = map(string)
 }
 
 variable "auto_register" {
--- a/modules/libvirt/host/variables.tf
+++ b/modules/libvirt/host/variables.tf
@@ -48,7 +48,7 @@ variable "quantity" {
 }
 
 variable "grains" {
-  description = "custom grain string to be added to this host's configuration"
+  description = "custom grain map to be added to this host's configuration"
   default     = {}
 }
 
--- a/modules/libvirt/host/variables.tf
+++ b/modules/libvirt/host/variables.tf
@@ -59,8 +59,7 @@ variable "swap_file_size" {
 
 variable "ssh_key_path" {
   description = "path of additional pub ssh key you want to use to access VMs, see README_ADVANCED.md"
-  default     = "/dev/null"
-  # HACK: "" cannot be used as a default because of https://github.com/hashicorp/hil/issues/50
+  default     = null
 }
 
 variable "gpg_keys" {
--- a/modules/libvirt/host/variables.tf
+++ b/modules/libvirt/host/variables.tf
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
 }
 
 variable "name" {
@@ -50,7 +49,7 @@ variable "quantity" {
 
 variable "grains" {
   description = "custom grain string to be added to this host's configuration"
-  default     = ""
+  default     = {}
 }
 
 variable "swap_file_size" {
--- a/modules/libvirt/host/variables.tf
+++ b/modules/libvirt/host/variables.tf
@@ -123,6 +123,6 @@ variable "cpu_model" {
 }
 
 variable "xslt" {
-  description = "module relative path to an xsl file to apply on the libvirt domain"
-  default     = ""
+  description = "xslt contents to apply on the libvirt domain"
+  default     = null
 }
--- a/modules/libvirt/host/variables.tf
+++ b/modules/libvirt/host/variables.tf
@@ -119,7 +119,7 @@ variable "additional_disk" {
 
 variable "cpu_model" {
   description = "Define what CPU model the guest is getting (host-model, host-passthrough or the default)."
-  default     = ""
+  default     = null
 }
 
 variable "xslt" {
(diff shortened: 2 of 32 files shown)


juliogonzalez:
Two questions:

1. Don't we need to update the markdown files? For example `TROUBLESHOOTING.md`
2. Did you check if the `main.tf` examples at the root folder need to be updated?

rjmateus:
> Two questions:
> 
> 1. Don't we need to update the markdown files? For example `TROUBLESHOOTING.md`
> 2. Did you check if the `main.tf` examples at the root folder need to be updated?

Thanks for the comment Julio.
Regarding the "md" files I don't think we need any update in the context of this PR.
I have updated the examples files.
Cheers

modules/libvirt/host/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

modules/libvirt/locust/variables.tf
```diff
@@ -1,11 +1,9 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

Also applies to `server_configuration` below.

modules/libvirt/minion/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

Also applies to `server_configuration` below.

modules/libvirt/minionssh/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

modules/libvirt/mirror/main.tf
```diff
@@ -17,15 +17,13 @@ module "mirror" {
...
-cc_password: ${var.base_configuration["cc_password"]}
-data_disk_device: vdb
-ubuntu_distros: [${join(", ", formatlist("'%s'", var.ubuntu_distros))}]
-use_mirror_images: ${var.base_configuration["use_mirror_images"]}
-
-EOF
+  grains = {
+    cc_username = var.base_configuration["cc_username"]
+    cc_password = var.base_configuration["cc_password"]
+    data_disk_device : vdb
```
moio:
```suggestion
    data_disk_device = "vdb"
```

I think this should not compile. Please test creating a mirror locally just to be sure.

modules/libvirt/mirror/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

modules/libvirt/pts/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

modules/libvirt/pxe_boot/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

modules/libvirt/suse_manager/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

modules/libvirt/suse_manager_proxy/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

Also applies to `server_configuration` in this file.

modules/libvirt/virthost/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
moio: Same question as in modules/libvirt/client/variables.tf - can we do better than `any` here?

Also applies to `server_configuration` in this file.

modules/libvirt/suse_manager/variables.tf
```diff
@@ -25,23 +24,23 @@ variable "wait_for_reposync" {
 
 variable "cloned_channels" {
   description = "a json formatted string representing a list of dictionaries containing SUSE channels information to clone"
-  default     = "null"
+  default     = null
   type        = string
```
moio: This could actually see some improvement. Back in the day we had no way of specifying an arbitrary data structure, so we ended up recommending to just use JSON-in-a-string:

https://github.com/uyuni-project/sumaform/blob/master/README_ADVANCED.md#cloning-channels-in-suse-manager-servers-upon-deployment

I'd suggest using proper objects now. The only place where this is used is in the PTS (which you can patch even without testing at this time).

Please do check that `cloned_channels` on a single Server works correctly in your local setup (ping me if you need any assistance). And remember to change it also in the README!

modules/libvirt/suse_manager/main.tf
```diff
@@ -38,54 +38,53 @@ module "suse_manager" {
   # HACK: work around "conditional operator cannot be used with list values"
```
moio: This hack ought to be removed (in the context of this PR or a following one!)

modules/libvirt/client/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "base_configuration" {
   description = "use module.base.configuration, see the main.tf example file"
-  type        = map(string)
```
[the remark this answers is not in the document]
rjmateus: The `any` type in map will not work since all values should have the same value. This is not true regarding the base configuration and in that case all values will be converted to `string`, changing the null value to null string.
Change it to object force us to be super verbose and specify all fields allowed. Since this object is used across all application modules, it will increase maintenance time in case we need to change a field in the base configuration output.

modules/libvirt/mirror/main.tf
moio: Fixed

moio:
Results from latest testsuite runs are positive, merging.

PR #651, opened 2019-12-04
mandatory minion and clients can set image

rjmateus:

## What does this PR change?
ISSUE: https://github.com/SUSE/spacewalk/issues/10235
Changes:
- Client, minion and sshminion renamed for names not tight with image version in use
- Image version to used in client, minion and sshminion loaded from a varible. Default value is: "sles12sp4"

We have the need to run test suites with different images for minions and clients. It was not possible since the image was hardcode to use "sles12sp4".

This will imply changing the minions name in all terraform-test-runer configuration files which use cucumber-testsuite module to run.

### DIFF
--- a/README_TESTING.md
+++ b/README_TESTING.md
@@ -80,7 +80,7 @@ host_settings = {
   pxy = {
     present = true
   }
-  minssh-sles12sp4 = {
+  sshminion = {
     present = true
   }
   min-centos7 = {
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -95,12 +95,12 @@ locals {
   minimal_configuration = { hostname = contains(local.hosts, "pxy") ? local.proxy_full_name : local.server_full_name }
 }
 
-module "cli-sles12sp4" {
+module "client" {
   source             = "../libvirt/client"
   base_configuration = module.base.configuration
   product_version    = var.product_version
-  name               = "cli-sles12sp4"
-  image              = "sles12sp4"
+  name               = "cli-${lookup(local.images, "client", "sles12sp4")}"
+  image              = lookup(local.images, "client", "sles12sp4")
 
   server_configuration = local.minimal_configuration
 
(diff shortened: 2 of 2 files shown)


rjmateus:
> Thanks for the nice work Ricardo. LGTM, but please test.

Eric if I understand correctly, you would test those changes in your branch for uyuni test suite refactor to use opensuse, and also update terraform-test-runner. Did I miss understood?
Thanks

srbarrios:
I'm fine, as far we still keeping all the possible clients in modules/libvirt/controller/variables.tf  ( I need it for QAM purposes)

Bischoff:
> > Thanks for the nice work Ricardo. LGTM, but please test.
> 
> @Bischoff if I understand correctly, you would test those changes in your branch for uyuni test suite refactor to use opensuse, and also update terraform-test-runner. Did I miss understood?
> Thanks

About testing: I just meant it looks good to me, but I can't guarantee I did not miss some error, so this PR would need some testing.

The openSUSE on Uyuni is a different problem. I had it work, but stopped before doing the real switch by changing the runner, and DHCP and DNS records. I had PRs ready for that, but due to all recent refactorings, they probably needs some love and care again. I'll probably recontact you about that. Issue is spacewalk#9798 .

moio:
@rjmateus can you please take care of the final round of testing and merging (no hurries)?

srbarrios:
Sorry to bother on last minute :trollface: Now I noticed we will use sles12sp4 as default image. I was thinking... could make sense to use sles15 if @Bischoff  already did the work to use them as our default clients in the testsuite?

P.S. Can be done in a future PR, for sure.

rjmateus:
Sure @srbarrios. I will make that change, test it and if everything goes well merge it

rjmateus:
@srbarrios @Bischoff @moio This PR doesn't make sense anymore since name and image are now a parameter for all clients/minions.
What we can do is change "variable names" for hosts in cucumber_testsuite, which should be addressed in https://github.com/SUSE/spacewalk/issues/10730

rjmateus:
I will close this PR and delete the associated branch.

Bischoff:
> @srbarrios @Bischoff @moio This PR doesn't make sense anymore since name and image are now a parameter for all clients/minions.

Correct.

> What we can do is change "variable names" for hosts in cucumber_testsuite, which should be addressed in [SUSE/spacewalk#10730](https://github.com/SUSE/spacewalk/issues/10730)

Yes, as you know it would have been okay to do the renaming, and the arbitrary number of clients, in separate PRs, but why not.

rjmateus:
> Yes, as you know it would have been okay to do the renaming, and the arbitrary number of clients, in separate PRs, but why not.
We could do it in two separated PR, but that would imply changing twice all main.tf files in sumaform-test-runner that uses cucumber_testsuite, and I think this will not bring a benefit that justifies it.

Bischoff:
> We could do it in two separated PR, but that would imply changing twice all main.tf files in sumaform-test-runner that uses cucumber_testsuite,

Correct.

> and I think this will not bring a benefit that justifies it.

I slightly disagree here :smile_cat: , but that's not important.

PR #654, opened 2019-12-10
Update taint wiki according to the new command interface

ncounter:
## What does this PR change?

Update `taint` wiki according to the new command interface

Using the previous command with the `-module` parameter I get the error message:
```
The -module option is no longer used. Instead, include the module path in the main resource address, like "module.foo.module.bar.null_resource.baz".
```


moio:
Thanks. And I thought I had fixed this long ago. Faulty memory... :wink:

PR #655, opened 2019-12-11
Adaptations for SLE 15 SP1 clients - batch 1

Bischoff:
## What does this PR change?

It prepares for testing with SLE 15 SP1 clients

* SLE 15 SP1 development tools repos are needed for Kiwi
* SLE 15 SP1 Python2 repos are needed for Docker
   (3.2 and 4.0 only; one can build containers with Python3 on 4.1)
* Using SLE 15 SP1 for containers too
* Mirror developement tools in minima
  (Python 2 and Containers seem OK already)

See also uyuni-project/uyuni#1721 for the cucumber side

### DIFF
--- a/salt/repos/minion.sls
+++ b/salt/repos/minion.sls
@@ -28,6 +28,16 @@ devel_updates_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Development-Tools/15-SP1/x86_64/update/
 
+{% if '3.2-nightly' in grains.get('product_version') or '4.0-nightly' in grains.get('product_version') %}
+python2_pool_repo:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Python2/15-SP1/x86_64/product/
+
+python2_updates_repo:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Python2/15-SP1/x86_64/update/
+{% endif %}
+
 {% endif %}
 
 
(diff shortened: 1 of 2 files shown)


Bischoff:
> Can you move these changes into the testsuite-specific file (unless those are really required for any use of a SLE 15 SP1 client)?
> 
> https://github.com/uyuni-project/sumaform/blob/master/salt/repos/testsuite.sls

After discussing with Silvio, moving it elsewhere will be done in a separate PR as part of sumaform refactorings, because it's more complicated than it seems.

> Additionally, please double check all are present in `minima.yaml`, or `mirror` based configurations will fail.

Ooops, right, I keep forgetting it... :blush:

Bischoff:
Added minima

salt/repos/minion.sls
```diff
@@ -14,11 +14,30 @@ containers_updates_repo:
 {% if '15' in grains['osrelease'] %}
 containers_pool_repo:
   pkgrepo.managed:
-    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Containers/15/x86_64/product/
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Containers/15-SP1/x86_64/product/
```
moio: Does this mean we will only support 15-SP1 and not 15 GA going forward?

Bischoff: * according to the code preceeding this one, you also consider only SLE 12 (no SP1 .. SP5)
* SLE15 is end of life in 10 days (apart from LTSS)
* in QA we will test only SP1
* this is about docker and kiwi, and there are anyway related bug fixes that were not backported to SLE 15 like https://bugzilla.suse.com/show_bug.cgi?id=1149686
* retail will officially not support SLE 15 as build host

Bischoff:
> LGTM, see one inline comment just to be sure

OK, see my reply to this comment.

I will merge now, we can still add SLE 15 and move to `testsuite.sls` in separate PRs.

PR #652, opened 2019-12-05
Refactor to provider independent modules

rjmateus:
## What does this PR change?

**add description**
Issue: https://github.com/SUSE/spacewalk/issues/10021

This PR change existing module to be independent of the backend in use. The goal is use these modules and delegate the backend creation on provider dependent modules.
To select which backend should be used a symbolic link needs to be created.

Existing modules migrations:

| Module name | Migrated | Local test | CI (Ref host) | CI (test suite) |
| ------------------ | ------------ | ------------- | ----------------- | ------------------ |
| cucumber_testsuite | X    | X              |                      | X                    |
| client | X | X | X |
| controller | X | X |
| grafana | X |  |
| locust | X |  |
| minion | X | X | X |
| sshminion | X | X | X |
| mirror | X | X |
| pts | X |  |
| suse_manager_server | X | X | X |
| suse_manager_proxy | X | X |
| virthost | X |  |
| pxe_boot | X | X |

Each provider should implement the following modules:
- base
- host
- pxe_boot
- volume 

Provider supported:
- [x] Null
- [X] Libvirt

TODO:

- [x] Upate documentation
- [x] Update examples
- [x] Update project `sumaform-test-runner`

PR #658, opened 2019-12-13
Dummy backend

moio:
## What does this PR change?

It adds a new "dummy" backend that will not create any infrastructure, rather, it will configure some via SSH.


moio:
At the end of the day it was not particularly difficult to extract this PR from #652. Code-wise it is not different, but it's easier to review small blocks.

moio:
I am not sure if you like this name!

rjmateus:
> I am not sure if you like this name!

Can you elaborate a bit more about the reason to choose this name? We didn't use ssh in this backend. 
What if we decide to add another backend that connects to existing machines using ssh, set some data (like grains), copy salt files and apply the high state? The name for a backend with this logic would probably be called `ssh`.

moio:
>  What if we decide to add another backend that connects to existing machines using ssh, set some data (like grains), copy salt files and apply the high state? The name for a backend with this logic would probably be called `ssh`.

Hmm, yeah, there must be something rather big I've missed here. I thought the intention was to use this module to do exactly that - I thought this was in the making, and we would add `provisioner` blocks to the `null_resource`s as an additional step.

But possibly I am missing something - what would be the tradeoff in keeping both such an "ssh backend" and a really empty "null backend"?

rjmateus:
> > What if we decide to add another backend that connects to existing machines using ssh, set some data (like grains), copy salt files and apply the high state? The name for a backend with this logic would probably be called `ssh`.
> 
> Hmm, yeah, there must be something rather big I've missed here. I thought the intention was to use this module to do exactly that - I thought this was in the making, and we would add `provisioner` blocks to the `null_resource`s as an additional step.
> 
> But possibly I am missing something - what would be the tradeoff in keeping both such an "ssh backend" and a really empty "null backend"?

We can do it, but I was thinking in the module as a really empty "null backend", to allow us to verify the configurations we have in our root module before connecting to machines.
If you think this is pointless I can update this module with the needed `provisioner` and `connection` blocks and also add the provisioner specific properties, like `ip_address` and `port`.
What do you think?

moio:
As per our discussion, we will want something separate from the SSH backend, which will also come (just later).

This could be used for Travis validations, for example.

As far as the name goes, I could suggest "dummy", which is hopefully clear to users and developers alike. I would not document it in main user READMEs.

Another possible name would be "stub", but that's probably clear only to programmers.

I will leave the final choice to you in any case.

PR #660, opened 2019-12-16
Mirror: add repos needed by testsuite

rjmateus:

## What does this PR change?

Add repositories need by the test suite to minima configuration file.

### DIFF
--- a/salt/mirror/minima.yaml
+++ b/salt/mirror/minima.yaml
@@ -6,6 +6,17 @@ scc:
   username: {{ grains.get("cc_username") }}
   password: {{ grains.get("cc_password") }}
   repo_names:
+    # required by testsuite
+    - SLES11-SP1-Pool
+    - SLES11-SP1-Updates
+    - SLES11-SP1-SUSE-Manager-Tools
+    - SLE11-SDK-SP1-Pool
+    - SLE11-SDK-SP1-Updates
+    - SUSE-Manager-Proxy-2.1-Pool
+    - SUSE-Manager-Proxy-2.1-Updates
+    - SLES11-SP3-Pool
+    - SLES11-SP3-Updates
+    - SLE-15-GA-Desktop-NVIDIA-Driver
     # SLES
     - SLES11-SP4-Pool
     - SLES11-SP4-Updates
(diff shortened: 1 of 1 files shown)


moio:
This works and I am OK merging it, although it would be even better to change the testsuite so that it does not require very old products and would save mirrors some bandwidth.

This can also be done as a second step.

rjmateus:
> This works and I am OK merging it, although it would be even better to change the testsuite so that it does not require very old products and would save mirrors some bandwidth.

I totally agree with you. But @Bischoff made a valid point that with this test we check if SCC works correctly with older product versions. For that reason, I didn't change the automatic test.

Bischoff:
> This can also be done as a second step.

I would prefer not, for the test coverage reasons explained by Ricardo.

Also, please consider there is no cost associated, other than maintaining more complete lists in sumaform: we don't sync those products.

moio:
The cost is just minima space and time to do the sync of several extra products. For remote devs or anyone needing a minima server for an on-the-road setup, that could be a couple hours of sync to their mirrors.

Bischoff:
> The cost is just minima space and time to do the sync of several extra products. For remote devs or anyone needing a minima server for an on-the-road setup, that could be a couple hours of sync to their mirrors.

Hmmm, right. I was focusing on the situation on the server, and overlook the consequences on the mirror.

Devil's advocate: this resync on the mirror is a one-off.

Also, I don't like much twisting the testing logic or scope for infrastructure reasons. We're just doing this too much already, and for example we keep paying over and over again for not doing real resyncs. Latest example: SUSE/spacewalk#10360.

But okay, feel free to reduce the list and to change the test suite accordingly. I don't want @rjmateus to have his remote work hindered by my being intransigent on testing old stuff.

rjmateus:
> But okay, feel free to reduce the list and to change the test suite accordingly. I don't want @rjmateus to have his remote work hindered by my being intransigent on testing old stuff.

Hey @Bischoff ,
My remote work is not hindered by this, I work without mirror (just take more time to download stuff).
I'm testing how mirror works and make sure everyone that wants to use it is able to (even run the testsuite).
With these changes, which I'm testing again, I think they will be able.

I will add to my list the task of changing the cucumber test, and after that remove the unnecessary repositories from the mirror.

Bischoff:
> Hey @Bischoff ,
> My remote work is not hindered by this, I work without mirror (just take more time to download stuff).
> I'm testing how mirror works and make sure everyone that wants to use it is able to (even run the testsuite).
> With these changes, which I'm testing again, I think they will be able.
> 
> I will add to my list the task of changing the cucumber test, and after that remove the unnecessary repositories from the mirror.

We have a plan @rjmateus :smile_cat: .

PR #661, opened 2019-12-18
All clients optional

srbarrios:
## What does this PR change?

That PR make client and minion optional when we set up a main.tf that includes a controller.

Related PRs:
https://github.com/SUSE/spacewalk/pull/10341

### DIFF
--- a/README_TESTING.md
+++ b/README_TESTING.md
@@ -80,6 +80,12 @@ host_settings = {
   pxy = {
     present = true
   }
+  cli-sles12sp4 = {
+    present = true
+  }
+  min-sles12sp4 = {
+    present = true
+  }
   minssh-sles12sp4 = {
     present = true
   }
@@ -98,7 +104,20 @@ host_settings = {
 }
 ```
 
-In addition to the `present` flag, each of the hosts (including `srv`, `ctl`, `cli-sles12sp4` and `min-sles12sp4` which are always present) accepts the following parameters:
+The default value for host_settings block has a SLES12SP4 Traditional Client and SLES12SP4 Minion present:
+
+```hcl
+host_settings = {
+  cli-sles12sp4 = {
+    present = true
+  }
+  min-sles12sp4 = {
+    present = true
+  }
+}
+```
+
+In addition to the `present` flag, each of the hosts (including `srv` and `ctl` which are always present) accepts the following parameters:
  - `mac`: to use a fixed MAC address
  - `additional_repos` to add software repositories (see [README_ADVANCED.md](README_ADVANCED.md))
  - `image` to use a different base image
--- a/modules/libvirt/controller/variables.tf
+++ b/modules/libvirt/controller/variables.tf
@@ -40,18 +40,18 @@ variable "proxy_configuration" {
 
 variable "client_configuration" {
   description = "use module.<CLIENT_NAME>.configuration, see main.tf.libvirt-testsuite.example"
-  type = object({
-    ids       = list(string)
-    hostnames = list(string)
-  })
+  default = {
+    ids       = []
+    hostnames = []
+  }
 }
 
 variable "minion_configuration" {
   description = "use module.<MINION_NAME>.configuration, see main.tf.libvirt-testsuite.example"
-  type = object({
-    ids       = list(string)
-    hostnames = list(string)
-  })
+  default = {
+    ids       = []
+    hostnames = []
+  }
 }
 
 variable "sshminion_configuration" {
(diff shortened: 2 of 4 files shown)


modules/libvirt/controller/variables.tf
```diff
@@ -44,6 +44,10 @@ variable "client_configuration" {
     ids       = list(string)
     hostnames = list(string)
   })
+  default = {
```
rjmateus: I think we can have the variable defined as the other optional systems: Without type, and default value only with `hostnames`

modules/libvirt/controller/variables.tf
```diff
@@ -52,6 +56,10 @@ variable "minion_configuration" {
...
```
rjmateus: I think we can have the variable defined as the other optional systems: Without type, and default value only with `hostnames`

README_TESTING.md
```diff
@@ -98,7 +98,7 @@ host_settings = {
 }
 ```
 
-In addition to the `present` flag, each of the hosts (including `srv`, `ctl`, `cli-sles12sp4` and `min-sles12sp4` which are always present) accepts the following parameters:
```
rjmateus: Since they are optional now, the HCL example above should be updated to contain those clients.

modules/libvirt/controller/variables.tf
srbarrios: I'm fine with it, but first, we will need to review that we are not assuming that type somewhere.
I will check it tomorrow.

README_TESTING.md
srbarrios: Uhmm but then we will need to include in that example all the optional clients? We have a lot more optional clients :)

srbarrios: Actually, that made me think if the cucumber_testsuite module shouldn't include all of them too :thinking:

modules/libvirt/controller/variables.tf
srbarrios: Oki, looks safe, I will remove the type

srbarrios:
> An additional angle: even if all clients are now optional, should they all be disabled by default?
> 
> Would there be trade-offs to have at least one or two enabled by default, for the sake of end (human) user convenience in most cases?

Fine for me, but where that's configured? :thinking: 
My only concern was to be able to don't have a sle client or sle minion in our main.tf, so we can select exactly which clients we need for each test (thinking on PR tests)

README_TESTING.md
rjmateus: Cucumber_testsuite only included a sub-set of the clients supported by the controller.
We set up this special module to optimize the run of cucumber testsuite, which (if I'm not wrong) only uses that sub-set of clients. If more will be needed, they can be added in the future, or in case of need creat a new util module.
Since you are changing these clients to optional, we need to update documentation according :)

rjmateus:
Maybe we can change the default variable for host, to include sle minion and sle client: https://github.com/uyuni-project/sumaform/blob/master/modules/cucumber_testsuite/variables.tf#L54-L55

If end-user defines the host he wants to use, then he should specific all. But if no value are defined he gets one client and one minon.

srbarrios:
> Maybe we can change the default variable for host, to include sle minion and sle client: https://github.com/uyuni-project/sumaform/blob/master/modules/cucumber_testsuite/variables.tf#L54-L55
> 
> If end-user defines the host he wants to use, then he should specific all. But if no value are defined he gets one client and one minon.

Ah ok, I got it. So, yeas, fine for me to add that change.

README_TESTING.md
srbarrios: Ok, so just let it as it is for now. on QAM, where we have the rest of clients, we are not using that module.

srbarrios:
@moio @rjmateus  Can I suggest that someone address the default host_settings in another PR?
I think the scope of that one was clear, making all clients optional.

rjmateus:
@srbarrios One important thing. After merge this PR, all main.tf files for cucumber testsuite used by CI needs to be updated, or client and minion will not be created.

srbarrios:
> @srbarrios One important thing. After merge this PR, all main.tf files for cucumber testsuite used by CI needs to be updated, or client and minion will not be created.

I can't see what must be added here: https://gitlab.suse.de/galaxy/sumaform-test-runner/blob/master/32/main-full-NUE.tf#L32
It just looks fine no? As we are using host_settings to add the mac, we are already telling it that must create that vm, Am I wrong?

rjmateus:
> > @srbarrios One important thing. After merge this PR, all main.tf files for cucumber testsuite used by CI needs to be updated, or client and minion will not be created.
> 
> @rjmateus I can't see what must be added here: https://gitlab.suse.de/galaxy/sumaform-test-runner/blob/master/32/main-full-NUE.tf#L32
> It just looks fine no? As we are using host_settings to add the mac, we are already telling it that must create that vm, Am I wrong?

You are correct. Since we are adding the `mac` field no changes are need.

PR #665, opened 2019-12-24
Missing glue in the new test suite module

Bischoff:
## What does this PR change?

Bridge code for new variables in new wrapper

### DIFF
--- a/modules/cucumber_testsuite/variables.tf
+++ b/modules/cucumber_testsuite/variables.tf
@@ -94,8 +94,28 @@ variable "git_repo" {
   default     = "default"
 }
 
+variable "git_profiles_repo" {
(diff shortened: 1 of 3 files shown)


modules/cucumber_testsuite/variables.tf
```diff
@@ -94,8 +94,28 @@ variable "git_repo" {
...
+  default     = "default"
+}
+
+variable "portus_uri" {
+  description = "URI of portus server, see README_ADVANCED.md"
+  default     = null
+}
+
+variable "portus_username" {
+  description = "username on portus server, see README_ADVANCED.md"
```
moio: Shouldn't this be in README_TESTSUITE.md instead?

Bischoff: You're right, I'll fix this (piggyback on another PR)

PR #669, opened 2020-01-01
New PXEBOOT_IMAGE variable

Bischoff:
## What does this PR change?

This PR:
* renames `PXEBOOTMAC` test suite variable to `PXEBOOT_MAC`
   (that's the way to identify the PXE boot minion as it does not have an IP in main network)
* introduces a new variable `PXEBOOT_IMAGE` that contains the operating system
   we want to install on it via PXE
   (was previously hardcoded)
* determines the latter automatically as identical to the starting operating system
   on that minion, so if eg we have SLE 15 SP1 on the PXE boot minion, we will reinstall
   a SLE 15 SP1 system via PXE.
   (doing so ensures a certain level of idempotency of tests)

## Links

Testsuite: uyuni-project/uyuni#1762


moio:
LGTM, thanks

Bischoff:
@moio piggybacking this PR to point to correct READMEs (in a separate commit)

moio:
Good, please "Rebase and merge" if it's good to you, so that commits stay "ungrouped" in the git history

Bischoff:
I will. But I'm still waiting for all tests to prove good, so it might take a few more hours.

PR #670, opened 2020-01-07
WIP: Allow to change image for minions

Bischoff:
## What does this PR change?

This PR enables to change the images for traditional client, normal minion, and other minions.

Example of use:
```
  host_settings = {
    (...)
    min-sles12sp4 = {
      mac = "AA:B2:93:00:00:42"
      image = "sles15sp1"
    }
```

@rjmateus : the names of the clients are not universal enough. We should rename (in a separate PR) :
```
cli-sles12sp4    -->  client
min-sles12sp4    -->  minion
minssh-sles12sp4 -->  sshminion
min-centos7      -->  centos
min-ubuntu1804   -->  ubuntu
```
or something alike.

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -25,13 +25,16 @@ module "base" {
 locals {
   server_full_name = "${var.name_prefix}srv.${var.domain}"
   hosts            = keys(var.host_settings)
-  macs = { for host_key in local.hosts :
-  host_key => lookup(var.host_settings[host_key], "mac", null) if var.host_settings[host_key] != null }
+  macs             = { for host_key in local.hosts :
+    host_key => lookup(var.host_settings[host_key], "mac", null) if var.host_settings[host_key] != null }
   additional_repos = { for host_key in local.hosts :
-  host_key => lookup(var.host_settings[host_key], "additional_repos", {}) if var.host_settings[host_key] != null }
-  images = { for host_key in local.hosts :
+    host_key => lookup(var.host_settings[host_key], "additional_repos", {}) if var.host_settings[host_key] != null }
+  images           = { for host_key in local.hosts :
     host_key => lookup(var.host_settings[host_key], "image", "default")
-  if var.host_settings[host_key] != null ? contains(keys(var.host_settings[host_key]), "image") : false }
+    if var.host_settings[host_key] != null ? contains(keys(var.host_settings[host_key]), "image") : false }
+  names            = { for host_key in local.hosts :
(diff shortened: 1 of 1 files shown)


moio:
Is this what we actually want looking forward in the testsuite, or would it be more appropriate to have many instances of minions and clients across the various OSs, and make them all optional?

(this is more of a testsuite design question and less of a sumaform question so perhaps this is not the best place to ask and definitely to discuss, still, I guess we need to set some direction as part of this PR)

Bischoff:
> Is this what we actually want looking forward in the testsuite, or would it be more appropriate to have many instances of minions and clients across the various OSs, and make them all optional?
> 
> (this is more of a testsuite design question and less of a sumaform question so perhaps this is not the best place to ask and definitely to discuss, still, I guess we need to set some direction as part of this PR)

@srbarrios would probably answer this better, but let me try:

Currently, the main part of sumaform is pretty agnostic about what VMs we have and how many.
Unfortunately, this does not extend to the new sumaform test suite "wrapper" module. This new module hardcodes two things:
1. the list of VMs we use in the test suite (server, proxy, client, minion, ssh minion, centos minion...)
2. their versions, as a. the name of the sumaform variable b. the name of their libvirt VM c. the base image

This PR addresses only 2. c., I suggested above to address also 2. a. in some separate PR, and we might probably need 2. b. for comfort as well. @moio, now you are discussing 1.

(Please notice this current PR does not really hinder nor favour 1. Still, as you say, this is a valid debate)

Most of our test suites have a fixed list of minions. All of them are optional now (even the normal minion), the only fixed VM is now the server. But this is a choice between 0 and 1, not in 0 to _n_. This is compatible with current harcoding of list of clients in sumaform. So no huge issue.

There is however an exception. The new QAM module developed by @srbarrios will use many clients of every kind. Since this is the exception rather than the rule, maybe it can not use the new test suite wrapper, and use "bare" sumaform. Oscar, I suppose this is the way you are deploying VMs currently?

My guts feeling is : You Are not Gonna Need It, but I might be wrong.

Bischoff:
@rjmateus I did not understand you had a different strategy in mind: create as many declarations as there are different OSes.

I took a different approach: parametrize the minions. I think it's more coherent with parameters like `image =` that we have already for server controler etc. But this can be discussed.

Bischoff:
After discussing with @moio and @rjmateus:
 * Ricardo will refactor this part
 * Ricardo will take the occasion to enable to have several minions of a given type
 * I will keep this (working) code in a branch of sumaform, so that Ricardo does not have a big rebase
 * this branch will enable me to work on DNS and DHCP part of SLE 15 clients

Closing for now

PR #673, opened 2020-01-23
ssh backend support (provision)

rjmateus:

## What does this PR change?

Add ssh backend support for provisioning existing servers

https://github.com/SUSE/spacewalk/issues/10489

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ In `sumaform` you define a set of virtual machines in a `main.tf` configuration 
 To choose the backend in use one should create a symbolic link to a `backend_module` module. Refer to specific READMEs to get started:
  * [libvirt README](backend_modules/libvirt/README.md)
  * [AWS README](backend_modules/aws/README.md)
+ * [SSH README](backend_modules/ssh/README.md)
  * [NULL README](backend_modules/null/README.md)
 
 ## Typical use
--- a/backend_modules/ssh/README.md
+++ b/backend_modules/ssh/README.md
@@ -1,0 +1,91 @@
+# SSH backend
+
+This backend, differently from others, assumes hosts already exist and can be accessed via SSH, thus configuring them for the desired roles (server, proxy, minion, etc.). You can use this backend to provision physical machines or in any situation where others are not applicable.
+
+## First-time configuration
+
+ - Copy `main.tf.ssh.example` to `main.tf`
+ - Complete the missing information
+ - Create a symbolic link to the `ssh` backend module directory inside the `modules` directory: `ln -sfn ../backend_modules/ssh modules/backend`
+
+## Requirements
+
+ - SSH access via password or private key to a user with root privileges
+    - Access through a [bastion host](https://en.wikipedia.org/wiki/Bastion_host) is also possible using password or private key
+ - Machine must come with `salt` pre-installed
+ - If additional_network is needed one should make sure it is set up correctly and have the IP mask for it (this backend will not create it)
+
+### Base module
+
+Available provider settings for the base module:
+
+| Variable name       | Type    | Default value | Description                                                                                                                                                                                   |
+|---------------------|---------|---------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| private_key         | string  | null          | The contents of an SSH key to use. These can be loaded from a file on disk using the file function. This takes preference over the password if provided                                       |
+| certificate         | string  | null          | The contents of a signed CA Certificate. The certificate argument must be used in conjunction with a private_key. These can be loaded from a file on disk using the the file function         |
+| host_key            | string  | null          | The public key from the remote host or the signing CA, used to verify the connection                                                                                                          |
+| bastion_host        | string  | null          | Bastion hostname                                                                                                                                                                              |
+| bastion_host_key    | string  | null          | The public key for the bastion host                                                                                                                                                           |
+| bastion_port        | string  | null          | The port for the bastion host                                                                                                                                                                 |
+| bastion_user        | string  | null          | The user for the bastion host                                                                                                                                                                 |
+| bastion_password    | string  | null          | The password for the bastion host                                                                                                                                                             |
+| bastion_private_key | string  | null          | The contents of an SSH key file to use for the bastion host. These can be loaded from a file on disk using the file function                                                                  |
+| bastion_certificate | string  | null          | The contents of a signed CA Certificate. The certificate argument must be used in conjunction with a bastion_private_key. These can be loaded from a file on disk using the the file function |
+| timeout             | string  | "20s"         | The timeout to wait for the connection to become available. Should be a value like 30s or 5m                                                                                                  |
+| additional_network  | string  | null          | A network mask for additional network                                                                                                                                                         |
+
+### Host modules
+
+Settings not set in host module will fall back to the value defined in base module it it exists there.
+
+Following settings apply to all modules, such as `server`, `proxy`, `client`, `grafana`, `minion`, `mirror`, `sshminion` and `virthost`:
+
+| Variable name       | Type    | Default value     | Description                                                                                                                                                                                   |
+|---------------------|---------|-------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| host                | string  | null              | The address of the host to connect to                                                                                                                                                         |
+| user                | string  | "root"            | The username for the connection                                                                                                                                                                   |
+| password            | string  | "linux"           | The password for the connection                                                                                                                                                               |
+| port                | number  | 22                | The port to connect to                                                                                                                                                                        |
+| type                | string  | "ssh"             | The connection type that should be used. Valid types are ssh and winrm                                                                                                                        |
+| private_key         | string  | base module value | The contents of an SSH key to use. These can be loaded from a file on disk using the file function. This takes preference over the password if provided                                       |
+| certificate         | string  | base module value | The contents of a signed CA Certificate. The certificate argument must be used in conjunction with a private_key. These can be loaded from a file on disk using the the file function         |
+| host_key            | string  | base module value | The public key from the remote host or the signing CA, used to verify the connection                                                                                                          |
+| bastion_host        | string  | base module value | Bastion hostname                                                                                                                                                                              |
+| bastion_host_key    | string  | base module value | The public key for the bastion host                                                                                                                                                           |
+| bastion_port        | string  | base module value | The port for the bastion host                                                                                                                                                                 |
+| bastion_user        | string  | base module value | The user for the bastion host                                                                                                                                                                 |
+| bastion_password    | string  | base module value | The password for the bastion host                                                                                                                                                             |
+| bastion_private_key | string  | base module value | The contents of an SSH key file to use for the bastion host. These can be loaded from a file on disk using the file function                                                                  |
+| bastion_certificate | string  | base module value | The contents of a signed CA Certificate. The certificate argument must be used in conjunction with a bastion_private_key. These can be loaded from a file on disk using the the file function |
+| timeout             | string  | base module value | The timeout to wait for the connection to become available. Should be a value like 30s or 5m                                                                                                  |
+
+An example follows:
+```hcl-terraform
+ ...
+ provider_settings = {
+   host        = "192.168.1.1"
+ }
+ ...
+```
+
+## Dependency between modules
+
+Clients an minions can only be set-up after the server provisioning. To ensure this behavior the output from server provision should be used as input on clients and minions.
+
+Alternatively, one can provision the server beforehand and on a second step provision clients and minions.
+
+An example with module dependency follows:
+
+```hcl-terraform
+
+module "server" {
+  source = "./modules/server"
+  ...
+}
+
+module "client" {
+  source = "./modules/client"
+  server_configuration = module.server.configuration
+  ...
+}
+```
--- a/backend_modules/ssh/host/main.tf
+++ b/backend_modules/ssh/host/main.tf
@@ -63,7 +63,7 @@ resource "null_resource" "provisioning" {
 
   provisioner "file" {
     source      = "salt"
-    destination = "/root"
+    destination = "/opt"
   }
 
   provisioner "file" {
@@ -100,7 +100,7 @@ resource "null_resource" "provisioning" {
 
   provisioner "remote-exec" {
     inline = [
-      "sh /root/salt/first_deployment_highstate.sh",
+      "sudo bash /opt/salt/first_deployment_highstate.sh"
     ]
   }
 }
--- a/backend_modules/ssh/host/main.tf
+++ b/backend_modules/ssh/host/main.tf
@@ -1,0 +1,115 @@
+locals {
+  resource_name_prefix = "${var.base_configuration["name_prefix"]}${var.name}"
+
+  provider_settings = merge({
+    host     = null
+    user     = "root"
+    password = "linux"
+    port     = 22
+    type     = "ssh"
+    // only supportted in ssh connection
+    private_key = null
+    certificate = null
+    host_key    = null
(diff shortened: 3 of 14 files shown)


README.md
```diff
@@ -58,6 +58,7 @@ In `sumaform` you define a set of virtual machines in a `main.tf` configuration
 Refer to specific READMEs to get started:
  * [libvirt README](backend_modules/libvirt/README.md)
  * [AWS README](backend_modules/aws/README.md)
+ * [SSH README](backend_modules/ssh\/README.md)
```
moio:
```suggestion
 * [SSH README](backend_modules/ssh/README.md)
```

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
+# SSH backend
+
+This backend can be set to provisioning existing servers with the desired role (server, proxy, minion, etc...)
```
moio:
```suggestion
This backend, differently from others, assumes hosts already exist and can be accessed via SSH, thus configuring them for the desired roles (server, proxy, minion, etc.). You can use this backend to provision physical machines or in any situation where others are not applicable.
```

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
...
+ - SSH access with root user
+ - salt packages installed
+
+### Base module
+
+Available provider settings for the base module:
+
+| Variable name       | Type    | Default value | Description                                                                                                                                                                                   |
+|---------------------|---------|---------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| bastion_host        | string  | null          | Bastion hostname                                                                                                                                                                              |
```
moio: Would it make sense to assume one sets a private key in `base` and then all hosts use it, unless overridden?

It seems cumbersome to have to specify details for all individual hosts in advance - I would guess in most setups they do not really need to be all different.

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## First-time configuration
+
+ - Copy `main.tf.ssh.example` to `main.tf`
+ - Complete the missing information
+ - Create a symbolic link to the `ssh` backend module directory inside the `modules` directory: `ln -sfn ../backend_modules/ssh modules/backend`
+
+## Machine requirements
+
+ - SSH access with root user
```
moio:
```suggestion
 - SSH access via password or private key to a user with root privileges
```

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
...
+ - salt packages installed
```
moio:
```suggestion
 - machine must come with `salt` preinstalled
```

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+This backend can be set to provisioning existing servers with the desired role (server, proxy, minion, etc...)
+
+## First-time configuration
+
+ - Copy `main.tf.ssh.example` to `main.tf`
+ - Complete the missing information
+ - Create a symbolic link to the `ssh` backend module directory inside the `modules` directory: `ln -sfn ../backend_modules/ssh modules/backend`
+
+## Machine requirements
```
moio:
```suggestion
## Requirements
```

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+ - SSH access with root user
+ - salt packages installed
+
```
moio: How do we handle/what are the requirements for particular roles with additional disks or networks?

backend_modules/ssh/README.md
moio: Also mention access via bastion host is possible

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
...
+| bastion_password    | string  | null          | The password for the bastion host                                                                                                                                                             |
+| bastion_private_key | string  | null          | The contents of an SSH key file to use for the bastion host. These can be loaded from a file on disk using the file function                                                                  |
+| bastion_certificate | string  | null          | The contents of a signed CA Certificate. The certificate argument must be used in conjunction with a bastion_private_key. These can be loaded from a file on disk using the the file function |
+| timeout             | string  | "20s"         | The timeout to wait for the connection to become available. Should be a value like 30s or 5m                                                                                                  |
+
+### Host modules
+
+Following settings apply to all modules, such as `suse_manager`, `suse_manager_proxy`, `client`, `grafana`, `minion`, `mirror`, `sshminion` and `virthost`:
```
moio:
```suggestion
Following settings apply to all modules, such as `server`, `proxy`, `client`, `grafana`, `minion`, `mirror`, `sshminion` and `virthost`:
```

Please consider renaming in other READMEs as well (maybe `grep` for any remaining instances). Feel free to commit directly to master.

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+ | Variable name       | Type    | Default value | Description                                                                                                                                                                                   |
+ |---------------------|---------|---------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+ | host                | string  | null          | The address of the host to connect to                                                                                                                                                         |
```
moio: What happens if this is left unspecified?

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,75 @@
...
+ ...
+ provider_settings = {
+   host        = "192.168.1.1"
+ }
+ ...
+```
+
+## Dependency between modules
+
+Clients an minions can only be set-up after the server provisioning. To ensure this behavior the output from server provision should be used as input on clients and minions.
```
moio: This should be enforced by the backend. Can we put "something" in the host's output to emulate what happens in other backends, like the id of a `null_resource`?

backend_modules/ssh/host/main.tf
```diff
@@ -0,0 +1,113 @@
...
+
+  count = var.provision ? 1 : 0
+
+  connection {
+    host     = local.provider_settings["host"]
+    user     = local.provider_settings["user"]
+    password = local.provider_settings["password"]
+    port     = local.provider_settings["port"]
+    type     = local.provider_settings["type"]
+    // only supportted in ssh connection
```
moio:
```suggestion
    // only supported in ssh connection
```

backend_modules/ssh/host/main.tf
```diff
@@ -0,0 +1,113 @@
...
+    private_key = local.provider_settings["private_key"]
+    certificate = local.provider_settings["certificate"]
+    host_key    = local.provider_settings["host_key"]
+    // ssh connection throught bastion host
```
moio:
```suggestion
    // ssh connection through a bastion host
```

backend_modules/ssh/host/main.tf
```diff
@@ -0,0 +1,113 @@
...
+    bastion_password    = var.base_configuration["bastion_password"]
+    bastion_private_key = var.base_configuration["bastion_private_key"]
+    bastion_certificate = var.base_configuration["bastion_certificate"]
+
+    timeout = var.base_configuration["timeout"]
+  }
+
+  provisioner "file" {
+    source      = "salt"
+    destination = "/root"
```
moio: Will this work for root-privileged users who are not literally `root`? Maybe we should opt for something in `/tmp` or other "neutral" directory

backend_modules/ssh/volume/main.tf
```diff
@@ -0,0 +1,6 @@
+// in ssh backend we don't provide any spupport for disk creation
```
moio:
```suggestion
// in ssh backend we don't provide any support for disk creation
```

main.tf.ssh.example
```diff
@@ -0,0 +1,59 @@
...
+  source = "./modules/client"
+  base_configuration = module.base.configuration
+
+  name = "client"
+  image = "sles12sp3"
+  server_configuration = module.server.configuration
+  // see modules/client/variables.tf for possible values
+
+  provider_settings = {
+    hostname = "192.168.1.2"
```
moio: Maybe we should use a different name for this variable like `address`, since it is not really a hostname (more exactly: it could also not be a hostname).

modules/cucumber_testsuite/main.tf
```diff
@@ -17,23 +17,24 @@ module "base" {
 }
 
 locals {
-  server_full_name = "${var.name_prefix}srv.${var.domain}"
-  hosts            = keys(var.host_settings)
+  server_full_name          = "${var.name_prefix}srv.${var.domain}"
+  hosts                     = keys(var.host_settings)
   provider_settings_by_host = { for host_key in local.hosts :
-  host_key => lookup(var.host_settings[host_key], "provider_settings", {}) if var.host_settings[host_key] != null }
```
moio: Changes in this file seem unrelated

salt/mirror/minima.yaml
```diff
@@ -6,6 +6,17 @@ scc:
   username: {{ grains.get("cc_username") }}
   password: {{ grains.get("cc_password") }}
   repo_names:
+    # required by testsuite
+    - SLES11-SP1-Pool
+    - SLES11-SP1-Updates
```
moio: Changes in this file seem unrelated

backend_modules/ssh/README.md
rjmateus: OK makes sense. What do you think if I change it to use the value defined at host level, and if it's not defined, fall back to the value defined at base module? With the lookup method it should be simple to implement.
This way we will have more flexibility for all possible cases.

backend_modules/ssh/README.md
rjmateus: At this point, I'm assuming it is already configured on the machine. I think the way we configure, for example, a new network interface to a machine would depend on the backend.
But I will add the "additional_network" information to grains, as we have in libvirt. Makes sense?

backend_modules/ssh/README.md
rjmateus: We will get a connection timeout error

modules/cucumber_testsuite/main.tf
rjmateus: PR was merging to an older branch. This file is not changed, please ignore it.

salt/mirror/minima.yaml
rjmateus: PR was merging to an older branch. This file is not changed, please ignore it.

backend_modules/ssh/README.md
rjmateus: I documented it here, but It could happen in other backends if we defined the static values in server_configuration.
When we define the modules in the main module we need to specify the dependency between them, as minion depends on server output.

main.tf.ssh.example
rjmateus: I have changed it to "host", since is the name also used by terraform connection. Sounds OK?

backend_modules/ssh/host/main.tf
rjmateus: I have tested using a different user with root-privileged and it didn't work.
I also try to change the folder to "/tmp" and realize we have that path hardcode in several salt files.
I totally agree we should change this, but I this it should be in a different PR. If you agree I will document it on readme, saying it needs to be the "root" user, and create an issue to address this problem. What do you think?

rjmateus: I have tried again with the current codebase and a user with root-privilege, and after analyzing the error it looks like it is unable run zypper ref, getting a certificate error.
After finishing all the changes I will try to get a workaround for it.

backend_modules/ssh/README.md
moio: Makes total sense

backend_modules/ssh/README.md
moio: Well, from a user POV I expect to be able to configure the network interface device name, which will depend on the provisioning mechanism. So I would expect it to be part of `provider_settings`to be passed via grains to Salt.

From a documentation POV I definitely expect a note here :wink:

backend_modules/ssh/README.md
moio: I guess we do not really have a way to make the mistake more evident, do we?

main.tf.ssh.example
moio: ok

backend_modules/ssh/host/main.tf
moio: Sounds like a different problem. What is the error trace exactly?

backend_modules/ssh/README.md
rjmateus: I don't think so

backend_modules/ssh/README.md
rjmateus: I fact the property adicional_network was already available on the base module (when I wrote the comment I didn't realize that) and is propagated to the grains file.
I documented it now, as a possible option in the bae module. Can you have a look if it's what you were expecting?

backend_modules/ssh/host/main.tf
rjmateus: I'm getting several errors like this one: http://pastebin.suse.de/25323
And then, since we cannot connect to the repository, the provision fails.

moio: This does seem unrelated. Let's have a chat today

rjmateus: @moio I have followed the same approach the Cloud team had in the AWS refactoring and copy all files to a temporary folder, then move to the final location and finally call the highstate with `sudo`.
I test it and worked in my local environment. I add the changes in a separate commit for easier review.

backend_modules/ssh/README.md
```diff
@@ -0,0 +1,90 @@
...
+## First-time configuration
+
+ - Copy `main.tf.ssh.example` to `main.tf`
+ - Complete the missing information
+ - Create a symbolic link to the `ssh` backend module directory inside the `modules` directory: `ln -sfn ../backend_modules/ssh modules/backend`
+
+## Requirements
+
+ - SSH access via password or private key to a user with root privileges
+    - Access through bastian is also possible using password or private key
```
moio:
```suggestion
    - Access through a [bastion host](https://en.wikipedia.org/wiki/Bastion_host) is also possible using password or private key
```

backend_modules/ssh/README.md
moio: It's good to have it documented, what I meant here is that it should be mentioned as a (physical or virtual) hardware requirement. As in: if you want to use `additional_network`, you will need to make sure it is set up correctly and that you have the IP mask for it. This plugin will not create it for you!

backend_modules/ssh/host/main.tf
rjmateus: I have changed the way provision is done to move salt folder from "/root" to "/opt" and in this way is not dependent from root user. I have tested it creating and provisioning a server and a minion.

salt/first_deployment_highstate.sh
```diff
@@ -3,7 +3,9 @@
 # Applies the "minimal" state in isolation to set the hostname and update Salt itself
 # then applies the highstate
 
-salt-call --local --file-root=/root/salt/ --log-level=quiet --output=quiet state.sls default.minimal ||:
-salt-call --local --file-root=/root/salt/ --log-level=info --retcode-passthrough --force-color state.highstate || exit 1
+FILE_ROOT="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
```
moio: Sorry, I am not good at bash, what is the purpose of this variable?

rjmateus: We need to pass the salt files root folder in the option "--file-root". With this variable we dynamically calculate is location based on the folder the script is also located.
My goal here is keep compatability with existing configuration for libvirt, for example, which is still copying the files to '/root/salt', so I cannot just change the folder for a new location.

PR #675, opened 2020-01-30
Don't always abort at first failed stage in run-testsuite

Bischoff:
## What does this PR change?

After 1.6 zillion removals of `set -e` in `run-testsuite` script, I finally decided to let the hard errors be optional.

related: uyuni-project/uyuni#1845


salt/controller/run-testsuite
```diff
@@ -0,0 +1,63 @@
+#! /bin/bash
+
+cd /root/spacewalk/testsuite
+
+ABORT_ON_ERROR=1
+if [ "$1" = "-d" -o "$1" = "--dontfail" ]; then
```
calancha: With a dash between the two words reads nicer:
`--dont-fail`

salt/controller/run-testsuite
```diff
@@ -0,0 +1,63 @@
...
+if [ "$1" = "parallel" ]; then
+  stage core
+  stage reposync
+  parallel_stage init_clients
+  stage secondary
+  parallel_stage secondary_parallelizable
+  stage finishing
+fi
+
+# prepare reference host
```
calancha: It might be useful to  allow something like:
`if [ "$1" = "essentiall" ]; then
    stage core
    stage init_clients
    stage reposync
fi
`

srbarrios:
LGTM.
But I'm wondering if we are not overengineering it.
Could it work doing something like that?

```
if [ -z "$1" ]; then
   rake cucumber:core ||:
   rake cucumber:reposync ||:
   rake cucumber:init_clients ||:
   rake cucumber:secondary ||:
   rake cucumber:secondary_parallelizable ||:
   rake cucumber:finishing ||:
fi
```

salt/controller/run-testsuite
Bischoff: ok

salt/controller/run-testsuite
Bischoff: ok

Bischoff:
> LGTM.
> But I'm wondering if we are not overengineering it.
> Could it work doing something like that?
> 
> ```
> if [ -z "$1" ]; then
>    rake cucumber:core ||:
>    rake cucumber:reposync ||:
>    rake cucumber:init_clients ||:
>    rake cucumber:secondary ||:
>    rake cucumber:secondary_parallelizable ||:
>    rake cucumber:finishing ||:
> fi
> ```

No. The problem is that you need to be able to know if the test suite succeeded as a whole or not. Be it from jenkins or not.

calancha:
LGTM

PR #676, opened 2020-02-04
Support official images in libvirt

moio:
## What does this PR change?

It adds official images from upstream distros (JeOS for SLES and openSUSE, GenericCloud for CentOS, `cloudimg` for Ubuntu) and `cloud-init` support to pre-configure them.

**This PR also adds CentOS 6, CentOS 8, SLE 12 SP5 and Ubuntu 16.04 images.**

### Trade-offs
 - pro: custom images are not needed any more. Adding OSs such as RHEL, Oracle Linux and others is straightforward
 - pro: many public cloud providers have official OS images that are very similar to the ones in this PR. Reusing the `cloud-init` mechanism should be straightforward and removes the need to create custom images per-provider
 - pro: debugging `cloud-init` is sometimes dark magic just like debugging custom image creation, but at least the cycle time is much quicker (one boot instead of several-hundred-megabyte image creation and boot)
 - pro: it seems like most images are smaller in terms of download size (only counterexample at this point in time is CentOS 7)

image

 - pro: it seems like sumaform custom images were even overzealous in terms of minimal package selection, actually we had fewer than JeOS (which probably meant we were in unsupported scenarios)

image

 - con: deployment mechanism is more complicated, as we have one `cloud-init` step before the usual Salt steps

New "official" images come with an `o` suffix in the image name for now and both types of images are supported at this time.

### Peculiarities

 - some OSs are added by this PR and will only be supported in "official image" form, there is no current plan to create a custom image: **centos6o**, **centos8o**, **sles12sp5o**
 - other OSs just do not have an official image, so custom images will have to be used: **sles11sp4**, **sles12sp2**, **sles12sp3**
 - other OSs have an official image but they are long out of support and there is no currently-available Salt version from them. They will also continue in custom image form: **sles12**, **sles12sp1**
 - all official images at this point need Internet access for the `cloud-init` deployment step. This is a limitation that can be lifted with a future effort. Note that SLES images need to access the SUSE internal network
 
### Reviewer notes
 - best reviewed commit-by-commit
 - "Bugfix" commits are just general fixes that could be merged individually
 - "Enablement" commits are needed to support the new OSs
 - the meat of this PR is in the remaining commits, and especially `cloud-init` configuration. There is a lot of OS specifics in it, and unfortunately it can't really be otherwise

It would be impossible to test all of these images in all configurations. This PR is meant to get the process started, in particular, I made sure the `sshminion` module worked with each making sure that:
  - all Salt states applied green
  - it is possible to SSH into the machine via its Avahi name
  - the machine could SSH into the mirror via its Avahi name

Everything else at the moment is untested and I propose fixing any bug in follow-up PRs. I expect the migration in our testsuites and for general users to happen gradually, and eventually we could get rid of the `o` image suffixes.


moio:
@Bischoff, @srbarrios, @ktsamis: please take a look, I hope this satisfies your needs. I will be mostly off in the next days, but plan to collect and respond to any comments from you as always. To me this concludes the JeOS topic at this point in time - follow ups will be treated as normal sumaform maintenance (I am aware this is only the kick-off for your side of the work and have no plans to abandon you!).

@rjmateus: once this passes your and @Bischoff's review, plus any other person who might take a look, this can be merged and concludes the topic momentarily. Of course feel free to draw any inspiration for the AWS work.

Bischoff:
As already discussed many times, I can see another big "con", is that we might skip errors that customers get when using the "full" images. On the contrary we might get errors that are specific to those "reduced" images or to `cloud-init`.

On the "pro" side, I suppose that it will make downloading of qcow2s faster.

That being said, my main concern is that we have images for all OS+version combinations, and that get them within reasonable delays. If they are not exactly abiding by the book, that's not ideal, but I guess we can survive it.

After all, we've been testing only with CentOS in the testsuite since I joined QA, and that's not ideal either. Or we were testing manually with SLES ES images, and that was wrong too.

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -0,0 +1,197 @@
...
+  epel:
+    baseurl: http://download.fedoraproject.org/pub/epel/8/$basearch
+    mirrorlist: https://mirrors.fedoraproject.org/metalink?repo=epel-8&arch=$basearch
+    failovermethod: priority
+    enabled: true
+    gpgcheck: false
+    priority: 99
+    name: epel
+
+packages: ["salt-minion", "avahi", "nss-mdns", "qemu-guest-agent"]
```
Bischoff: We already get avahi packages from another SLS file, `salt/default/avahi.sls`. Does it make sense to mention it here too? Or is it some kind of chicken+egg problem?

salt/default/hostname.sls
```diff
@@ -44,12 +44,4 @@ hosts_file_hack:
     - args: "--no-ipv6 {{ grains['hostname'] }} {{ grains['domain'] }}"
```
Bischoff: I fail to see the change on this line...

backend_modules/libvirt/host/user_data.yaml
moio: It's a catch-22 indeed!

Reasoning goes like this:
 - Salt needs to install some packages
 - to install packages we need a repo
 - repo needs a URL
 - URL needs a host, which happens to be the mirror
 - mirror is reachable via avahi
 - one needs avahi to install packages
 - one needs avahi to install avahi

We break the chicken/egg cycle by having cloud-init install avahi, so that at Salt time mirror is reachable and can be used to install whatever (including, in principle, avahi).

It would make sense to remove the installation of Avahi from Salt files if all we supported was mirror setups, but sometimes the mirror might not be there (eg. ssh backend), hence no catch-22.

salt/default/hostname.sls
moio: There is no change on that line indeed. Are we looking at line 44?

backend_modules/libvirt/host/user_data.yaml
Bischoff: > It's a catch-22 indeed!

thanks :chicken: :egg:

salt/default/hostname.sls
Bischoff: forget it... error between chair and keyboard.

moio:
> As already discussed many times, I can see another big "con", is that we might skip errors that customers get when using the "full" images. On the contrary we might get errors that are specific to those "reduced" images or to `cloud-init`.

Thanks, I forgot to analyze this fact and updated the description accordingly.

rjmateus:
So far LGTM. I didn't see any red flag.
But I'm testing some images. 
sles15sp1o on server looks good.
Then I was testing opensuse150o on minion, and I keeping getting the errors: 
```
module.min-opensuse150o.module.minion.module.host.null_resource.provisioning[0] (remote-exec): [ERROR   ] Failed to configure repo 'os_pool_repo': Repository 'os_pool_repo' already exists as 'repo-oss'.
```
and
```
module.min-opensuse150o.module.minion.module.host.null_resource.provisioning[0] (remote-exec): [ERROR   ] Failed to configure repo 'os_update_repo': Repository 'os_update_repo' already exists as 'repo-update'.
```
I got the same error on AWS and end-up removing all pre configured repos from the image with cloud-init.

moio:
@rjmateus I did not get that error because I am running with mirror. This is a Salt bug and should be reported to Salt guys - created https://bugzilla.opensuse.org/show_bug.cgi?id=1166712, please have them informed.

In the mean time we should add a cloud-init command to clean up any existing repo.

backend_modules/libvirt/host/main.tf
```diff
@@ -21,15 +21,39 @@ locals {
     var.provider_settings,
     contains(var.roles, "virthost") ? { cpu_model = "host-model", xslt = file("${path.module}/sysinfos.xsl") } : {},
     contains(var.roles, "pxe_boot") ? { xslt = file("${path.module}/pxe.xsl") } : {})
+    cloud_init = length(regexall("o$", var.image)) > 0
```
Bischoff: OK. It makes it necessary to strictly respect name convention, but why not.

backend_modules/libvirt/host/network_config.yaml
```diff
@@ -0,0 +1,11 @@
+network:
+  version: 1
+  config:
+  - type: physical
+%{ if image == "ubuntu1804o" || image == "ubuntu1604o" }
```
Bischoff: only for official images, or do we want this for other ubuntu images as well?

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -0,0 +1,197 @@
+#cloud-config
+
+disable_root: false
+ssh_pwauth: true
+chpasswd:
+  expire: false
+  list: |
+     root:linux
+
+%{ if image == "centos6o" }
```
Bischoff: Same question: only official images? we don't want this for all images?

backend_modules/libvirt/host/main.tf
moio: We can also have a list of maps instead of `image_urls`, something like:

```hcl
images = [
  {name = "centos6o", cloud-init=true, url="https://cloud.centos.org/centos/6/images/CentOS-6-x86_64-GenericCloud.qcow2"}
]
```

At that point we could propagate the `cloud-init` bit down the line. It's cleaner but a bit more complicated. Would you prefer that?

backend_modules/libvirt/host/network_config.yaml
moio: For now `cloud-init` is exclusively for official images. I will amend the `DESIGN.md` file to explain better.

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -91,6 +91,9 @@ runcmd:
 
 %{ if image == "opensuse151o" }
 packages: ["qemu-guest-agent"]
+
+runcmd:
+  - zypper removerepo --all
```
rjmateus: In the test I made on leap 15.1 the problem doesn't happen. Both OS (Leap 15.0 and 15.1) have the same salt version: `salt-call 2019.2.0 (Fluorine)`

backend_modules/libvirt/host/main.tf
Bischoff: No strong feelings. It would be better, but honestly, I don't think it's worth the trouble...

backend_modules/libvirt/host/network_config.yaml
Bischoff: Got it.

PR #677, opened 2020-02-05
Fix some fixes when applying "repos" state

meaksh:
## What does this PR change?

This PR fixes two issues I've found on the `repos` state:

- Fix Uyuni Stable POOL repository url after 202001 release (the URL changed)
- Do not break "repos" state when system is "rhel/centos" and `product_version` is undefined.

### DIFF
--- a/salt/repos/server.sls
+++ b/salt/repos/server.sls
@@ -99,7 +99,7 @@ module_python2_update_repo:
 {% if 'uyuni-released' in grains['product_version'] %}
 server_pool_repo:
   pkgrepo.managed:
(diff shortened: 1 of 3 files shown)


salt/repos/server.sls
```diff
@@ -99,7 +99,7 @@ module_python2_update_repo:
 {% if 'uyuni-released' in grains['product_version'] %}
 server_pool_repo:
   pkgrepo.managed:
-    - baseurl: http://{{ grains.get("mirror") | default("download.opensuse.org", true) }}/repositories/systemsmanagement:/Uyuni:/Stable/images/repo/Uyuni-Server-4.0-POOL-x86_64-Media1/
+    - baseurl: http://{{ grains.get("mirror") | default("download.opensuse.org", true) }}/repositories/systemsmanagement:/Uyuni:/Stable/images/repo/Uyuni-Server-POOL-x86_64-Media1/
```
moio: Can you please also fix the corresponding URL in minima.yaml?

Apart from that thanks for the fixes, good to merge!

PR #684, opened 2020-02-13
Use custom version of Avahi

Bischoff:
## What does this PR change?

This PR makes sure we use our own custom version of Avahi, whenever it is enabled (even if we use no reflector).

This custom version solves the "spurious names conflict" bug (https://bugzilla.suse.com/show_bug.cgi?id=1163683). It is meant to slowly disappear as Avahi 0.8 is out with the fix, as the fix is backported to Avahi 0.7 in SLE 15, and as SLE 12 gets deprecated.

## Yet to be determined

This is using repositories that are currently under `home:ebischoff` in OBS. @moio, maybe you  rather want it under some Sumaform project? That could also spare us an extra repo.

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -2,13 +2,39 @@ include:
   - default.hostname
 
 {% if grains['use_avahi'] %}
-avahi_pkg:
-  pkg.installed:
-    {%- if grains['os_family'] == 'Debian' %}
-    - name: avahi-daemon
+
+# TODO: remove the following state when fix to bsc#1163683 is applied to all the SLES versions we use
+{% if grains['osfullname'] == 'SLES' %}
+custom_avahi_repo:
+  pkgrepo.managed:
+    - humanname: custom_avahi
+    {% if grains.get('osmajorrelease', None)|int() == 12 %}
+    - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.6.32/SLE_12_SP4/
     {% else %}
-    - name: avahi
+    - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.7/SLE_15_SP1/
     {% endif %}
+    - priority: 95
+    - gpgcheck: 0
+{% endif %}
+
+# TODO: replace 'pkg.latest' with 'pkg.installed' when fix to bsc#1163683 is applied to all the SLES versions we use
+avahi_pkg:
+  pkg.latest:
+    - pkgs:
+      {% if grains['os_family'] == 'Debian' %}
+      - avahi-daemon
+      - libavahi-common-data
+      - libavahi-common3
+      - libavahi-core7
+      {% elif grains['os_family'] == 'RedHat' %}
+      - avahi
+      - avahi-libs
+      {% elif grains['os_family'] == 'Suse' %}
+      - avahi
+      - avahi-lang
+      - libavahi-common3
+      - libavahi-core7
+      {% endif %}
 
 avahi_change_domain:
   file.replace:
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -310,7 +310,7 @@ module "server" {
 }
 ```
 
-If Avahi is enabled and you are running Docker on a minion, you will need an Avahi reflector on the minion to provide Avahi resolution inside of the containers. A typical example is the Cucumber testsuite which uses such a setup. An Avahi reflector can be enabled via:
+If Avahi is enabled and you are running Docker on a minion, you will need an Avahi reflector on the minion to provide multicast domain name resolution inside of the containers. A typical example is the Cucumber testsuite which uses such a setup. An Avahi reflector can be enabled via:
 
 ```hcl
 module "minion" {
@@ -319,8 +319,6 @@ module "minion" {
   ...
 }
 ```
-
-Beware this may trigger [a known Avahi bug](https://github.com/lathiat/avahi/issues/117). This bug causes wrong names to be assigned to hosts (with unexpected `-2`, `-3`, etc. suffixes) in many circumstances, in particular when more than one reflector is in use on the same network.
 
 
 ## Additional network and SUSE Manager for Retail
--- a/salt/minion/reflector.sls
+++ b/salt/minion/reflector.sls
@@ -1,27 +1,5 @@
 {% if grains.get('use_avahi') and grains.get('avahi_reflector') %}
 
-# We upgrade to latest version because of a bug in Avahi:
-#   https://github.com/lathiat/avahi/issues/117
-# It does not remove the problem, but makes it less likely to happen
-reflector_package:
-  pkg.latest:
-    - pkgs:
-      {% if grains['os'] == 'SUSE' %}
-      - avahi
-      - avahi-lang
-      - libavahi-common3
-      - libavahi-core7
-      {% elif grains['os'] == 'Ubuntu' %}
-      - avahi-daemon
-      - libavahi-common-data
-      - libavahi-common3
-      - libavahi-core7
-      {% elif grains['os_family'] == 'RedHat' %}
-      - avahi
-      - avahi-libs
-      {% endif %}
-    - refresh: true
-
 reflector_configuration:
   file.replace:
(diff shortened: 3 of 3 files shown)


salt/default/avahi.sls
```diff
@@ -2,13 +2,55 @@ include:
...
+# TODO: remove the following two states when avahi-0.8 is out
+{% if grains['osfullname'] == 'SLES' %}
+custom_avahi_repo:
+  pkgrepo.managed:
+    - humanname: custom_avahi
+    {% if grains.get('osmajorrelease', None)|int() == 12 %}
+    - baseurl: http://download.suse.de/ibs/home:/ebischoff:/avahi_32/SLE_12_SP4/
     {% else %}
-    - name: avahi
+    - baseurl: http://download.suse.de/ibs/home:/ebischoff:/avahi_40/SLE_15_SP1/
```
srbarrios: If we use IBS, it would br possible for other users not working for SUSE, with access to IBS, to use it?

juliogonzalez: No. IBS is only accessible to SUSE employees. But we can host this at OBS. I don't see what the problem could be.

salt/default/avahi.sls
```diff
@@ -2,13 +2,55 @@ include:
...
-    {%- if grains['os_family'] == 'Debian' %}
-    - name: avahi-daemon
+
+# TODO: remove the following two states when avahi-0.8 is out
+{% if grains['osfullname'] == 'SLES' %}
+custom_avahi_repo:
+  pkgrepo.managed:
+    - humanname: custom_avahi
+    {% if grains.get('osmajorrelease', None)|int() == 12 %}
+    - baseurl: http://download.suse.de/ibs/home:/ebischoff:/avahi_32/SLE_12_SP4/
```
juliogonzalez: Same here. We need to consider OBS.

salt/default/avahi.sls
Bischoff: @srbarrios @juliogonzalez Can we put source code for SLES version of avahi on OBS? That might be the problem. It might have SLES only patches, SLES dependencies, etc.

If I'm not wrong, the real solution is to use only openSUSE clients for Uyuni. We're not there yet, but that's the plan, isn't it? When that happens, I can make sumaform point to that repos for those clients.

salt/default/avahi.sls
Bischoff: See above.

salt/default/avahi.sls
Bischoff: Also, there is another thing to decide: this is currently in `home:ebischoff` in IBS. Maybe @moio wants it under some Sumaform project.

juliogonzalez: The source code of SLE packages is public:
https://build.opensuse.org/package/show/SUSE:SLE-15-SP1:Update/avahi
https://build.opensuse.org/package/show/SUSE:SLE-12-SP4:Update/avahi

Morever, avahi is `LGPL-2.1-or-later`, and we have **external** contributors to the package maintenance:
https://build.opensuse.org/package/view_file/SUSE:SLE-15-SP1:Update/avahi/avahi-mono.changes?expand=1
https://build.opensuse.org/package/view_file/SUSE:SLE-12-SP4:Update/avahi/avahi-mono.changes?expand=1
https://build.opensuse.org/package/view_file/SUSE:SLE-12-SP4:Update/avahi/avahi-qt4.changes?expand=1
https://build.opensuse.org/package/view_file/SUSE:SLE-15-SP1:Update/avahi/avahi-qt4.changes?expand=1

The only restriction about not making patches public is for the embargoed bug reports (besides: even the SPEC is free software as it uses MIT license).

So shouldn't be a problem at all :-)

Bischoff: Thanks for answering my legal doubts, @juliogonzalez .I'll mark the thread as resolved.

juliogonzalez: Disclaimer: I am **not** a lawer :-D

Bischoff: aka "IANAL" :smile_cat:

salt/default/avahi.sls
```diff
@@ -2,13 +2,55 @@ include:
...
+# TODO: remove the following two states if fix to bsc#1163683 gets backported to these distributions
+{% if grains['osfullname'] == 'SLES' %}
+custom_avahi_repo:
+  pkgrepo.managed:
+    - humanname: custom_avahi
+    {% if grains.get('osmajorrelease', None)|int() == 12 %}
+    - baseurl: http://download.opensuse.org/repositories/home:/ebischoff:/avahi:/0.6.32/SLE_12_SP4/
     {% else %}
-    - name: avahi
+    - baseurl: http://download.opensuse.org/repositories/home:/ebischoff:/avahi:/0.7/SLE_15_SP1/
```
moio: This should really go in this OBS project:

https://build.opensuse.org/project/show/systemsmanagement:sumaform:tools

(Tagline is "software for sumaformed machines" :-) )

salt/default/avahi.sls
```diff
@@ -2,13 +2,55 @@ include:
...
     {% endif %}
+    - priority: 95
+    - gpgcheck: 0
+
+avahi_pkg:
+  pkg.installed:
+    - pkgs:
+      {% if grains.get('osmajorrelease', None)|int() == 12 %}
+      - avahi: 0.6.32-42.1
```
moio: Pinning versions is pretty fragile, shouldn't we simply use a higher version than the one we are overriding?

If the version is correctly set in OBS all should be fine.

salt/default/avahi.sls
```diff
@@ -2,13 +2,55 @@ include:
...
+      - libavahi-core7: 0.6.32-42.1
+      {% else %}
+      - avahi: 0.7-42.1
+      - avahi-lang: 0.7-42.1
+      - libavahi-common3: 0.7-42.1
+      - libavahi-core7: 0.7-42.1
+      {% endif %}
+{% else %}
+
+avahi_pkg:
```
moio: Why is this ID separated from the one above, can't it be one only state?

salt/default/avahi.sls
Bischoff: ok, that was my question in the PR description...

When is this repo activated on hosts? I don't see it with "zypper lr" .

salt/default/avahi.sls
Bischoff: Honestly, I found this a lot less fragile than using `pkg.latest` as I was doing previously.

salt/default/avahi.sls
Bischoff: It is only one state. what is before the `else` is the hack, and should go away on the long run, while the part below should stay.

salt/default/avahi.sls
srbarrios: I also prefer concrete versions for dependencies and upgrade them when we consider it. Just in case some new version has some breaking change.

Bischoff: ok, then we are two control freaks :sunglasses:

moio: Problem is what happens when a new SLE update comes out with other patches.

As I understand, the best solution is to branch the package in OBS so that it contains a link diff, and that will produce a package with a higher build version even when the original is updated.

As for vendor changes, we explicitly enable them and AFAIK they do work:

https://github.com/uyuni-project/sumaform/blob/master/salt/repos/default.sls#L303-L318

Bischoff: > Problem is what happens when a new SLE update comes out with other patches.
> 
> As I understand, the best solution is to branch the package in OBS so that it contains a link diff, and that will produce a package with a higher build version even when the original is updated.

This is black magic to me, do you have the procedure I need to follow?

> As for vendor changes, we explicitly enable them and AFAIK they do work:
> 
> https://github.com/uyuni-project/sumaform/blob/master/salt/repos/default.sls#L303-L318

When I tried, I was building in IBS, that's probably the reason why it was not working for me.

Bischoff: Works, apart the versions are unpredictible... resolving conversation.

Bischoff:
Latest version works, but now we don't pin anymore a given release, there seems to be no guarantee that our version is more recent than the one from the distro.

moio:
Thank you very much

Bischoff:
Thanks for the merge @moio.

PR #687, opened 2020-02-19
Aws backend support

rjmateus:
## What does this PR change?
Issue: https://github.com/SUSE/spacewalk/issues/10025

This PR adds support for aws provider, defining the needed specific backend modules. Also tests support for several different client OS's.

Used as base the refactoring from https://github.com/uyuni-project/sumaform/pull/627.

## Backend modules:
| module   | Create Machine | Provisioning |
| ----------- | --------------------- | ----------------- |
| base       |               X           |       X           |
| network  |               X           |       X           |
| host        |               X           |         X         |

## Modules Test:
| module                       | Create Machine | Provisioning |
| --------------------------- | -------------------- | ---------------- |
| base                           |         X             | X (leap15.1) |
| client                           |        X              |         X          | 
| controller                    |         X              |         X         |
| cucumber_testsuite    |         X              |        X          |
| grafana                       |                         |                    |
| host                            |          X             |         X         |
| locust                         |                          |                    |
| minion                        |           X             |         X        |
| mirror                         |          X              |         X        |
| proxy                          |          X             |         X        |
| pts                              |                          |                   |
| pxe_boot                    |           X             |                   |
| server                         |           X            |         X        |

## Images for server/proxy
|Product             | Image                | Server       | Proxy |
| --------------------| -------------------- | -------------- | -------- |
| uyuni-released | Opensuse 15.1 | X               | X |
| uyuni-master    | Opensuse 15.1 | X               | X |
| ~head~ | ~sles15sp2~ | |
| 4.0-nightly                                    | sles15sp1 | X | X |
| 4.0-released                                 | sles15sp1 | X | X |
| 3.2-nightly                                    | sles15sp1 | X | X |
| 3.2-released                                 | sles15sp1 | X | X |

## Images for client/minion
| Image                | client | minion | sshminon |
| --------------------- | ------- | --------- |  ------------ |
| Opensuse 15.1  | X       | X         | X  |
| Opensuse 15     | X       | X         | X  |
| ~sles15sp2~      | Image not available on aws | | |
|sles15sp1           |  X      | X         | X  |
|sles15                 | X       | X         | X  |
|sles12sp5           | X       | X         | X  |
|sles12sp4           | X       | X         | X  |
|sles12sp3           | X       | X         | X  |
|sles12sp2           | X       | X         | X  |
|~sles12sp1~       | Image not available on aws |
|~sles12~             | Image not available on aws |
|sles11sp4           | X       |  X        | error [2] (TLS 1.2)  |
|centos8           | no official image available [1]  |
|centos7           | X           | X         | X  |
|centos6           | X           | X         | error [2] (TLS 1.2)  |
| ubuntu1804    | -           | X  | X  |
| ubuntu1604    | -           | X  | X  |

Links:
[1] https://bugs.centos.org/view.php?id=16614

ERRORS:
[3] Known error and is documented (https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/client-configuration/tshoot-clients.html#_ssl_errors). Nothing do do here.

Smoke tests were successfully done on April 14.

## TODO:
- [x] Backport changes on create volume from aws to libvirt (PR: )
- [x] Update salt with changes from Cloud team PR
- [x] Document how to create a full working mirror
- [x] Document how to create a mirror data disk snapshot with terraform
- [X] Circular dependency between mirror/bastian and base (possible solution under test) 
- [X] Bastion and mirror in different machines
- [x] Change network module to allow outbound connection to intenet
- [x] Move mirror to private network 
- [X] versions.tf file cannot be shared across back-ends since each one has is different versions
- [X] After creat an instance terraform wants to replace some automatic tags
- [X] Machine ssh user with password "uyumi:linux". Created with cloud-init
- [X] Server is unable to connect to mirror (error: no metadata)
- [X] Extra disk tag change trigger's recreate
- [X] adapt script `configure_aws_tunnels.rb`
- [X] Create machines with the JeOS
- [x] Enable IPv6 on aws network (Open issue and do latter if needed) (https://github.com/SUSE/spacewalk/issues/11182)
- [x] Open Bug for sles12sp5 aws image - traditional client bootstrap. Missing dbus-1-glib (https://github.com/SUSE/spacewalk/issues/11177)
- [ ] Define and use the additional network
- [ ] Add an highstate for bastion to deploy extra authentication keys 
- [ ] Contact cloud team to check use case


backend_modules/aws/host/main.tf
```diff
@@ -1,91 +1,234 @@
...
     virtual_name = "ephemeral1"
   }
 
-  tags {
-    Name = "${var.name_prefix}-${var.name}-${count.index}"
+  tags = {
+    Name = "${local.resource_name_prefix}${var.quantity > 1 ? "-${count.index + 1}" : ""}"
+  }
+
+  // ## HACH
```
moio:
```suggestion
  // ## HACK
```

backend_modules/aws/README.md
```diff
@@ -2,40 +2,112 @@
...
+    ssh_allowed_ips   = ["1.2.3.4"]
+    key_name = "my-aws-key"
+    key_file = "/path/to/key.pem"
+}
+...
+```
+
+### Host modules
+
+Following settings apply to all modules that create one or more hosts of the same kind, such as `suse_manager`, `suse_manager_proxy`, `client`, `grafana`, `minion`, `mirror`, `sshminion`, `pxe_boot` and `virthost`:
```
moio, line 61: Please document `pxe_boot` is not supported if it is actually not at least in the version we are merging.

(IOW it's completely OK, just let's be upfront about limitations)

backend_modules/aws/README.md
```diff
@@ -2,40 +2,112 @@
...
+
+| Variable name   | Type     | Default value                                                    | Description                                                         |
+|-----------------|----------|------------------------------------------------------------------|---------------------------------------------------------------------|
+| key_name        | string   | [from base Module](base-module)                                  | ssh key name in AWS                                                 |
+| key_file        | string   | [from base Module](base-module)                                  | ssh key file                                                        |
+| ssh_user        | string   | Aws OS specify ssh user (ec2-user, centos, ubuntu, etc)          | ssh user to use in ssh into the machine for provisioning            |
```
moio, line 67:
```suggestion
| ssh_user        | string   | OS-specific SSH user (ec2-user, centos, ubuntu, etc)          | ssh user to use in ssh into the machine for provisioning            |
```

backend_modules/aws/README.md
```diff
@@ -2,40 +2,112 @@
...
+| bastion_host    | string   | [from base Module](base-module)                                  | bastian host use to connect machines in private network             |
```
moio, line 68:
```suggestion
| bastion_host    | string   | [from base Module](base-module)                                  | bastion host used to connect to machines in the private network             |
```

backend_modules/aws/README.md
```diff
@@ -2,40 +2,112 @@
...
+| public_instance | boolean  | `false`                                                          | boolean to set host to private or public network                    |
```
moio, line 69:
```suggestion
| public_instance | boolean  | `false`                                                          | boolean to connect host either to the private or the public network                    |
```

backend_modules/aws/README.md
```diff
@@ -2,40 +2,112 @@
...
+  provider_settings = {
+    public_instance = true
+    instance_type   = "t2.small"
+  }
+...
+```
+
+`server`, `proxy` and `mirror` modules have configuration settings specific for extra data volumes, those are set via the `volume_provider_settings` map variable. They are described below.
+
+ * `volume_snapshot_id = <String>` data volume snapshot id to be used as base for the new disk (default value `null`)
```
moio, line 85:
```suggestion
 * `volume_snapshot_id`: data volume snapshot id to be used as base for the new disk (default value: `null`)
```

backend_modules/aws/README.md
```diff
@@ -2,40 +2,112 @@
...
+ * `type = <String>` Disk type that should be used (default value `sc1`). One of: "standard", "gp2", "io1", "sc1" or "st1".
```
moio, line 86:
```suggestion
 * `type`: volume type that should be used (default value `sc1`). See the list at the [AWS EBS Volume Type documentation page](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html).
```

(you'll never know when they add more)

backend_modules/aws/README.md
```diff
@@ -2,40 +2,112 @@
...
+
+ An example follows:
+ ``` hcl
+volume_provider_settings = {
+  volume_snapshot_id = "my-data-snapshot"
```
moio, line 91:
```suggestion
  volume_snapshot_id = "snap-0123abcd"
```

backend_modules/aws/README_ADVANCED.md
```diff
@@ -0,0 +1,139 @@
...
+
+## mirror
+
+In addition to acting as a bastion host for all other instances, the `mirror` host serves all repos and packages used by other instances.
+It works similarly to the one for the libvirt backend, allowing instances in the private subnet to be completely disconnected from the Internet.
+The `mirror` host's data volume can be created from a pre-populated snapshot, which allows it to be operational without lengthy channel synchronization.
+
+### Set up mirror machine
+
+When creating mirror machine one have two possibilities:
```
moio, line 11: I'd reverse the ordering as I can expect most users not to have a snapshot the first time they go through the guide.

backend_modules/aws/README_ADVANCED.md
```diff
@@ -0,0 +1,139 @@
...
+}
+```
+
+**Sync mirror data (this will take some time):**
+
+   If you are using released versions only or creating a mirror disk from scratch:
+   1. `ssh root@<MIRROR_HOST>`
+   1. `sudo bash /root/mirror.sh`
+
+   If you need access to nightly builds or DEVEL packages (one needs access to a full sync mirror, which could be on a local machine):
```
moio, line 57:
```suggestion
If you need access to SUSE internal nightly builds or `Devel:` packages, you will first of all need a fully set up and synced libvirt based mirror on a machine in the engineering network. Once you have it you can transfer development packages to your AWS mirror via rsync, for example via the commands below:
```

backend_modules/aws/README_ADVANCED.md
```diff
@@ -0,0 +1,139 @@
...
+   1. `scp <YOUR_AWS_KEY> root@<LOCAL_MIRROR_HOST>://root/key.pem`
+   2. `ssh root@<MIRROR_HOST>`
+   3. `zypper in rsync`
+   4. `rsync -av0 --delete -e 'ssh -i key.pem -o "proxycommand ssh -W %h:%p -i key.pem ec2-user@<BASTION_MACHINE_PUBLIC_DNS>"' /srv/mirror/ ec2-user@<MIRROR_PRIVATE_DNS>://srv/mirror/`
+
+#### Creating data disk snapshot
+1. Create a disk snapshot (this will take some time)
```
moio, line 64: Is this equivalent to taking a snapshot from the AWS console? If so, note it, as it might be easier at least to some users.

backend_modules/aws/README_ADVANCED.md
```diff
@@ -0,0 +1,139 @@
+# Advanced configurations
+
+## mirror
+
+In addition to acting as a bastion host for all other instances, the `mirror` host serves all repos and packages used by other instances.
```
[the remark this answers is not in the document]
rjmateus: This is not true, mirror and bastion are separated machines now.

salt/default/minimal.sls
```diff
@@ -6,6 +6,7 @@ include:
   - default.avahi
   - repos
 
+{% if not grains.get('no_install') | default(false) %}
```
moio, line 9: This can create a lot of problems if the right version of Salt is not installed at this place, and getting the right version in all cases is pretty tricky, which is the reason why it is here.

I would suggest not to make this something that can be disabled unless we have a very strong reason to do so.

Maybe I am missing something...

salt/server/init.sls
```diff
@@ -47,7 +47,9 @@ server_setup:
     - name: /usr/lib/susemanager/bin/migration.sh -l /var/log/susemanager_setup.log -s
     - creates: /root/.MANAGER_SETUP_COMPLETE
     - require:
+      {% if not grains.get('no_install') | default(false) %}
```
moio, line 50: I think I understand the reasoning behing the addition of this check here #627, although I am still unconvinced this is really needed. What should happen is that the `server_packages` state is still executed and does nothing - assuming the image packages are set up correctly.

Please remove this flag for now, and let's reintroduce it only if and when we have a problem.

backend_modules/aws/host/main.tf
```diff
@@ -1,91 +1,222 @@
...
+        gpg_keys                      = var.gpg_keys
+        connect_to_base_network       = var.connect_to_base_network
+        connect_to_additional_network = var.connect_to_additional_network
+        reset_ids                     = true
+        ipv6                          = var.ipv6
+        data_disk_device              = contains(var.roles, "server") || contains(var.roles, "proxy") || contains(var.roles, "mirror") ? "xvdf" : null
+
+        // optimizations for cloud install
+        no_install = contains(var.roles, "server") ? true : false
+        no_update  = contains(var.roles, "mirror") ? true : false
```
moio, line 197: I think those new flags should be removed at least for now, see details in later comments.

backend_modules/aws/host/user_data.yaml
```diff
@@ -0,0 +1,157 @@
...
+  expire: false
+  list: |
+    root:linux
+%{ endif }
+
+%{ if image == "opensuse151" || image == "opensuse150"}
+packages: ["salt-minion"]
+
+runcmd:
+  # HACH for cloud image, since it has this package which blocks server deployment
```
moio, line 17:
```suggestion
  # HACK this package should not be in any image, as it is internal to OBS. To be removed when new images ship without it
```

backend_modules/aws/host/user_data.yaml
```diff
@@ -0,0 +1,157 @@
...
+  - zypper rm -y gettext-runtime-mini
+  # HACH: machine already have some repos that where conflicting with the ones we want to install.
+  # Possible bug in zypper salt module
```
moio, line 20:
```suggestion
  # HACK: instance might already have some repos that end up conflicting with the ones we want to install.
  # Possible bug in zypper's Salt module
```

backend_modules/aws/host/user_data.yaml
```diff
@@ -0,0 +1,157 @@
...
+  - zypper removerepo --all
+#  - echo "public_instance = ${public_instance}" > /tmp/public_instance
```
moio, line 22: Please uncomment or remove

backend_modules/aws/host/user_data.yaml
```diff
@@ -0,0 +1,157 @@
...
+    - id: os_pool_repo
+      name: os_pool_repo
+      baseurl: http://${mirror_url}/SUSE/Products/SLE-SERVER/12-SP2/x86_64/product/
+      enabled: 1
+      autorefresh: 1
+
+packages: ["salt-minion"]
+%{ endif }
+%{ if image == "sles11sp4"}
+# Big hack to work with sles11Ssp4
```
moio, line 71:
```suggestion
```

backend_modules/aws/host/user_data.yaml
```diff
@@ -0,0 +1,157 @@
...
+runcmd:
```
moio, line 72:
```suggestion
# HACK: SLES 11 SP4 does not ship with Salt installed
runcmd:
```

main.tf.aws-testsuite.example
```diff
@@ -0,0 +1,99 @@
...
+    key_file          = ...
+    key_name          = ...
+}
+
+//provider "aws" {
+//  region     = local.region
+//  access_key = ...
+//  secret_key = ...
+//}
+
```
moio, line 13: I guess this block should not be commented out - or am I missing something?

main.tf.aws-testsuite.example
```diff
@@ -0,0 +1,99 @@
...
+    key_file          = local.key_file
+  }
+}
+
+output "key_file" {
+  value = local.key_file
+}
+//output "bastion_public_name" {
+//  value = module.cucumber_testsuite.configuration.base.bastion_host
+//}
```
moio, line 63: Why is this block commented out?

main.tf.aws.example
```diff
@@ -1,143 +1,100 @@
...
-  ami = "${data.aws_ami.opensuse422.image_id}"
-  key_name = "${local.key_name}"
-  key_file = "${local.key_file}"
-  data_volume_snapshot_id = "${data.aws_ebs_snapshot.data_disk_snapshot.id}" // see top comment in modules/aws/mirror/main.tf
-  public_subnet_id = "${module.aws_network.public_subnet_id}"
-  public_security_group_id = "${module.aws_network.public_security_group_id}"
-  cc_username = "${local.cc_username}"
-  cc_password = "${local.cc_password}"
-  name_prefix = "${local.name_prefix}"
+// ================= Configuration with Mirror =================
```
moio, line 32: I would suggest to have this as the default, thus uncommenting this section.

main.tf.aws.example
```diff
@@ -1,143 +1,100 @@
...
+//    mirror = length(module.mirror.configuration["hostnames"]) > 0 ? module.mirror.configuration["hostnames"][0] : null
+//  })
+//}
+//
+// ================= END MIRROR =================
+
+
+module "server" {
+  source             = "./modules/server"
+  //base_configuration = local.base_configuration
```
moio, line 65: What is the reason to leave this commented out?

main.tf.aws.example
```diff
@@ -1,143 +1,100 @@
...
-  product_version = "3.2-nightly"
-  roles = ["suse_manager_server"]
-  cc_username = "${local.cc_username}"
-  cc_password = "${local.cc_password}"
-  mirror_public_name = "${module.aws_mirror.public_name}"
-  mirror_private_name = "${module.aws_mirror.private_name}"
-}
+module "minion" {
+  source             = "./modules/minion"
+  //base_configuration = local.base_configuration
```
moio, line 77: What is the reason to leave this commented out?

modules/mirror/variables.tf
```diff
@@ -27,6 +27,12 @@ variable "ubuntu_distros" {
   default     = []
 }
 
+variable "minima_config" {
```
moio, line 30: Please use `minima_config_path` as the name, for coherence with other configurable paths we have around.

salt/mirror/init.sls
```diff
@@ -118,9 +123,10 @@ mirror_script:
     - watch:
       - file: /root/mirror.sh
 
+{% set fstype = grains.get('data_disk_fstype') | default('ext4', true) %}
```
moio, line 126: I do not see where this grain is set. Am I missing anything?

salt/mirror/minima-pubcloud.yaml
```diff
@@ -0,0 +1,152 @@
+storage:
```
moio, line 1: I am not sure whether this file should be in the main `sumaform` repo, or rather stored together with the Public Cloud Team's specific `main.tf`. The latter seems like a more sensible option to me, as we would not really know how to maintain this properly.

salt/mirror/minima_aws_testsuite.yaml
```diff
@@ -0,0 +1,318 @@
+storage:
```
moio, line 1: What is the need to have this file different from the default one?

backend_modules/aws/create_aws_mirror_snapshot
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/python3
```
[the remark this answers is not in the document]
rjmateus: This script was developed by the cloud team. I let it here in case these team need it. In a previous conversation with you (through email) it was suggested that if cloud team need it, could keep "in their own repos or in a sumaform fork". I was assuming (wrongly) it could stay in here. 
I will remove these script.

backend_modules/aws/host/main.tf
```diff
@@ -1,91 +1,222 @@
...
+  ami = lookup(lookup(var.base_configuration["ami_info"], var.image, {}), "ami", var.image)
+
+  provider_settings = merge({
+    key_name        = var.base_configuration["key_name"]
+    key_file        = var.base_configuration["key_file"]
+    ssh_user        = lookup(lookup(var.base_configuration["ami_info"], var.image, {}), "ssh_user", "ec2-user")
+    public_instance = false
+    volume_size     = 50
+    bastion_host    = lookup(var.base_configuration, "bastion_host", null)
+    instance_type = "t2.micro" },
```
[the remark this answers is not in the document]
rjmateus: Thanks for the explanation. I add it to todo list in future PR.

backend_modules/null/host/variables.tf
```diff
@@ -96,6 +96,12 @@ variable "provision" {
   default     = true
 }
 
+variable "cloud_init" {
+  description = "Indicates whether cloud-init should be applied to base image"
+  type        = bool
+  default     = true
```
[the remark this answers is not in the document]
rjmateus: This flag indicates if we should apply cloud init to the created VM. The only scenario I see now is when We create the bastion host, which at the moment we don't need to apply it. 
It's probably not a problem if we apply it, but it will take more time to create this machine.
I just had one idea. We can add a role "bastion" to that specific machine and decide if we should or not run the cloud init base on roles.
I will try it and revert this change.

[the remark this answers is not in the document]
moio: Hmm, ok, I would still suggest to treat the bastion like any other VM unless we have a compelling reason to.

main.tf.aws-testsuite.example
rjmateus: It should not be commented. I have change the libvirt example on master and adapt CI script but didn't rebase yet.
On todo list now

main.tf.aws-testsuite.example
rjmateus: I correct it on aws base example but forget to port to testsuite example

main.tf.aws.example
rjmateus: If we use the mirror we can use the local variable base_configuration to force all modules to wait for mirror to be created before creating all other modules.

main.tf.aws.example
rjmateus: comment above

backend_modules/null/host/variables.tf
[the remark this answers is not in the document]
rjmateus: Done. I removed the cloud_init variable

backend_modules/aws/create_aws_mirror_snapshot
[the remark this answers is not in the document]
rjmateus: script removed. It was ported from PR https://github.com/uyuni-project/sumaform/pull/627
I case of need it can be picked from there again

backend_modules/aws/host/main.tf
```diff
@@ -1,91 +1,222 @@
...
 
+resource "aws_volume_attachment" "data_disk_attachment" {
+  depends_on = [aws_instance.instance, aws_ebs_volume.data_disk]
+
+  count = var.additional_disk_size == null ? 0 : var.additional_disk_size > 0 ? var.quantity : 0
+
+  device_name = "/dev/xvdf"
+  volume_id   = aws_ebs_volume.data_disk[count.index].id
+  instance_id = aws_instance.instance[count.index].id
+  // volume tends not to detach, breaking terraform destroy, so skip destroying
```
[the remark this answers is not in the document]
rjmateus: When destroying it sometimes it stops when detaching the data disk from mirror. I was thinking that those lines were copied from AWS team PR, but I cannot find it.

salt/mirror/init.sls
```diff
@@ -2,9 +2,14 @@ include:
   - default
 
 system_update:
+{% if not grains.get('no_update') | default(false) %}
```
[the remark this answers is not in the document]
rjmateus: Done.

salt/default/init.sls
```diff
@@ -1,7 +1,9 @@
 include:
   - default.locale
   - default.minimal
+  {% if not grains.get('no_install') | default(false) %}
   - default.pkgs
+  {% endif %}
```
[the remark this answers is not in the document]
rjmateus: I have reverted the commit which adds these changes

salt/default/minimal.sls
rjmateus: revert it by removing the commit which adds these changes

salt/server/init.sls
rjmateus: done, fag removed. We can always re-apply if needed

salt/mirror/minima-pubcloud.yaml
rjmateus: Agree. File was removed.

salt/mirror/minima_aws_testsuite.yaml
rjmateus: This file exists, for now, so we don't have on AWS the following repositories:
- All that is "Maintenance:" 
- SLES15SP2 repositories 

As we talked, for now we don't want to have those packages on AWS MIrror. In future if is ok to have, we can just remove this file.

salt/mirror/init.sls
rjmateus: You didn't miss anythig, the grain is not set.

salt/mirror/minima_aws_testsuite.yaml
moio: How about using some Jinja templating, or just leaving them there (they should just fail and not really hurt)?

rjmateus:
@moio I'm adding support for additional network on AWS.
Some considerations about the network characteristics: 
* Network only allows connection from inside the network
* No access to outside 

Now the question.
When testing it, I faced a problem that I don't know if it should/needed to solve. If we set the property `connect_to_additional_network = true` and `connect_to_base_network = false` but no additional_network is configured on base host, meaning the additional network will not be created, the apply fails since the machine will not network to be associated with and AZ doesn't have a default subnet.

It also happens if we set both properties to false.

I see at least two options:
* Let the error occur and document it
* If no additional network is defined on base, fall back to private network

On libvirt backend I think the machine is created without any network interface or it will raise an error (I didn't test it yet, but looking at code is what I'm expecting. I will test it during the weekend) 

We also have a problem if we have a host connect only to the additional network, and then we remove the additional network configuration from base module. The apply will be hanging, since we have a machine associated with the network which will be destroyed. 

what are your thoughts about these problems?

rjmateus:
@moio Some updates about last comment. 
Minion host always set properties to connect to base and additional networks.
If we don't define the additional network configuration on base module we will not be able to create hosts that connects to both networks.

In aws host module we cannot use the output of base module to control a count variable to define if we should or not have a second network attached. This will raise an error when creating are environment from scratch.

This a very typical use case, not define additional network on base module and use the minion module.

The most simple solution: always create an additional network with some default cidr_block.

backend_modules/aws/network/main.tf
```diff
@@ -4,157 +4,223 @@
...
   }
 }
 
-resource "aws_subnet" "private" {
-  availability_zone = "${var.availability_zone}"
-  vpc_id = "${aws_vpc.main.id}"
-  cidr_block = "172.16.128.0/17"
-  map_public_ip_on_launch = false
+locals {
+  vpc_id = var.create_network? aws_vpc.main[0].id: null
```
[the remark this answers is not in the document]
rjmateus: done

salt/mirror/minima_aws_testsuite.yaml
rjmateus: The goal was to not have these repositories on AWS when we sync from a local mirror to aws. So this file should be used to the local mirror.
This is expect to be something temporary until it's get clarified if we can have those repositories on cloud.
Do you think we should have the effort to use jinja Template for it or just leave it there?

moio: I would even let them there, internal repos will simply silently fail to sync and we do not get the extra overhead of having templating. Only if this becomes a problem I'd try to address it.

moio:
> @moio I'm adding support for additional network on AWS.
> Some considerations about the network characteristics:
> 
>     * Network only allows connection from inside the network
> 
>     * No access to outside
> 
> 
> Now the question.
> When testing it, I faced a problem that I don't know if it should/needed to solve. If we set the property `connect_to_additional_network = true` and `connect_to_base_network = false` but no additional_network is configured on base host, meaning the additional network will not be created, the apply fails since the machine will not network to be associated with and AZ doesn't have a default subnet.
> 
> It also happens if we set both properties to false.
> 
> I see at least two options:
> 
>     * Let the error occur and document it
> 
>     * If no additional network is defined on base, fall back to private network
> 
> 
> On libvirt backend I think the machine is created without any network interface or it will raise an error (I didn't test it yet, but looking at code is what I'm expecting. I will test it during the weekend)
> 
> We also have a problem if we have a host connect only to the additional network, and then we remove the additional network configuration from base module. The apply will be hanging, since we have a machine associated with the network which will be destroyed.
> 
> what are your thoughts about these problems?

Those are all true problems, nevertheless, at the moment we do not have very many users/use cases that really require the additional network, so I do not think those are very urgent to solve.

I would recommend in this case to ship something imperfect (such as a solution in documentation only, or initially shipping AWS support with no additional network support at all) rather than something more perfect but later. I see much more urgency to merge this already big body of work first.

IMO, long term we could consider lifting `additional_network` from a provider-dependent attribute to a provider-independent one, unless there are reasons why that cannot happen. At that point we can at least guarantee it's set correctly in `cucumber_testsuite`, which is one (and probably the most important) of our use cases.

moio:
> The most simple solution: always create an additional network with some default cidr_block.

I would not have a problem with that at all.

salt/mirror/minima_aws_testsuite.yaml
rjmateus: agree. I will remove the duplicated file.

rjmateus:
@moio All smoke tests have passed. I will merge it now and any issue will be addressed in future PR's.

PR #688, opened 2020-02-20
Libvirt: refactor volume module/additional disk

rjmateus:

## What does this PR change?

Problem: Have a data volume in a separated module and use is output in host  doesn't allow the provider to determine all fields at plan phase, returning an error and didn't let us apply.

Have a data volume module is bringing more problems than helping.
In Libvirt we get was able to find a hack to work with it. In aws we are getting even more problems, which are expected to happen with other cloud providers.

This pr moves the extra data volume creation inside the host module, which solves the problem for good.


backend_modules/null/host/variables.tf
```diff
@@ -100,3 +100,13 @@ variable "provider_settings" {
   description = "Map of provider-specific settings, see the modules/libvirt/README.md"
   default     = {}
 }
+
+variable "repository_disk_size" {
```
moio: "Repository" does not make sense as a name if we are talking about a generic host. I think we need a more generic name (and description) here.

modules/mirror/variables.tf
```diff
@@ -32,6 +32,11 @@ variable "provider_settings" {
   default     = {}
 }
 
+variable "repository_disk_size" {
+  description = "Size of an aditional disk for /var/spacewalk partition, defined in GiB"
```
moio: mirror does not have such partition, rather, it has a partition where content is stored

rjmateus:
> Do I understand correctly that no changes are needed on main.tfs of regular mirror users?
You have understood it correctly, no changes are needed. I need to test one scenario that I just remember. Create a mirror with old code, and then re-apply with the new and see how terraform deal the data disk move to another sub-module.

rjmateus:
@moio I have tested this change in my local existing mirror and users needs to replace some information in the terraform.state file to point the data disk to is different location:
```
sed -i 's/module.mirror.module.data_disk/module.mirror.module.mirror.module.host/g'  terraform.tfstate
sed -i 's/"name": "disk"/"name": "data_disk"/g'  terraform.tfstate
```
Do you think it is ok? 
Should I add it to documentation or should it be present in a notification email, as we had done when we move to provider-independent modules?

moio:
I like it in the email better

modules/mirror/variables.tf
```diff
@@ -32,6 +32,11 @@ variable "provider_settings" {
   default     = {}
 }
 
+variable "repository_disk_size" {
+  description = "Size of an aditional disk for /srv/mirror partition, defined in GiB"
```
moio:
```suggestion
  description = "Size of an aditional disk for the /srv/mirror partition, defined in GiB"
```

backend_modules/libvirt/host/main.tf
```diff
@@ -30,6 +29,14 @@ resource "libvirt_volume" "main_disk" {
   count            = var.quantity
 }
 
+resource "libvirt_volume" "data_disk" {
+  name  = "${local.resource_name_prefix}${var.quantity > 1 ? "-${count.index + 1}" : ""}-data-disk"
+  // needs to be converted to bytes
+  size  = var.additional_disk_size * 1024 * 1024 * 1024
+  pool  = lookup(var.volume_provider_settings, "pool", "default")
+  count = var.additional_disk_size > 0 ? 1 : 0
```
moio:
```suggestion
  count = var.additional_disk_size > 0 ? var.quantity : 0
```

moio: Besides, is `0` an acceptable value for `additional_disk_size`? I would assume it is either `null` or a positive number.

modules/proxy/variables.tf
```diff
@@ -105,7 +105,7 @@ variable "image" {
 }
 
 variable "repository_disk_size" {
-  description = "Size of an aditional disk for /var/spacewalk partition"
+  description = "Size of an aditional disk for /var/spacewalk partition, defined in GiB"
```
moio: Reminder: users should know units of measure changed, or at next apply they will find a disk some orders of magnitude bigger than expected! Please note this in the email (and check our automated testsuites just in case we use this).

backend_modules/libvirt/host/main.tf
rjmateus: This is a good point. I was assuming the default value as 0 for not having a data disk, but in fact that are no reason one shouldn't be possible to set null as value.
I will change it to be possible to set null or zero.

rjmateus: I have changed the code to allow null values. I will mark this as resolved.

rjmateus:
I have made all changes, and checked sumaform-testsuite-runner and we have one job that are using the disk size configuration:
https://gitlab.suse.de/galaxy/sumaform-test-runner/-/blob/master/QAM/40/main-pip.tf
I'm waiting on QA to reply if it is in use or not. But after merge this PR I will change it accordingly.
Waiting for your approval to send e-mail and merge this PR.

PR #689, opened 2020-02-25
Update TROUBLESHOOTING.md

renner:
## What does this PR change?

This is just an update to make it more clear how to create the symlink in case of the backend not found error.


moio:
Thanks!

PR #690, opened 2020-03-05
Improve mirroring wrapper that handles errors in all cases

juliogonzalez:
## What does this PR change?

So now cron will send email to the administrator **only if** there are errors.

It also allows a more flexible usage of mirror.sh, for example to spread execution of each step... or to define a reduced set of repositories with a new yaml file, and run a second cron job with that an extra time each day.

This comes from the SUSE Manager internal repository, as we precisely needed that.

I am also porting a fix from @mcalmer for handling of the nvidia repo. Sorry for not using a separate PR, but I was not aware our script was taking from sumaform :-)

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -133,6 +133,13 @@ module "mirror" {
 
 Note that `mirror` must be populated before any host can be deployed - by default its cache is refreshed nightly via `cron`, you can also schedule a one-time refresh via the `/root/mirror.sh` script.
 
+It is also possible to use parameters with `/root/mirror.sh` to limit what is going to be done.
+
+Check the help with:
+```bash
+/root/mirror.sh -h
+```
+
 ## Virtual hosts
 
 Virtualization hosts are Salt minions that are also capable to run virtual machines using the KVM hypervisor.
--- a/salt/mirror/init.sls
+++ b/salt/mirror/init.sls
@@ -25,11 +25,28 @@ minima:
       - pkg: mozilla_certificates
       - pkg: system_update
 
+mirror_configuration_dir:
+  file.directory:
+    - name: /root/.minima
+    - user: root
+    - group: root
+    - mode: 750
+
 minima_configuration:
   file.managed:
-    - name: /root/minima.yaml
+    - name: /root/.minima/minima.yaml
     - source: salt://mirror/minima.yaml
     - template: jinja
+    - require:
+      - file: mirror_configuration_dir
+
+mirrorsh_configuration:
+  file.managed:
+    - name: /root/.minima/mirror.sh.conf
+    - source: salt://mirror/mirror.sh.conf
+    - template: jinja
+    - require:
+      - file: mirror_configuration_dir
 
 tar:
   pkg.installed:
@@ -78,9 +95,12 @@ mirror_script:
     - name: /root/mirror.sh
     - source: salt://mirror/mirror.sh
     - mode: 755
-    - template: jinja
   cron.present:
-    - name: /root/mirror.sh
+{% if grains.get('use_mirror_images') %}
+    - name: /root/mirror.sh --mirror-images --refresh-scc-data --apt-mirror --minima-sync=/root/.minima/minima.yaml --config=/root/.minima/mirror.sh.conf &> /var/log/full-mirror.log || cat /var/log/full-mirror.log
+{% else %}
+    - name: /root/mirror.sh --refresh-scc-data --apt-mirror --minima-sync=/root/.minima/minima.yaml --config=/root/.minima/mirror.sh.conf &> /var/log/full-mirror.log || cat /var/log/full-mirror.log
+{% endif %}
     - identifier: MIRROR
     - user: root
     - hour: 20
@@ -89,6 +109,7 @@ mirror_script:
       - pkg: cron
       - archive: minima
       - file: minima_configuration
+      - file: mirrorsh_configuration
       - file: mirror_script
   service.running:
     - name: cron
--- a/salt/mirror/mirror.sh
+++ b/salt/mirror/mirror.sh
@@ -1,28 +1,150 @@
 #!/bin/bash
-set -e
-set -x
+
+# Get script name
+SCRIPT=$(basename ${0})
+
+function show_help() {
+  echo "Wrapper to help with mirror syncing"
+  echo ""
+  echo "Syntax: "
+  echo ""
+  echo "${SCRIPT} <ARGUMENTS>"
+  echo ""
+  echo "Mandatory arguments:"
+  echo ""
+  echo " --mirror-images      Mirror QCOW2 images"
+  echo " --refresh-scc-data   Refresh data from SCC"
+  echo " --apt-mirror         Mirror Apt repositories"
+  echo " --minima-sync=<FILE> Use minima to sync repositories specified on <FILE>"
+  echo "                      By default: ${HOME}/.minima/minima.yaml"
+  echo " --config=<FILE>      Config file with SCC credentials and list of images."
+  echo "                      By default: ${HOME}/.minima/mirror.sh.conf"
+  echo ""
+  echo "If called withou targuments, thiat will be equivalent to calling:"
+  echo "${SCRIPT} --mirror-images --refresh-scc-data --apt-mirror --minima-sync=${HOME}/.minima/minima.yaml"
+}
+
+function print_incorrect_syntax() {
+  echo "ERROR: Incorrect syntax (use -h for help)"
+}
+
+function set_error() {
+  if [ ${1} -ne 0 ]; then
+    ERRORS="TRUE"
+  fi
+}
+
+function minima_sync() {
+  if [ ! -f ${1} ]; then
+    echo "ERROR: minima configuration file ${1} does not exist!"
+    set_error 1
+    return 1
+  fi
+  minima -c ${1} sync
+  set_error ${?}
+}
+
+function refresh_scc_data() {
+  if [ -z "${SCC_CREDS}" ]; then
+    echo "ERROR: configuration file ${CONFIG} does not define SCC_CREDS!"
+    set_error 1
+    return 1
+  fi
+  /root/refresh_scc_data.py "$SCC_CREDS"
+  set_error ${?}
+}
+
+function apt_mirror() {
+  /usr/sbin/apt-mirror
+  set_error ${?}
+}
+
+function mirror_images() {
+  if [ -z "${IMAGES}" ]; then
+    echo "ERROR: configuration file ${CONFIG} does not define IMAGES!"
+    set_error 1
+    return 1
+  fi
+  # will mirror them in /srv/mirror under the same path as the URL
+  for IMAGE in ${IMAGES}; do
+    wget --mirror --no-host-directories ${IMAGE}
+    set_error ${?}
+  done
+}
+
+function adjust_external_repos() {
+  # external repos are special
+  mkdir -p repo/RPMMD
+  set_error ${?}
+  if [ ! -L repo/RPMMD/SLE-15-GA-Desktop-NVIDIA-Driver ]; then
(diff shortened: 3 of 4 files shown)


salt/mirror/mirror.sh
```diff
@@ -1,28 +1,118 @@
...
+  /usr/sbin/apt-mirror
+  set_error ${?}
+}
+
+function mirror_images() {
+  # will mirror them in /srv/mirror under the same path as the URL
+  REPOS="https://github.com/moio/sumaform-images/releases/download/4.3.0/centos7.qcow2
+https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse150.x86_64.qcow2
+https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse151.x86_64.qcow2
+ttp://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles15.x86_64.qcow2
```
rjmateus: I think is something missing in these URL. Shouldn't be http://..

salt/mirror/mirror.sh
```diff
@@ -1,28 +1,118 @@
...
+
+function adjust_external_repos() {
+  # external repos are special
+  mkdir -p repo/RPMMD
+  set_error ${?}
+  if [ ! -L repo/RPMMD/SLE-15-GA-Desktop-NVIDIA-Driver ]; then
+    ln -s ../../suse/sle15/ repo/RPMMD/SLE-15-GA-Desktop-NVIDIA-Driver
+  fi
+  set_error ${?}
+  jdupes --linkhard -r -s /srv/mirror/
```
rjmateus: Why are we setting this line in here? is related to external repos?

salt/mirror/mirror.sh
```diff
@@ -1,28 +1,118 @@
...
+  set_error ${?}
+  chmod -R 777 .
```
rjmateus: Why are we setting this line in here? is related to external repos?

salt/mirror/mirror.sh
```diff
@@ -1,28 +1,118 @@
...
+    echo "ERROR: minima configuration file ${1} does not exist!"
+    set_error 1
+    return 1
+  fi
+  minima -c ${1} sync
+  set_error ${?}
+}
+
+function refresh_scc_data() {
+  /root/refresh_scc_data.py "{{salt['pillar.get']('scc:org_user', '')}}:{{salt['pillar.get']('scc:org_pass', '')}}"
```
rjmateus: This can be a dummy question :) 
In the old code user and password were loaded from grains. Now in this new version, they are loaded from pillar data, But I don't see any change in the location they were written. What I'm missing?

salt/mirror/mirror.sh
```diff
@@ -1,28 +1,118 @@
...
+ttp://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp3.x86_64.qcow2
+ttp://download.suse.de/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp4.x86_64.qcow2
+https://github.com/moio/sumaform-images/releases/download/4.4.0/ubuntu1804.qcow2"
+  for REPO in ${REPOS}; do
+    wget --mirror --no-host-directories ${REPO}
+    set_error ${?}
+  done
+}
+
+function adjust_external_repos() {
```
rjmateus: For this function, I think you could look at the solution provided by the cloud team at, which looks more generic (I don't have much knowledge about bach scripting, so I may be wrong)
https://github.com/uyuni-project/sumaform/pull/627/files#diff-b4e829dbcc4f8f0f0e63101d4418ca0bR12

salt/mirror/mirror.sh
```diff
@@ -1,28 +1,118 @@
...
+  echo "Syntax: "
+  echo ""
+  echo "${SCRIPT} <ARGUMENTS>"
+  echo ""
+  echo "Mandatory arguments:"
+  echo ""
+  echo " --mirror-images      Mirror QCOW2 images"
+  echo " --refresh-scc-data   Refresh data from SCC"
+  echo " --apt-mirror         Mirror Apt repositories"
+  echo " --minima-sync=<FILE> Use minima to sync repositories specified on <FILE>"
```
rjmateus: Should we use as default value '/root/minima.yaml' if no yaml file is specified, and avoid allways passing a path to a file? In this way only in the exceptional cases we pass a different value.

salt/mirror/init.sls
```diff
@@ -80,7 +80,7 @@ mirror_script:
     - mode: 755
     - template: jinja
   cron.present:
-    - name: /root/mirror.sh
+    - name: /root/mirror.sh --mirror-images --refresh-scc-data --apt-mirror --minima-sync=/root/minima.yaml &> /var/log/minima.log || cat /var/log/minima.log
```
rjmateus: Why are e calling cat at the end?

salt/mirror/mirror.sh
```diff
@@ -1,28 +1,118 @@
...
+
+ARGS=$(getopt -o h --long help,mirror-images,refresh-scc-data,apt-mirror,minima-sync: -n "${SCRIPT}" -- "$@")
+if [ $? -ne 0 ];
+then
+  print_incorrect_syntax
+  exit 1
+fi
+eval set -- "${ARGS}"
+# extract options and their arguments into variables
+while true; do
```
rjmateus: Make sense to keep the behaviour of the preview script where if I run just 'mirror.sh' and all the steps were executed?

salt/mirror/mirror.sh
juliogonzalez: Right. Typo, with the copy past, as we don't have this at the internal copy of the script. Will fix.

salt/mirror/mirror.sh
juliogonzalez: Not really. The old script was... well, not really that well documented. I guess this part is... removing duplicated and creating links? (this comes from the old script)

If so maybe this need to go to another function.

salt/mirror/mirror.sh
juliogonzalez: Same as before, and in this case I don't even know why we are using 777. So we all users in the system can use this mirror? Is this needed?

salt/mirror/mirror.sh
juliogonzalez: Then at means assuming a hardcoded location for the yaml file... and adding a lot of extra code only for that. I'd rather avoid that in this case.

As you are using salt for sumaform, the new cron call will be deployed and will just work out of the box.

salt/mirror/mirror.sh
juliogonzalez: Not sure what is the relationship. Can you explain?

salt/mirror/mirror.sh
juliogonzalez: Ok... but then having `/root/minima.yaml` is, IMHO, wrong. If you want a default config, it should be at `/etc/` FMPOV.

salt/mirror/mirror.sh
juliogonzalez: Not a dummy question. It seems this was changed for the internal copy. I will need to restore it here.

salt/mirror/init.sls
juliogonzalez: Basically this means:

`Run mirror.sh and if the return code is not 0, print the log"

And inside cron if your command produces **any** output, you will get an email.

The reason it's implemented this way is because, internally, we need to get email when there are errors.

Until now, if you log at your mirror and you check mail for root user, you should get one mail each day no matter if there were errors or not.

salt/mirror/mirror.sh
juliogonzalez: Or otherwise if the config is only per-user, inside the home directory inside a hidden folder.

juliogonzalez: I guess you want this so the user can quickly call the command? If so let's consider moving the config to a right place.

salt/mirror/mirror.sh
juliogonzalez: But I see your point, I guess you want this not to stay compatible with what we had, but to make things easier for users wanting to force a mirror sync from CLI, right?

moio:
Thank you for all this effort @juliogonzalez!

Future note: at least personally I consider the script now right about the complexity where bash stops making sense - I would really love to see this in Python at the next iteration.

I am not against merging this at this point though. Please do not take it personally - I am just kind of allergic to bash script maintenance :wink:

juliogonzalez:
For me, it depends on what a script does.

If it's calling system commands all the time (such as this one is doing), them bash is OK. Reason: just calling a command, getting the return code and evaluating is more complicated on Python.

If the script starts processing data, then it's time to move on to something else like Python. Bash is just horrible for that.

salt/mirror/mirror.sh
juliogonzalez: Moved to its own function.

salt/mirror/mirror.sh
juliogonzalez: Moved to a separate function.

salt/mirror/mirror.sh
juliogonzalez: Done now. Calling without parameters runs `mirror.sh` as before, using a default yaml at `/root/.minima/` directory.

salt/mirror/mirror.sh
juliogonzalez: Moved to a separate  function.

salt/mirror/mirror.sh
juliogonzalez: Fixed.

salt/mirror/mirror.sh
juliogonzalez: Done.

juliogonzalez:
Ready for re-review.

salt/mirror/init.sls
rjmateus: ok, got it. Thanks for the explanation :)

salt/mirror/mirror.sh
rjmateus: For what I can understand the script change from cloud team will work for all sle versions, even SP, and create a link for each version. The difference is the location of the symbolic link ( 'repo/RPMMD' in this case, 'RPMMD' on cloud script)
With this version, it will work for SLE15SP1 and SP2? Or we need to update the script?

juliogonzalez: > With this version, it will work for SLE15SP1 and SP2? Or we need to update the script?

Ah, no I understand. You are talking about https://github.com/uyuni-project/sumaform/pull/627/files#diff-b4e829dbcc4f8f0f0e63101d4418ca0bR11

I guess this won't work for SP1 or SP2. However I'd suggest that since they have that pending on their PR and there's an ongoing discussion about that part of the code, we can leave it to them to fix it later instead of extending the scope of this PR.

rjmateus: Their PR will not be merged since we are refactoring it in a new PR with backend independent modules. I really think this is the right PR to address this since the refactoring that is done in here

juliogonzalez:
After a mumble @rjmateus and I spotted some more problems, so I will improve the PR a little bit more to have an external config file for `mirror.sh` and avoid hardcoded stuff inside the script.

salt/mirror/mirror.sh
juliogonzalez: As agreed: will do it on a separate task and PR, next week.

rjmateus:
Should the documentation be updated with the information about the script options? 
Apart from that, LGTM. Thanks for the excellent contribution :+1:

README_ADVANCED.md
```diff
@@ -133,6 +133,13 @@ module "mirror" {
 
 Note that `mirror` must be populated before any host can be deployed - by default its cache is refreshed nightly via `cron`, you can also schedule a one-time refresh via the `/root/mirror.sh` script.
 
+It is also possible to use parameters with `/root/mirror.sh` to limit what is going to be done.
```
juliogonzalez: Doc added.

PR #699, opened 2020-03-31
Add new sanity check stage into the run-testsuite command

srbarrios:
## What does this PR change?

To support the change made in our YAML files, we need to add the new sanity check stage into the run-testsuite command.
I added it to the sequential and parallel test suite executions, leaving the "essentials" run without it.


srbarrios:
> I suggest to add it also to "essential" runs.

OK, if you prefer, I don't mind :)

PR #700, opened 2020-04-01
Let the resolver work correctly with multicast DNS

Bischoff:
## What does this PR change?

This PR works around this bug: https://bugzilla.suse.com/show_bug.cgi?id=1168306

It sets apparmor so that the resolver can read /etc/mdns.allow.


moio:
Excellent work @Bischoff

Bischoff:
@moio Argh, I did not finish the tests! it will fail !!!

PR #704, opened 2020-04-09
Use a custom version of avahi for SLE 11 as well

Bischoff:
## What does this PR change?

We need a custom repository for our test packages (`orion-dummy` and friends) also for SLE 11.

This PR also fixes a bug: the states file `avahi.sls` is run twice, the first time with `osmajorrelease` grain undefined. This resulted in wrong evaluation of the release the first time.

Fixes: SUSE/spacewalk#11002

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -1,14 +1,16 @@
 include:
   - default.hostname
 
-{% if grains['use_avahi'] %}
+{% if grains['use_avahi'] and grains.get('osmajorrelease', None) != None %}
 
 # TODO: remove the following state when fix to bsc#1163683 is applied to all the SLES versions we use
 {% if grains['osfullname'] == 'SLES' %}
 custom_avahi_repo:
   pkgrepo.managed:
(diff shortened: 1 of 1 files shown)


salt/default/avahi.sls
```diff
@@ -29,7 +31,7 @@ apparmor_allow_avahi_custom_domains:
     - content: "  /etc/mdns.allow         r,"
 {% endif %}
 
-{% if grains['os_family'] == 'RedHat' and grains.get('osmajorrelease', None)|int() == 6 %}
+{% if grains['os_family'] == 'RedHat' and grains['osmajorrelease'] == '6' %}
```
moio, line 34: :man_facepalming:

Bischoff: @moio This was not really an error. Both expressions are equivalent if the grain is defined.

The problem is, the sls file is run twice, and the first time the grain is not defined.

I have moved the test whether the grain is undefined to the top, so we can protect ourselves against this case in *all* the states.

I realize this is a bit not straightforward.

PR #708, opened 2020-05-01
add support for additional network

rjmateus:

## What does this PR change?

https://github.com/SUSE/spacewalk/issues/11396

Add support for the additional network on aws backend.

An additional network is always created. By default, it has the ip mask 172.16.2.0/24.
This network only allows connections between host on the network, this means not in-bound or out-bound communication
The goal of this network configuration is to simulate client scenarios where minions are placed in a totally independent network, only reachable through a susemanager proxy.

Bastion host is connected with two networks: public (which also allow connections to private network) and the additional private network.

The approach followed on libvirt, which only creates the additional network if the field named `additional_network` where defined on the base module **cannot be applied on AWS backend**.

On libvirt, the decision about associate a host with the additional network is made base on the flag `connect_to_additional_network` and field from base module output.
AWS provider don't allow use that value to calculate the count in another resource (in this case secondary network interface). It would force us to first create the base module only, and then start creating the remaining modules.
This is not very user frendly.

### DIFF
--- a/backend_modules/aws/README.md
+++ b/backend_modules/aws/README.md
@@ -4,8 +4,9 @@
 
 Base Module will create:
  - a VPC
- - two subnets
-   - one private, that can only access other hosts in the VPC
+ - three subnets
+   - one private, that can only access other hosts in the VPC and outbound connection to internet
+   - one additional network, private too, that only allows communications between hosts inside the subnet and no outbound connections allowed
    - one public, that can also access the Internet and accepts connections from an IP whitelist
  - security groups, routing tables, Internet gateways, NAT gateway as appropriate
  - one `bastion` host is also created in the public network
@@ -35,15 +36,15 @@ Most modules have configuration settings specific to the AWS backend, those are 
 ### Base Module
 Available provider settings for the base module:
 
-| Variable name      | Type   | Default value | Description                                                              |
-|--------------------|--------|---------------|--------------------------------------------------------------------------|
-| region             | string | `null`        | AWS region where infrastructure will be created                          |
-| availability_zone  | string | `null`        | AWS availability zone inside region                                      |
-| ssh_allowed_ips    | array  | `[]`          | Array of IP's to white list for ssh connection                           |
-| key_name           | string | `null`        | ssh key name in AWS                                                      |
-| key_file           | string | `null`        | ssh key file                                                             |
-| bastion_host       | string | `null`        | bastian host use to connect machines in private network                  |
-| additional_network | string | `null`        | A network mask for PXE tests                                             |
+| Variable name      | Type   | Default value   | Description                                                                                                    |
+|--------------------|--------|-----------------|----------------------------------------------------------------------------------------------------------------|
+| region             | string | `null`          | AWS region where infrastructure will be created                                                                |
+| availability_zone  | string | `null`          | AWS availability zone inside region                                                                            |
+| ssh_allowed_ips    | array  | `[]`            | Array of IP's to white list for ssh connection                                                                 |
+| key_name           | string | `null`          | ssh key name in AWS                                                                                            |
+| key_file           | string | `null`          | ssh key file                                                                                                   |
+| bastion_host       | string | `null`          | bastian host use to connect machines in private network                                                        |
+| additional_network | string | `172.16.2.0/24` | A network mask for the additional network (needs to follow the pattern 172.16.X.Y/24, where X cannot be 0 or 1) |
 
 An example follows:
 ```hcl-terraform
--- a/backend_modules/aws/host/main.tf
+++ b/backend_modules/aws/host/main.tf
@@ -21,10 +21,12 @@ locals {
     contains(var.roles, "virthost") ? { instance_type = "t2.small" } : {},
   var.provider_settings)
 
-  public_subnet_id          = var.base_configuration.public_subnet_id
-  private_subnet_id         = var.base_configuration.private_subnet_id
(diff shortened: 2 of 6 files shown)


rjmateus:
Tests made: network is created and machines are connected to the correct network when bootstrapped

Bischoff:
> Add support for the additional network on aws backend.
> 
> An additional network is always created.

Normally, we should create it only if `private_network =` option is set in main.tf, if we want to be compatible with the libvirt implementation.

> The approach followed on libvirt, which only creates the additional network if the field named `additional_network` where defined on the base module cannot be applied on AWS backend.

OK.

rjmateus:
> > Add support for the additional network on aws backend.
> > An additional network is always created.
> 
> Normally, we should create it only if `private_network =` option is set in main.tf, if we want to be compatible with the libvirt implementation.
> 
In aws I decided to always create the additional network, since depending on the output of a module to connect or not to it is a mess and will be much more harder to use aws backend for the simpler cases.
When creating a VM we can say if we want to connect to additional network or not.

> > The approach followed on libvirt, which only creates the additional network if the field named `additional_network` where defined on the base module cannot be applied on AWS backend.
> 
> OK.
This is not correct. I forget to update the description. The additional network is always created.

Bischoff:
> When creating a VM we can say if we want to connect to additional network or not.

Right. This mechanism already existed, BTW.

backend_modules/aws/README.md
```diff
@@ -4,8 +4,9 @@
 
 Base Module will create:
  - a VPC
- - two subnets
-   - one private, that can only access other hosts in the VPC
+ - three subnets
+   - one private, that can only access other hosts in the VPC and outbound connection to internet
+   - one additional private, that only allow communications between hosts inside the subnet and not outbound connections allowed
```
Bischoff, line 9: to use same terminology: "- one additional network, private too,"

English: "that only allow**s**", "and **no** outbound connection"

backend_modules/aws/README.md
```diff
@@ -35,15 +36,15 @@ Most modules have configuration settings specific to the AWS backend, those are
...
+| region             | string | `null`          | AWS region where infrastructure will be created                                                                |
+| availability_zone  | string | `null`          | AWS availability zone inside region                                                                            |
+| ssh_allowed_ips    | array  | `[]`            | Array of IP's to white list for ssh connection                                                                 |
+| key_name           | string | `null`          | ssh key name in AWS                                                                                            |
+| key_file           | string | `null`          | ssh key file                                                                                                   |
+| bastion_host       | string | `null`          | bastian host use to connect machines in private network                                                        |
+| additional_network | string | `172.16.2.0/24` | A network mask for the additional network (need to follow the pattern 172.16.X.Y/24, where X cannot be 0 or 1) |
```
Bischoff, line 47: "need**s**"

backend_modules/aws/host/main.tf
```diff
@@ -83,6 +85,21 @@ resource "aws_instance" "instance" {
   }
 }
 
+resource "aws_network_interface" "additional_network" {
+  count           = var.connect_to_base_network && var.connect_to_additional_network ? var.quantity : 0
```
Bischoff, line 89: ok

backend_modules/aws/network/main.tf
```diff
@@ -211,16 +231,49 @@ resource "aws_security_group" "private" {
   }
 
   tags = {
-    Name = "${var.name_prefix}-private-security-group"
+    Name = "${var.name_prefix}private-security-group"
+  }
+}
+
+resource "aws_security_group" "private_additional" {
```
Bischoff, line 238: or simply "additional" ?

rjmateus:
> > When creating a VM we can say if we want to connect to additional network or not.
> 
> Right. This mechanism already existed, BTW.

exactly, we are just reusing the existing mechanism

backend_modules/aws/network/main.tf
```diff
@@ -211,16 +231,49 @@ resource "aws_security_group" "private" {
...
+  }
+
+  tags = {
+    Name = "${var.name_prefix}private-additional-security-group"
   }
 }
 
 output "configuration" {
-  depends_on = [aws_route_table_association.private, aws_route_table_association.public]
+  depends_on = [aws_route_table_association.private , aws_route_table_association.public]
```
Bischoff, line 269: space needed here ?

backend_modules/aws/network/variables.tf
```diff
@@ -18,6 +18,11 @@ variable "name_prefix" {
   default     = "sumaform"
 }
 
+variable "additional_network" {
+  description = "Additional network cidr_block bo te created (needs to be in the range of 172.16.x.x)"
```
Bischoff, line 22: CIDR block
of the form

backend_modules/aws/network/main.tf
rjmateus: In this case I don't agree. The network is private and with this name it's explicit. Besides, it's an internal name only, it should be useful for the developer

backend_modules/aws/network/main.tf
rjmateus: Thanks. In fact I was missing the dependency for the private additional network

rjmateus:
@Bischoff thanks for the review. 
@moio I think it is read for review when you have the time.

backend_modules/aws/network/main.tf
Bischoff: ok

PR #709, opened 2020-05-04
use official ubuntu image by default in cucumber test

rjmateus:

## What does this PR change?

Change ubuntu default image used in cucumber test suite to be the official one.

The image packed by sumaform are generating some python/salt error when used with the new salt 3000 version. The errors are related with python 2/3 packing and salt team are aware of it.

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -187,7 +187,7 @@ module "min-ubuntu1804" {
 
   base_configuration = module.base.configuration
   product_version    = var.product_version
-  image              = lookup(local.images, "min-ubuntu1804", "ubuntu1804")
+  image              = lookup(local.images, "min-ubuntu1804", "ubuntu1804o")
   name               = lookup(local.names, "min-ubuntu1804", "min-ubuntu1804")
 
   server_configuration   = local.minimal_configuration
--- a/modules/cucumber_testsuite/variables.tf
+++ b/modules/cucumber_testsuite/variables.tf
@@ -31,7 +31,7 @@ variable "name_prefix" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
-  default     = ["centos7", "opensuse150", "opensuse151", "sles15", "sles15sp1", "sles15sp2", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4", "ubuntu1804"]
+  default     = ["centos7", "opensuse150", "opensuse151", "sles15", "sles15sp1", "sles15sp2", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4", "ubuntu1804o"]
 }
 
 variable "mirror" {


rjmateus:
hum, ok I see. It is also a simple change, but in that case I would like @moio opinion on this one.
We will do it in some point in time, let's see if we should do it just for ubuntu or for all images.

moio:
@Bischoff: in principle `o` images should be used wherever possible for test uses.

In practice I believe it is not wise to just transition them all in one shot, hence the strategy of adding "`o` images" instead of just replacing existing ones, and to move them over one by one.

Another practical issue I see is that we need some agreement with the PO to allocate time (either in Hexagon or the QA Squad) so that some engineer has cycles to discover and fix or drive the fix to any testsuite regression which will predictably result as soon as we switch any image to its `o` counterpart.

Ultimately, I have zero objections in merging this change and any similar changes going forward, let's just be sure we do that in a moment when we have time to address any issues that might consequently arise.

moio:
> This still requires changing all the tf files to download the image.

@rjmateus ideally we should also patch this line so that the right subset of images is pulled when using the testsuite module, and users do not have to worry about picking the right set:

https://github.com/uyuni-project/sumaform/blob/0f6a102e4d4e50f5d4320890b39726ed9877e668/modules/cucumber_testsuite/variables.tf#L34

Bischoff:
> @Bischoff: in principle `o` images should be used wherever possible for test uses.
> 
> In practice I believe it is not wise to just transition them all in one shot, hence the strategy
> of adding "`o` images" instead of just replacing existing ones, and to move them over
> one by one.

OK.

> This still requires changing all the tf files to download the image.
>
> @rjmateus ideally we should also patch this line so that the right subset of images is pulled
> when using the testsuite module, and users do not have to worry about picking the right
> set:

Oh yes! I don't count the times when I had to restart `terraform apply` because I forgot an image when I changed something to `main.tf`. Also, I have lost maybe hours and Gigabytes downloading images I did not really need. If you find a way to generate the images list automatically, that would be great 😄 .

rjmateus:
>@rjmateus ideally we should also patch this line so that the right subset of images is pulled when using the testsuite module, and users do not have to worry about picking the right set:

done

@Bischoff what's your opinion about this? Should it be merged or wait for 4.1 GA, so we can have more time to look at potential problems?

The problem that raises this need was solved with this PR: https://github.com/uyuni-project/sumaform/pull/713

So this mean we don't have any urgency on merging this one.

Bischoff:
> > @rjmateus ideally we should also patch this line so that the right subset of images is pulled when using the testsuite module, and users do not have to worry about picking the right set:
> 
> @moio done
> 
> @Bischoff what's your opinion about this? Should it be merged or wait for 4.1 GA, so we can have more time to look at potential problems?
> 
> The problem that raises this need was solved with this PR: #713
> 
> So this mean we don't have any urgency on merging this one.

@rjmateus The current PR is indeed not urgent now and can be postponed to after 4.1.

The need to specify the list of images is a more general problem and it's a pain in the neck for us sumaform users. But we have been living with it for a long time, so it can also wait.

juliogonzalez:
@rjmateus I'd close this, and would do a single PR to change all images to official images (where possible), after the kickoff.

What do you think?

rjmateus:
I agree. Will close it and then we need to plan with QA when is the best for the change. We can even change it incrementally to measure the impact on the testsuite

PR #710, opened 2020-05-04
Add buildhost VM

calancha:
## What does this PR change?
Add a new VM 'buildhost'

The idea is to have one VM for the 'salt minion' and another VM
for the 'Docker and Kiwi build host' in the Cucumber test suite.

This change adds flexibility to the test suite; by decoupling the 'salt minion'
from the build host we a are able to test more minion versions (not just
those with a build host functionality available).

Another advantage is a reduction in the number of clean up scenarios
(the slow readd entitlements).
See https://github.com/SUSE/spacewalk/issues/10680

- modules/controller/variables.tf
  Add buildhost configuration.

- modules/cucumber_testsuite/main.tf
  Add min-build module section.
  Remove avahi_reflector from the normal minion; keep it only for the min-build.

- salt/controller/bashrc:
  Export environment variable with the build host hostname.

- modules/controller/main.tf
  Read build host configuration.

- README_TESTING.md (Advanced deployments):
  Include the new VM in the example.

### DIFF
--- a/README_TESTING.md
+++ b/README_TESTING.md
@@ -105,6 +105,9 @@ host_settings = {
   min-sles12sp4 = {
     present = true
   }
+  min-build = {
+    present = true
+  }
   minssh-sles12sp4 = {
     present = true
   }
--- a/modules/controller/main.tf
+++ b/modules/controller/main.tf
@@ -36,6 +36,7 @@ module "controller" {
     proxy         = var.proxy_configuration["hostname"]
     client        = length(var.client_configuration["hostnames"]) > 0 ? var.client_configuration["hostnames"][0] : null
     minion        = length(var.minion_configuration["hostnames"]) > 0 ? var.minion_configuration["hostnames"][0] : null
+    build_host    = length(var.buildhost_configuration["hostnames"]) > 0 ? var.buildhost_configuration["hostnames"][0] : null
     centos_minion = length(var.centos_configuration["hostnames"]) > 0 ? var.centos_configuration["hostnames"][0] : null
     ubuntu_minion = length(var.ubuntu_configuration["hostnames"]) > 0 ? var.ubuntu_configuration["hostnames"][0] : null
     ssh_minion    = length(var.sshminion_configuration["hostnames"]) > 0 ? var.sshminion_configuration["hostnames"][0] : null
--- a/modules/controller/variables.tf
+++ b/modules/controller/variables.tf
@@ -54,6 +54,13 @@ variable "minion_configuration" {
   }
 }
 
+variable "buildhost_configuration" {
+  description = "use module.<BUILDHOST_NAME>.configuration, see main.tf.libvirt-testsuite.example"
+  default = {
+    hostnames = []
+  }
+}
+
 variable "sshminion_configuration" {
   description = "use module.<SSHMINION_NAME>.configuration, see main.tf.libvirt-testsuite.example"
   default = {
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -119,10 +119,29 @@ module "min-sles12sp4" {
   auto_connect_to_master  = false
   use_os_released_updates = true
   ssh_key_path            = "./salt/controller/id_rsa.pub"
-  avahi_reflector         = var.avahi_reflector
 
   additional_repos  = lookup(local.additional_repos, "min-sles12sp4", {})
   provider_settings = lookup(local.provider_settings_by_host, "min-sles12sp4", {})
+}
+
+module "min-build" {
+  source = "../minion"
+
+  quantity           = contains(local.hosts, "min-build") ? 1 : 0
+  base_configuration = module.base.configuration
+  product_version    = var.product_version
+  image              = lookup(local.images, "min-build", "sles12sp4")
+  name               = lookup(local.names, "min-build", "min-build")
+
+  server_configuration = local.minimal_configuration
+
+  auto_connect_to_master  = false
+  use_os_released_updates = true
+  ssh_key_path            = "./salt/controller/id_rsa.pub"
+  avahi_reflector         = var.avahi_reflector
+
+  additional_repos  = lookup(local.additional_repos, "min-build", {})
+  provider_settings = lookup(local.provider_settings_by_host, "min-build", {})
 }
 
 module "minssh-sles12sp4" {
@@ -219,6 +238,7 @@ module "ctl" {
   proxy_configuration     = contains(local.hosts, "pxy") ? module.pxy.configuration : { hostname = null }
   client_configuration    = contains(local.hosts, "cli-sles12sp4") ? module.cli-sles12sp4.configuration : { hostnames = [], ids = [] }
   minion_configuration    = contains(local.hosts, "min-sles12sp4") ? module.min-sles12sp4.configuration : { hostnames = [], ids = [] }
+  buildhost_configuration = contains(local.hosts, "min-build") ? module.min-build.configuration : { hostnames = [], ids = [] }
   centos_configuration    = contains(local.hosts, "min-centos7") ? module.min-centos7.configuration : { hostnames = [], ids = [] }
   ubuntu_configuration    = contains(local.hosts, "min-ubuntu1804") ? module.min-ubuntu1804.configuration : { hostnames = [], ids = [] }
   sshminion_configuration = contains(local.hosts, "minssh-sles12sp4") ? module.minssh-sles12sp4.configuration : { hostnames = [], ids = [] }
@@ -245,6 +265,7 @@ output "configuration" {
     pxy = module.pxy.configuration
     cli-sles12sp4 = module.cli-sles12sp4.configuration
     min-sles12sp4 = module.min-sles12sp4.configuration
+    min-build = module.min-build.configuration
     minssh-sles12sp4 = module.minssh-sles12sp4.configuration
     min-centos7 = module.min-centos7.configuration
     min-ubuntu1804 = module.min-ubuntu1804.configuration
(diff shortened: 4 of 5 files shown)


README_TESTING.md
```diff
@@ -28,7 +28,7 @@ module "cucumber_testsuite" {
 }
 ```
 
-This will create a test server, client and minion instances, plus a coordination node called a `controller` which runs the testsuite.
+This will create a test server, client, minion and buildhost instances, plus a coordination node called a `controller` which runs the testsuite.
```
Bischoff, line 31: ", and"

The build host is a client, after all - maybe we don't need to specify here.

README_TESTING.md
```diff
@@ -105,6 +105,9 @@ host_settings = {
   min-sles12sp4 = {
     present = true
   }
+  bldh-sles12sp4 = {
```
Bischoff, line 108: `min-build` instead of `bldh-sles12sp4`, like `min-kvm` below.

As a side note, `sles12sp4` is an error, our minions have no reason to be SLES 12 SP4. @rjmateus is going to rectify these names introduced during the refactoring.

README_TESTING.md
```diff
@@ -123,7 +126,7 @@ host_settings = {
 }
 ```
 
-The default value for host_settings block has a SLES12SP4 Traditional Client and SLES12SP4 Minion present:
+The default value for host_settings block has a SLES12SP4 Traditional Client, a SLES12SP4 Minion and a SLES12SP4 Docker/Kiwi present:
```
Bischoff, line 129: No, I think the build host should be optional.

BTW, according to `modules/controller/main.tf`below, you made it optional.

README_TESTING.md
```diff
@@ -133,6 +136,9 @@ host_settings = {
   min-sles12sp4 = {
     present = true
   }
+  bldh-sles12sp4 = {
```
Bischoff, line 139: see above about its name

modules/controller/main.tf
```diff
@@ -36,6 +36,7 @@ module "controller" {
     proxy         = var.proxy_configuration["hostname"]
     client        = length(var.client_configuration["hostnames"]) > 0 ? var.client_configuration["hostnames"][0] : null
     minion        = length(var.minion_configuration["hostnames"]) > 0 ? var.minion_configuration["hostnames"][0] : null
+    buildhost     = length(var.buildhost_configuration["hostnames"]) > 0 ? var.buildhost_configuration["hostnames"][0] : null
```
Bischoff, line 39: by analogy with `kvm_host` below, I would rather use `build_host` with an underscore.

modules/controller/variables.tf
```diff
@@ -54,6 +54,14 @@ variable "minion_configuration" {
   }
 }
 
+variable "buildhost_configuration" {
```
Bischoff, line 57: yes

modules/controller/variables.tf
```diff
@@ -54,6 +54,14 @@ variable "minion_configuration" {
...
+  description = "use module.<BUILDHOST_NAME>.configuration, see main.tf.libvirt-testsuite.example"
+  default = {
+    ids       = []
```
Bischoff, line 60: do we need the ids ? I don't think so

modules/cucumber_testsuite/main.tf
```diff
@@ -125,6 +125,26 @@ module "min-sles12sp4" {
   provider_settings = lookup(local.provider_settings_by_host, "min-sles12sp4", {})
 }
 
+module "bldh-sles12sp4" {
```
Bischoff, line 128: see above about name

modules/cucumber_testsuite/main.tf
```diff
@@ -125,6 +125,26 @@ module "min-sles12sp4" {
...
+  source = "../minion"
+
+  quantity           = contains(local.hosts, "bldh-sles12sp4") ? 1 : 0
```
Bischoff, line 131: same

modules/cucumber_testsuite/main.tf
```diff
@@ -125,6 +125,26 @@ module "min-sles12sp4" {
...
+  base_configuration = module.base.configuration
+  product_version    = var.product_version
+  image              = lookup(local.images, "bldh-sles12sp4", "sles12sp4")
```
Bischoff, line 134: same

modules/cucumber_testsuite/main.tf
```diff
@@ -125,6 +125,26 @@ module "min-sles12sp4" {
...
+  name               = lookup(local.names, "bldh-sles12sp4", "bldh-sles12")
```
Bischoff, line 135: same

modules/cucumber_testsuite/main.tf
```diff
@@ -125,6 +125,26 @@ module "min-sles12sp4" {
...
+
+  server_configuration = local.minimal_configuration
+
+  auto_connect_to_master  = false
+  use_os_released_updates = true
+  ssh_key_path            = "./salt/controller/id_rsa.pub"
+  avahi_reflector         = var.avahi_reflector
```
Bischoff, line 142: right - please remove the `avahi_reflector` declaration from the normal minion declaration. It will be needed only for the build host.

modules/cucumber_testsuite/main.tf
```diff
@@ -219,6 +239,7 @@ module "ctl" {
   proxy_configuration     = contains(local.hosts, "pxy") ? module.pxy.configuration : { hostname = null }
   client_configuration    = contains(local.hosts, "cli-sles12sp4") ? module.cli-sles12sp4.configuration : { hostnames = [], ids = [] }
   minion_configuration    = contains(local.hosts, "min-sles12sp4") ? module.min-sles12sp4.configuration : { hostnames = [], ids = [] }
+  buildhost_configuration = contains(local.hosts, "bldh-sles12sp4") ? module.bldh-sles12sp4.configuration : { hostnames = [], ids = [] }
```
Bischoff, line 242: see remarks about names.

modules/cucumber_testsuite/main.tf
```diff
@@ -245,6 +266,7 @@ output "configuration" {
     pxy = module.pxy.configuration
     cli-sles12sp4 = module.cli-sles12sp4.configuration
     min-sles12sp4 = module.min-sles12sp4.configuration
+    bldh-sles12sp4 = module.bldh-sles12sp4.configuration
```
Bischoff, line 269: same

modules/cucumber_testsuite/variables.tf
```diff
@@ -59,6 +59,9 @@ variable "host_settings" {
     min-sles12sp4 = {
       present = true
     }
+    bldh-sles12sp4 = {
```
Bischoff, line 62: same

salt/controller/bashrc
```diff
@@ -5,6 +5,7 @@ export SERVER="{{ grains.get('server') }}"
 {% if grains.get('proxy') | default(false, true) %}export PROXY="{{ grains.get('proxy') }}" {% else %}# no proxy defined {% endif %}
 {% if grains.get('client') | default(false, true) %}export CLIENT="{{ grains.get('client') }}"{% else %}# no client defined {% endif %}
 {% if grains.get('minion') | default(false, true) %}export MINION="{{ grains.get('minion') }}"{% else %}# no minion defined {% endif %}
+{% if grains.get('buildhost') | default(false, true) %}export BUILDHOST="{{ grains.get('buildhost') }}"{% else %}# no buildhost defined {% endif %}
```
Bischoff, line 8: let's continue copying `kvm_host` line below:
```
{% if grains.get('build_host') | default(false, true) %}export BUILD_HOST="{{ 
  grains.get('build_host') }}"{% else %}# no build host defined {% endif %}
```

with underscores + space

salt/top.sls
```diff
@@ -18,6 +18,10 @@ base:
     - match: grain
     - minion
 
+  'roles:buildhost':
```
Bischoff, line 21: ok

calancha:
> could you please do a test to see if it works with a minion, but no build host? this machine should be optional.

Addressed the comments.  Now, testing if works as expected.

calancha:
> > could you please do a test to see if it works with a minion, but no build host? this machine should be optional.
> 
> Addressed the comments. Now, testing if works as expected.

OK, it works as expected.

modules/cucumber_testsuite/main.tf
```diff
@@ -125,6 +125,26 @@ module "min-sles12sp4" {
...
 }
 
+module "min-build" {
+  source = "../minion"
+
+  quantity           = contains(local.hosts, "min-build") ? 1 : 0
+  base_configuration = module.base.configuration
+  product_version    = var.product_version
+  image              = lookup(local.images, "min-build", "sles12sp4")
+  name               = lookup(local.names, "min-build", "bldh-sles12")
```
rjmateus, line 135: name should be consistent with module name. Change to "min-build"

salt/top.sls
```diff
@@ -18,6 +18,10 @@ base:
...
+    - match: grain
+    - buildhost
```
rjmateus, line 23: If I understand correctly if machine have role `buildhost` the states located in folder `buildhost` will be applied. I don't any place setting this role neither the needed salt states to be apply.

calancha: Good catch.  That's not needed.  Deleted it.

modules/cucumber_testsuite/main.tf
Bischoff: not addressed yet

modules/controller/variables.tf
Bischoff: not addressed yet

calancha:
Got the OK from Moio at rockechat to merge, thus I am merging it.

PR #712, opened 2020-05-11
AWS: add image support for RHEL 7/8

rjmateus:

## What does this PR change?

Add support for rhel 8 and rhel 7 images on aws.
At the moment this PR is using community images provided by red Hat.

https://github.com/SUSE/spacewalk/issues/11453

### DIFF
--- a/backend_modules/aws/base/main.tf
+++ b/backend_modules/aws/base/main.tf
@@ -288,6 +288,48 @@ data "aws_ami" "ubuntu1604" {
   }
 }
 
+data "aws_ami" "rhel8" {
+  most_recent = true
+  name_regex  = "^RHEL-8.2.0_HVM-"
+  owners      = ["309956199498"]
+
+  filter {
+    name   = "architecture"
+    values = ["x86_64"]
+  }
+
+  filter {
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  filter {
+    name   = "root-device-type"
+    values = ["ebs"]
+  }
+}
+
+data "aws_ami" "rhel7" {
+  most_recent = true
+  name_regex  = "^RHEL-7.7_HVM-"
+  owners      = ["309956199498"]
+
+  filter {
+    name   = "architecture"
+    values = ["x86_64"]
+  }
+
+  filter {
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  filter {
+    name   = "root-device-type"
+    values = ["ebs"]
+  }
+}
+
 module "network" {
   source = "../network"
 
@@ -333,6 +375,8 @@ locals {
       centos6     = { ami = data.aws_ami.centos6.image_id, ssh_user = "centos" },
       ubuntu1804  = { ami = data.aws_ami.ubuntu1804.image_id, ssh_user = "ubuntu" },
       ubuntu1604  = { ami = data.aws_ami.ubuntu1604.image_id, ssh_user = "ubuntu" },
+      rhel8       = { ami = data.aws_ami.rhel8.image_id},
+      rhel7       = { ami = data.aws_ami.rhel7.image_id},
     }
     },
     local.create_network ? module.network.configuration : {
--- a/backend_modules/aws/host/user_data.yaml
+++ b/backend_modules/aws/host/user_data.yaml
@@ -76,7 +76,20 @@ runcmd:
   - zypper removerepo --all
   - rm /etc/modprobe.d/50-ipv6.conf
 %{ endif }
-%{ if image == "centos7" }
+%{ if image == "rhel8"}
+yum_repos:
+  # repo for salt
+  temp_tools_pool_repo:
+    baseurl: http://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/CentOS8-Uyuni-Client-Tools/CentOS_8/
+    failovermethod: priority
+    enabled: true
+    gpgcheck: false
+    name: temp_tools_pool_repo
+    priority: 98
+
+packages: ["salt-minion"]
+%{ endif }
+%{ if image == "centos7" || image == "rhel7"}
 yum_repos:
   # repo for salt
   temp_tools_pool_repo:


backend_modules/aws/host/user_data.yaml
```diff
@@ -76,7 +76,20 @@ runcmd:
   - zypper removerepo --all
   - rm /etc/modprobe.d/50-ipv6.conf
 %{ endif }
-%{ if image == "centos7" }
+%{ if image == "rhel8"}
+yum_repos:
+  # repo for salt
+  temp_tools_pool_repo:
+    baseurl: http://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/CentOS8-Uyuni-Client-Tools/CentOS_8/
```
moio, line 83: I'd suggest using RES tools instead of CentOS tools, even if that implies having a mirror.

rjmateus: Ok, sounds good. 
I was looking at repositories in mirror configuration and we will need to be updated it to download tools for RES 8 (I will include in this PR, in separated commit)

@juliogonzalez I was looking at ibs and looks like we are missing tools for RES8. IIRC, 3.2 will not support RES8 clients. If it's not the case, the repo is missing

juliogonzalez: Two things:

- Uyuni CentOS tools for RHEL, CentOS and Oracle on Uyuni.
- SUSE Manager RES tools for RES, RHEL, CentOS and Oracle on SUSE Manager.

So we need to support both.

> @juliogonzalez I was looking at ibs and looks like we are missing tools for RES8. IIRC, 3.2 will not support RES8 clients. If it's not the case, the repo is missing

3.2 will not support RES8, that's right. 

But please, remember that the client tools are common for all supported SUSE Manager versions (exception: development).

That means that, as of today:

## SUSE manager
- The Released Client tools on SCC are used for 4.0 and 3.2 released versions of SUSE Manager.
- 4.0 client tools from D:G:M:4.0 are used for 3.2 and 4.0 on development environments (except HEAD) and on top of the released client tools.
- Head client tools from D:G:M:Head are used for Head development environments. and from time to time also to test them together with 3.2 development and 4.0 development (reason: Head will become stable at some point, so 3.2 and 4.0 need to be able to work wi them).

## Uyuni
- The "Stable" Uyuni client are used for Uyuni Stable
- The "Master" Uyuni clients are used for Uyuni Master (on top of the Uyuni "Stable" client tools).

juliogonzalez: The way I see it the title and description of this PR is incomplete. What clients? Uyuni Clients? SUSE Manager clients? Both?

As I said above, both Uyuni and SUSE Manager support RHEL.

moio: OK, let me clarify two aspects: first, the title of the PR should be "add support for RHEL 7 and 8 images". Images are not clients, although currently we have no other use for RHEL images.

Second: the Salt version specified here is the one that will be upgraded during first apply to the actual Salt version we need. It is included here, in the cloud-init configuration, only because the target image does not ship with one, however old - but indeed we can in principle use any version, even upstream - the only requirement is that it can be upgraded (or downgraded) to the target Salt version at apply time.

So in a way, using CentOS tools is OK and expected to work. In case of RHEL clients of SUSE Manager, it will be replaced by the SUSE RES client tools at first apply time.

From another angle, upgrading Salt with Salt is always a risk, and we have seen brittle cases here and there, so using the same "version" of the tools should make that a bit more robust. I did overlook Uyuni use cases though when first recommending otherwise.

So maybe the right thing is really to use the CentOS tools unless we find a case that is so broken, we are really forced to use the RES ones. Maybe we are lucky and never encounter one, either.

juliogonzalez: > From another angle, upgrading Salt with Salt is always a risk
It is, indeed. IIUC another issue we have, it seems we to be a nice problem on Ubuntu because of that.

> So maybe the right thing is really to use the CentOS tools unless we find a case that is so broken, we are really forced to use the RES ones. Maybe we are lucky and never encounter one, either.

Problem with that is that then... what will happen to Uyuni developers that do are not SUSE employees? If that happens I guess we'll need to use one or other repository depending on the product that sumaform is going to deploy.

But as you said... maybe we will never need it. So for now using the Uyuni client tools for CentOS is fine.

paususe:
@rjmateus 

Could you please give me the links to the images? From the id's in the commits I am unable to find them.

I need to check if those are the images we want to work with.

rjmateus:
Hey @paususe 
The image AMI is computed dynamically. If we execute now, the ami's in use will be:
RHEL7: ami-0dc7b4dac85c15019
RHEL8: ami-09e973def6bd1ad96

rjmateus:
@moio @juliogonzalez do you think this PR is ready to merge as is or should I change the repository in use to point to a mirror instead? 
Force using a mirror may not work in uyuni.

juliogonzalez:
FMPOV we need this to work on Uyuni as well, so if @paususe  approves the images (not in the sense of the image IDs, but in the sense of the EULA for the images we use from AWS even if the ID changes because they get an update), I'd say we can merge.

rjmateus:
@juliogonzalez @moio I just talked with @paususe and we need ok from legal before merging any support for rhel images

backend_modules/aws/base/main.tf
```diff
@@ -288,6 +288,48 @@ data "aws_ami" "ubuntu1604" {
...
+  }
+
+  filter {
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  filter {
+    name   = "root-device-type"
+    values = ["ebs"]
```
juliogonzalez, line 308: Because we can't use instances with only ephemeral, right? :-D

Those cost less money.

backend_modules/aws/host/user_data.yaml
```diff
@@ -76,7 +76,20 @@ runcmd:
   - zypper removerepo --all
   - rm /etc/modprobe.d/50-ipv6.conf
 %{ endif }
-%{ if image == "centos7" }
+%{ if image == "rhel8"}
```
juliogonzalez, line 79: Why not:
```
%{ if image == "centos8" || image == "rhel8"}
```
?

paususe: AFAIK there is no official centos8 image

juliogonzalez: Right, Ricardo told me. So as you prefer, but if having centos8 here doesn't break anything, I'd add it already.

rjmateus: Pau is correct, there is no centos8 image available yet. I would prefer to add it after because we are not sure at this point if the configuration for centos8  will be the same as rhel8 (it should be, but until test it we are not sure)

backend_modules/aws/base/main.tf
rjmateus: Quick answer, most images only have support for EBS :)
But IIRC instance store will lose all data if the instance gets stoped. So if you restart your instance all data we have there (provisioning included) will be lots, and IMHO is not what we want

backend_modules/aws/host/user_data.yaml
juliogonzalez: No problem from my side.

backend_modules/aws/base/main.tf
juliogonzalez: Only if you do a cold restart (stop and start): https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html

Could be useful for clients, but since we planned to use small instances... they don't have ephemeral.

Consider it, however for the server (if available). I don't recall the instance type we selected, but most probably the type doesn't have enough for everything, but it could be used for part of SUSE Manager/Uyuni, so we pay less for the disk.

For CI instances using ephemerals is perfectly fine. They don't fail so frequently, and if they do one time... well... wait for the next build :-D

But of course, nothing for this PR.

rjmateus: OK, sounds good.
But from a quick search on AMI's page, it doesn't have any supporting Instance Store as root device. 
I think if we want to use it, we need to upload our own images, and that is something we don't want

juliogonzalez: > I think if we want to use it, we need to upload our own images, and that is something we don't want

Couldn't agree more. If none of the official images for the operating systems we use (SLE/openSUSE, CentOS/RHEL and Ubuntu) support it as root disk, then that's how it is.

Still, for the server, keep in mind that if the instance type you use has ephemerals, they are available to be used as non-root disks. I remember than when I created such instances with CloudFormation, I had to add them specifically to make them available (not required if you start the instance via AWS WebUI). No idea of how that works with terraform.

PR #713, opened 2020-05-11
Ubuntu 1804 unable to get apt lock when updating salt

rjmateus:

## What does this PR change?
https://github.com/SUSE/spacewalk/issues/11454

After stop all apt services make sure lock is released before continue to apt update steps.

The code checks if the file don't exist or if exist it is not locked by another process.

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -461,6 +461,14 @@ disable_apt_daily_upgrade_service:
     - name: apt-daily-upgrade.service
     - enable: False
 
+wait_until_apt_lock_file_unlock:
+  cmd.run:
+    - name: "test ! -f /var/lib/apt/lists/lock || ! lsof /var/lib/apt/lists/lock"
+    - retry:
+      - attempts: 10
+      - interval: 5
+      - until: True
+
 tools_update_repo:
   pkgrepo.managed:
     - file: /etc/apt/sources.list.d/Ubuntu{{ short_release }}-Client-Tools.list


salt/repos/default.sls
```diff
@@ -461,6 +461,15 @@ disable_apt_daily_upgrade_service:
     - name: apt-daily-upgrade.service
     - enable: False
 
+disable_unattended_upgrades_service:
+  service.dead:
+    - name: unattended-upgrades.service
+    - enable: False
+
+remove_apt_lock_file:
```
juliogonzalez, line 469: Why do we need this?

If the file exists because a program is running, we should not remove it. We should stop the program itself.

salt/repos/default.sls
```diff
@@ -461,6 +461,15 @@ disable_apt_daily_upgrade_service:
...
+  file.absent:
+    - name: /var/lib/apt/lists/lock
+
```
meaksh, line 472: As @juliogonzalez mentioned, we should no simply remove the lock if there is actually an `apt` installed being executed. I would suggest the following state instead, that should wait until the file is absent but does not remove it:

```suggestion
wait_until_apt_lock_file_absent:
  cmd.run:
    - name: test ! -f /var/lib/apt/lists/lock
    - retry:
      - attempts: 15
      - interval: 5
      - until: True

```

salt/repos/default.sls
moio: Do we know what that program is, if any exists?

rjmateus: I'm doing a new run now, to try to find out the process that could be locking. If we remove the file, it will work, but I agree is not the best way to do it. Tomorrow I will try Pablo's approach.

juliogonzalez: An explanation of why removing the lock is a bad idea:

If we remove the file, it will work, but it can break a lot of things.

That file is there and locked to avoid two instances of the package manager (in fact IIRC the file is locked by another `apt` call) running at the same time. Otherwise it can mean two instances of the package manager removing the same files, changing the same files, one removing while other is adding something... And that can provoke a mess. zypper does basically the same thing.

package manager locks should never be removed except as last resort, and followed by a research of what happened.

Removing a package manager lock file is only acceptable if there are **no** processes using it. And if no process is using it, it means the process crashed, which is not good.

So:

a) We need to find the process that is crashing, and disable the start.
or
b) We just need to find the process creating the lock, and disable the start.

For example, this is from one of my debian servers, with one console running an `apt-get install postgresql` (and waiting), and another for a `lsof`:
```
# lsof /var/cache/apt/archives/lock
COMMAND   PID USER   FD   TYPE DEVICE SIZE/OFF    NODE NAME
apt-get 20385 root    6uW  REG  253,3        0 6684939 /var/cache/apt/archives/lock
```
If I now try to install something else without cancelling the previous call to `apt-get`.
```
# apt-get install lighthttpd
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```
This is not the same lock where you have the issue, but IIRC different `apt-*`/`apt` commands or subcommands will lock different files (and IIRC even different stages for a cummand/subcommand execition).

In short, yes: we need to find where the problem is.

And AFAIK what I said about the package manager locks is basically applicable to any application using file locks.

rjmateus: I totally agree with what you exposed :)
This PR is a draft and my goal whit the lock remove was made sure I'm on the right path.
I have made some changes to try to find out which process is running. I already disabled one, but it looks like it's not enough.

juliogonzalez: I'd not consider myself a Debian/Ubuntu guru (I am specially a noob when it comes to packaging DEB), but I am using Debian it since 3.0 Woody (for years even just SID for the desktop), so if you need help... ping and I will see if I can help :-D

rjmateus:
It looks to deploy successfully: https://ci.suse.de/view/Manager/view/Manager-4.0/job/manager-4.0-cucumber-PRV/91/

It could the combination of one of the following:
- Service.masked in services  apt-daily-upgrade.service and apt-daily.service
- Usage of sig field on disable_apt_daily_service state

More tests tomorrow to make sure what's happening

salt/repos/default.sls
```diff
@@ -451,15 +451,42 @@ disable_apt_daily_upgrade_timer:
     - name: apt-daily-upgrade.timer
     - enable: False
 
+masked_apt_daily_service:
+  service.masked:
+    - name: apt-daily.service
```
juliogonzalez, line 456: OMG... Even Debian has this enabled all this crap on boot? https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=844453

I guess I never noticed on my serves as they are upgraded from at least from Debian 7 or 8, and I don't have such services, so at least I don't get this launched on boot.

However I have timers that it seems that luckily are NOT running at the same time `cron-apt` is running.

So yes @rjmateus it seems the services are the problem.But I agree with the general change, also disabling timers.

FMPOV we should see if we can disable this not from salt, but when booting the image, so this thing is never, ever running. Maybe https://unix.stackexchange.com/a/471192/234557 could do the trick?

Now I have an interesting question: Are we disabling all this crap when we onboard the the minions on Uyuni/SUSE Manager **without** sumaform?

rjmateus: >OMG... Even Debian has all this crap enabled on boot? https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=844453
yeah, I notice that last night :)

And when it fails, tthe service `apt-daily-upgrade.service` toke 30 seconds on startup.

>So yes @rjmateus it seems the services are the problem.But I agree with the general change, also disabling timers.

If I add a 20 seconds sleep after disabling the services it works. 
@meaksh I think the problem could be the salt state is not correctly detecting is the service has died. 
My suspicion is that `apt-daily-upgrade.service` have a pre-init step command, and that is the one which keeps running for some more time

>Now I have an interesting question: Are we disabling all this crap when we onboard the minions on Uyuni/SUSE Manager without sumaform?

I don't know, but I would bet they keep running. Should ping someone about this? I thing Matei is the expert on debian/ubuntu onboard, this right?

juliogonzalez: > I don't know, but I would bet they keep running. Should ping someone about this? I thing Matei is the expert on debian/ubuntu onboard, this right?

I will have a quick look and will ping someone if needed. IIRC either Matei or Michele can handle this, if it's really a problem.

salt/repos/default.sls
meaksh: >
> @meaksh even your suggestion is dangerous :-)

I don't get why my suggestion is dangerous, maybe I'm missing something :)

The state I suggested never removes a lock, it just wait until the lock is removed by the `apt` instance that owns it. Then make Salt to continue.

juliogonzalez: Mmmm, maybe I didn't really understood the sate?

What happens if the lock never gets removed?

meaksh: @juliogonzalez if that's the case (the lock is never removed), the state will fail after 15 attempts (every 5 seconds) of checking if the file is gone. And that would mean we have a problem with that image because it contains (or produce) an state lock apt file.

rjmateus: The behavior we are facing is related to the `apt-daily.service`. After it gets stopped looks like a process is getting some more time to really stop. If I add a sleep after calling the dead state on the service the high state works.
I will add a state after the service dead to insure look gets released, and see if it solves the problem.

rjmateus:
@juliogonzalez @meaksh is ready for review. It checks if the files don't exist or if it exists it is not locked by another process.
Run started using this code: https://ci.suse.de/view/Manager/view/Manager-4.0/job/manager-4.0-cucumber-PRV/96/

salt/repos/default.sls
juliogonzalez: Done now, with a lsof. Resolving.

salt/repos/default.sls
```diff
@@ -461,6 +461,14 @@ disable_apt_daily_upgrade_service:
     - name: apt-daily-upgrade.service
     - enable: False
 
+wait_until_apt_lock_file_unlock:
+  cmd.run:
+    - name: "test ! -f /var/lib/apt/lists/lock || ! lsof /var/lib/apt/lists/lock"
```
juliogonzalez, line 466: Only comment from my side: did you try this with the sumaform images or with the official Ubuntu images?

I am asking because IIRC a minimal debian instalation doesn't ship `lsof` (we need to keep it in mind for the future, as we'd a similar code), and in that case we could have problems.

juliogonzalez: Reason I think it's important to verify. This is from a Debian 10 cloud instance I have at OVH:

```
# lsof
-bash: lsof: command not found
# fuser
-bash: fuser: command not found
# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
But if the Ubuntu cloud images have lsof available, then the PR is good to go for now :-)

rjmateus: I have tested all libvirt and aws images we are using to deploy ubuntu and it works, all of them have the lsof installed.
I will now merge this PR

rjmateus:
Merging this PR. Thanks all for the help

PR #716, opened 2020-05-16
specify an alternative container registry (without auth) for the testsuite

mcalmer:
## What does this PR change?

configure a different container registry for the testsuite 

Part of a series of PRs in different repos:

https://github.com/uyuni-project/sumaform/pull/716
https://github.com/SUSE/susemanager-ci/pull/78
https://github.com/uyuni-project/uyuni/pull/2224

### DIFF
--- a/README_TESTING.md
+++ b/README_TESTING.md
@@ -190,14 +190,14 @@ module "cucumber_testsuite" {
 
 ## Alternative Container Registry server
 
-If you want the test suite to use an unauthenticated container registry server, you can specify it with the `registry_uri` variable.
+If you want the test suite to use an unauthenticated container registry server, you can specify it with the `no_auth_registry` variable.
 
 Example:
 
 ```hcl
 module "cucumber_testsuite" {
    ...
-   registry_uri = "uri.of.registry:443/used"
+   no_auth_registry = "uri.of.registry:443/used"
    ...
 }
 ```
--- a/README_TESTING.md
+++ b/README_TESTING.md
@@ -190,7 +190,7 @@ module "cucumber_testsuite" {
 
 ## Alternative Container Registry server
 
-If you want the test suite to use a unauthenticated container registry server, you can specufy it with the `registry_uri` variable.
+If you want the test suite to use an unauthenticated container registry server, you can specify it with the `registry_uri` variable.
 
 Example:
 
@@ -204,7 +204,7 @@ module "cucumber_testsuite" {
 
 ## Alternative Portus server
 
-If you want the test suite to use a Portus server, you can specify it with the `portus_uri`, `portus_username`, and `portus_password` variables.
+If you want the test suite to use a Portus server (authenticated registry), you can specify it with the `portus_uri`, `portus_username`, and `portus_password` variables.
 
 Example:
 
--- a/README_TESTING.md
+++ b/README_TESTING.md
@@ -188,6 +188,20 @@ module "cucumber_testsuite" {
 }
 ```
 
+## Alternative Container Registry server
+
+If you want the test suite to use a unauthenticated container registry server, you can specufy it with the `registry_uri` variable.
+
+Example:
+
+```hcl
+module "cucumber_testsuite" {
+   ...
+   registry_uri = "uri.of.registry:443/used"
+   ...
+}
+```
+
 ## Alternative Portus server
 
 If you want the test suite to use a Portus server, you can specify it with the `portus_uri`, `portus_username`, and `portus_password` variables.
--- a/README_TESTING.md
+++ b/README_TESTING.md
@@ -204,16 +204,16 @@ module "cucumber_testsuite" {
 
 ## Alternative Portus server
 
-If you want the test suite to use a Portus server (authenticated registry), you can specify it with the `portus_uri`, `portus_username`, and `portus_password` variables.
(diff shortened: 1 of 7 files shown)


README_TESTING.md
```diff
@@ -197,6 +197,20 @@ module "cucumber_testsuite" {
 }
 ```
 
+## Alternative Container Registry server
+
+If you want the test suite to use a unauthenticated container registry server, you can specufy it with the `registry_uri` variable.
```
rjmateus, line 202:
```suggestion
If you want the test suite to use an unauthenticated container registry server, you can specify it with the `registry_uri` variable.
```
In this case, we want to say user can specify an unauthenticated container registry or it can set an alternative one, with or without authentication?

Bischoff: This is "alternative" because there is a default server. I therefore suggest to add here:

     "If you don't specify an alternative server, it will default to `registry.mgr.suse.de`."

Note: we could have chosen as an alternative to make this server optional. In that case, the corresponding Docker tests would have been skipped in the test suite. I'm not saying this would have been a better solution, I'm just saying that would have been possible too :smile_cat: .

salt/controller/bashrc
```diff
@@ -48,6 +48,7 @@ export GITPROFILES="{{ grains.get('git_profiles_repo') }}"
 export SCC_CREDENTIALS="{{ grains.get('cc_username') }}|{{ grains.get('cc_password') }}"
 export PORTUS_URI="{{ grains.get('portus_uri') }}"
 export PORTUS_CREDENTIALS="{{ grains.get('portus_username') }}|{{ grains.get('portus_password') }}"
+export REGISTRY_URI="{{ grains.get('registry_uri') | default('registry.mgr.suse.de', true) }}"
```
Bischoff, line 51: Instead of a default value, we could accept this variable to be undefined. In that case, we would still be able to run the test suite, only without registry tests.

I kind of prefer this approach, it's more "variable geometry" for the test suite. It enables to run it if e.g. the registry servers are broken, under maintenance, etc.

juliogonzalez: I have a crazy idea about this, to avoid hardcoding the registry hostname on the Docker files.

It's not clean, but I think it's better than duplicating all the stuff: Instead of just exporting the URI, we could add a hostname, something like `registry.localnet` to `/etc/hosts` that perform the resolution.

README_TESTING.md
mcalmer: test fixed and for now the registry is optional

salt/controller/bashrc
mcalmer: > something like registry.localnet to /etc/hosts

first stumbling block: we need to setup SSL certificates which provide this name for all mirrors.

juliogonzalez: @mcalmer well, here's another suggestion then:

1. Our servers already trust registry.mgr.suse.de, so I guess they trust the CA used to create its certificate.
2. Create fake-registry.mgr.suse.de on our DNS, point it to 127.0.0.1 and generate a private key and SSL cert.
3. Add the private key and SSL cert to the mirrors, for the registry part.
4. Apply the change to `/etc/hosts` so fake-registry.mgr.suse.de points to the mirror.

moio:
@juliogonzalez we have so far resisted the temptation of fiddling with `/etc/hosts`, and I would recommend if at all possible to try and maintain this "tradition".

juliogonzalez:
@moio everything has pros and cons.

The cons of not doing it, is that we'll need to maintain a lot of duplicated info, and FMPOV it will do maintenance harder.

But not my call, I will leave the decision to @mcalmer

Bischoff:
+1 with @moio. Fiddling with `/etc/hosts` is almost always a bad idea.

salt/controller/bashrc
mcalmer: all suggestens rejected as discussed in our meetings

salt/controller/bashrc
```diff
@@ -48,8 +48,9 @@ export VIRTHOST_XEN_PASSWORD="linux" {% else %}# no Xen host defined {% endif %}
 # various test suite settings
 export GITPROFILES="{{ grains.get('git_profiles_repo') }}"
 export SCC_CREDENTIALS="{{ grains.get('cc_username') }}|{{ grains.get('cc_password') }}"
-export PORTUS_URI="{{ grains.get('portus_uri') }}"
-export PORTUS_CREDENTIALS="{{ grains.get('portus_username') }}|{{ grains.get('portus_password') }}"
+export PORTUS_CREDENTIALS="{{ grains.get('auth_registry_username') }}|{{ grains.get('auth_registry_password') }}"
+{% if grains.get('auth_registry') | default(false, true) %}export PORTUS_URI="{{ grains.get('auth_registry') }}" {% else %}# no authenticated registry defined {% endif %}
```
Bischoff: Why not rename this variable AUTH_REGISTERY to be consistent everywhere ?

Bischoff (commented):
See question

salt/controller/bashrc
mcalmer: You commented after I have fixed it already :-)

Bischoff (approved):
this time i could not find anything problematic; thanks for the changes

salt/controller/bashrc
Bischoff: hehe :-)

PR #718, opened 2020-05-25
Update instruction to run single Cucumber feature

moio:
## What does this PR change?

It updates the README_TESTING.md instruction as the current one will not work.

### DIFF
--- a/README_TESTING.md
+++ b/README_TESTING.md
@@ -66,7 +66,7 @@ To enable/disable features, edit these YAML files. Keep in mind that:
 
 Once all `core_` features have been executed you can run a non-core Cucumber feature as follows:
 ```
-ssh -t head-ctl.tf.local cucumber spacewalk/testsuite/features/my_feature.feature
+ssh -t head-ctl.tf.local cucumber -r spacewalk/testsuite/features spacewalk/testsuite/features/secondary/my_feature.feature
 ```
 
 Read HTML results at:


moio:
@Bischoff, @srbarrios : my understanding is that the splitting of features into the `core/` and `secondary/` subdirectories unexpectedly broke this Cucumber convention. From `cucumber --help`:

```
Usage: cucumber [options] [ [FILE|DIR|URL][:LINE[:LINE]*] ]+
[...]
    -r, --require LIBRARY|DIR        Require files before executing the features. If this
                                     option is not specified, all *.rb files that are
                                     siblings or below the features will be loaded auto-
                                     matically
[...]
```

Since `*.feature` files now live either under `spacewalk/testsuite/features/core` or `spacewalk/testsuite/features/secondary` instead of just `spacewalk/testsuite/features` this means that `support` and `step_definition` are not sibling directories any longer - thus an explicit `-r` flag is needed.

So the semi-philosophical question now is: do we accept to break the Cucumber convention here and require users to specify `-r` (thus this PR might be just accepted I guess) or are we considering other solutions, if any exist?

srbarrios:
I vote to let the users know that need to use -r parameter. IMHO, it make sense to keep this organization.
I don't understand why cucumber framework don't recomend it as the default structure.

Bischoff:
> I vote to let the users know that need to use -r parameter. IMHO, it make sense to keep this organization.
> I don't understand why cucumber framework don't recomend it as the default structure.

Without pipelines, it used to work without `-r`.

Don't ask me why pipelines makes it necessary.

Bischoff:
> So the semi-philosophical question now is: do we accept to break the Cucumber convention here and require users to specify `-r` (thus this PR might be just accepted I guess) or are we considering other solutions, if any exist?

I don't really think it is really a "convention". I think it was just not thought enough :troll: .

The solution is to document `-r` as you did here.

moio:
Merging not to feed the :trollface: !

PR #720, opened 2020-05-26
Revert change to old repositories name

Bischoff:
## What does this PR change?

This PR reverts commit 0a0391dccc49233a2f089be2c6b6e511c62202bb from 17 October 2019. 

That commit tried to revert to old repositories names in sumaform, while the issue was already fixed on the test suite side by switching to the new names on 16 October 2019 in commit  SUSE/spacewalk@05a0aa0fbde999e8e6412a49bd965d333ec8d7fa .


Bischoff:
> But I would like to understand why it was not failing before 

Me too... :disappointed:  Maybe some different Jinja case handling that was hiding the issue? I have no better explanation so far.

PR #721, opened 2020-06-02
Enable SUSE Manager 4.1 on the mirror

juliogonzalez:
## What does this PR change?

Enable SUSE Manager 4.1 on the mirror. Do not merge yet. I need the product definitions building at 4.1 devel project (in a few days)

3.2 are not used by any product since 4.0 GM was published, and are not needed anymore.

### DIFF
--- a/salt/mirror/minima.yaml
+++ b/salt/mirror/minima.yaml
@@ -100,10 +100,10 @@ scc:
     - SLE-Product-SUSE-Manager-Server-4.0-Updates
     - SLE-Module-SUSE-Manager-Server-4.0-Pool
     - SLE-Module-SUSE-Manager-Server-4.0-Updates
-    #- SLE-Product-SUSE-Manager-Server-4.1-Pool
-    #- SLE-Product-SUSE-Manager-Server-4.1-Updates
-    #- SLE-Module-SUSE-Manager-Server-4.1-Pool
-    #- SLE-Module-SUSE-Manager-Server-4.1-Updates
+    - SLE-Product-SUSE-Manager-Server-4.1-Pool
+    - SLE-Product-SUSE-Manager-Server-4.1-Updates
+    - SLE-Module-SUSE-Manager-Server-4.1-Pool
+    - SLE-Module-SUSE-Manager-Server-4.1-Updates
     # SUSE Manager Proxy
     - SUSE-Manager-Proxy-3.2-Pool
     - SUSE-Manager-Proxy-3.2-Updates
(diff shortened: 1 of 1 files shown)


salt/mirror/minima.yaml
```diff
@@ -243,29 +243,9 @@ http:
     archs: [x86_64]
 
   # SUSE Manager 4.1 devel
-  #- url: http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/4.1/SLE_15_SP2
-  #  archs: [x86_64]
-  #- url: http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/4.1:/ToSLE/SLE_15_SP2/
-  #  archs: [x86_64]
-
-  # SLE 11 SP4 Manager Tools 3.2 devel
-  - url: http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/3.2:/SLE11-SUSE-Manager-Tools/images/repo/SLE-11-SP4-CLIENT-TOOLS-ia64-ppc64-s390x-x86_64-Media1/suse
```
rjmateus, line 252: Why are you removing 3.2 tools for sles 11SP4?

salt/mirror/minima.yaml
```diff
@@ -243,29 +243,9 @@ http:
...
+  - url: http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/4.1/SLE_15_SP2
     archs: [x86_64]
-
-  # SLE 12 (GA and all SPs) Manager Tools 3.2 devel
-  - url: http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/3.2:/SLE12-SUSE-Manager-Tools/images/repo/SLE-12-Manager-Tools-POOL-x86_64-Media1
```
rjmateus, line 256: why remove tools for manager 3.2?

juliogonzalez: "One set client tools to rule them all and bind them to the most recent SUSE Manager version."

Or with a not-so-Tolkien-style: They must not be used. They are unmaintained for almost 2 years already.

Remember: we only have two set of client tools.

1. The most recent released version (4.0) that is used for all released versions (4.0 and 3.2).
2. The development version (Head) that is used for the development version (Head) and at some point becomes the set of client tools (as 4.1 in this case).

So after 4.1 is out, it will be the time to remove 4.0 client tools.

rjmateus: ok, thanks for the information :)

juliogonzalez:
Al 4.1 repos created, so this should work fine. Merging.

PR #723, opened 2020-06-23
Add openSUSE Leap 15.2 and move Uyuni Master to it

juliogonzalez:
## What does this PR change?

Add openSUSE Leap 15.2 and move Uyuni Master to it

Libvirt image: https://build.opensuse.org/package/show/systemsmanagement:sumaform:images:libvirt/opensuse152
OpenStack iamge: https://build.opensuse.org/package/show/systemsmanagement:sumaform:images:openstack/opensuse152

### DIFF
--- a/backend_modules/libvirt/base/main.tf
+++ b/backend_modules/libvirt/base/main.tf
@@ -10,6 +10,7 @@ locals {
     opensuse150o = "https://download.opensuse.org/distribution/leap/15.0/jeos/openSUSE-Leap-15.0-JeOS.x86_64-15.0.1-OpenStack-Cloud-Current.qcow2"
     opensuse151  = "https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse151.x86_64.qcow2"
     opensuse151o = "https://download.opensuse.org/distribution/leap/15.1/jeos/openSUSE-Leap-15.1-JeOS.x86_64-15.1.0-OpenStack-Cloud-Current.qcow2"
+    opensuse152o = "https://download.opensuse.org/distribution/leap/15.2/appliances/openSUSE-Leap-15.2-JeOS.x86_64-OpenStack-Cloud.qcow2"
     sles15       = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles15.x86_64.qcow2"
     sles15o      = "http://download.suse.de/install/SLE-15-JeOS-GM/SLES15-JeOS.x86_64-15.0-OpenStack-Cloud-GM.qcow2"
     sles15sp1    = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles15sp1.x86_64.qcow2"
--- a/backend_modules/null/host/variables.tf
+++ b/backend_modules/null/host/variables.tf
@@ -86,7 +86,7 @@ variable "connect_to_additional_network" {
 }
 
 variable "image" {
-  description = "One of: opensuse150, opensuse151, sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles15, centos7"
+  description = "One of: opensuse150, opensuse151, opensuse152o, sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles15, centos7"
   type        = string
 }
 
--- a/main.tf.uyuni.example
+++ b/main.tf.uyuni.example
@@ -22,7 +22,7 @@ module "base" {
 //  }
 
   // Required images
-  images = ["centos7", "opensuse151", "ubuntu1804"]
+  images = ["centos7", "opensuse151", "opensuse152o", "ubuntu1804"]
 }
 
 module "server" {
@@ -30,7 +30,7 @@ module "server" {
   base_configuration = module.base.configuration
   product_version = "uyuni-master"
   name = "srv"
-  image = "opensuse151"
+  image = "opensuse152o"
   use_os_released_updates = true
   // see modules/server/variables.tf for possible values
 
--- a/modules/base/variables.tf
+++ b/modules/base/variables.tf
@@ -60,6 +60,6 @@ variable "provider_settings" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
-  default     = ["centos7", "opensuse150", "opensuse151", "sles15", "sles15sp1", "sles15sp2", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4", "ubuntu1804"]
+  default     = ["centos7", "opensuse150", "opensuse151", "opensuse152o", "sles15", "sles15sp1", "sles15sp2", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4", "ubuntu1804"]
   type        = set(string)
 }
--- a/modules/proxy/main.tf
+++ b/modules/proxy/main.tf
@@ -5,7 +5,7 @@ variable "images" {
     "4.0-released"   = "sles15sp1"
     "4.0-nightly"    = "sles15sp1"
     "head"           = "sles15sp2"
-    "uyuni-master"   = "opensuse151"
+    "uyuni-master"   = "opensuse152o"
     "uyuni-released" = "opensuse151"
   }
 }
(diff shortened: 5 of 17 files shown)


modules/proxy/main.tf
```diff
@@ -5,7 +5,7 @@ variable "images" {
     "4.0-released"   = "sles15sp1"
     "4.0-nightly"    = "sles15sp1"
     "head"           = "sles15sp2"
-    "uyuni-master"   = "opensuse151"
+    "uyuni-master"   = "opensuse152"
     "uyuni-released" = "opensuse151"
```
juliogonzalez, line 9: `uyuni-released` will be changed with Uyuni 2020.07 is released.

backend_modules/libvirt/base/main.tf
```diff
@@ -10,6 +10,8 @@ locals {
     opensuse150o = "https://download.opensuse.org/distribution/leap/15.0/jeos/openSUSE-Leap-15.0-JeOS.x86_64-15.0.1-OpenStack-Cloud-Current.qcow2"
     opensuse151  = "https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse151.x86_64.qcow2"
     opensuse151o = "https://download.opensuse.org/distribution/leap/15.1/jeos/openSUSE-Leap-15.1-JeOS.x86_64-15.1.0-OpenStack-Cloud-Current.qcow2"
+    opensuse152  = "https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse152.x86_64.qcow2"
```
rjmateus, line 13: should we have the sumaform specific image on new OS?  We are planning on moving away from it soon and only let the one that doesn't provide official images, like `sles12sp3` for example.

main.tf.uyuni.example
```diff
@@ -22,15 +22,15 @@ module "base" {
 //  }
 
   // Required images
-  images = ["centos7", "opensuse151", "ubuntu1804"]
+  images = ["centos7", "opensuse151", "opensuse152", "ubuntu1804"]
```
rjmateus, line 25: should `opensuse151` be remove in here, and update to version 15.2 any minion we have defined in her?

salt/mirror/mirror.sh.conf
```diff
@@ -2,6 +2,7 @@ SCC_CREDS="{{ grains.get('cc_username') }}:{{ grains.get('cc_password') }}"
 IMAGES="https://github.com/moio/sumaform-images/releases/download/4.3.0/centos7.qcow2
 https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse150.x86_64.qcow2
 https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse151.x86_64.qcow2
+https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse152.x86_64.qcow2
```
rjmateus, line 5: adding this line will depend on the comment above, and decide if we want to add support to sumaform specific images

rjmateus:
I have tested on libvirt and LGTM

backend_modules/libvirt/base/main.tf
moio: I agree we should not create a custom image unless we have a particular reason to.

main.tf.uyuni.example
Bischoff: it's still needed for KVM; that change will be done later

backend_modules/libvirt/base/main.tf
juliogonzalez: OK. Will change it.

juliogonzalez: But then I will call it already `opensuse152`, as anyway I guess you are goin to the same for the others, right?

moio: No, please call it `opensuse152o` - all "official" images (in the sense of non-custom-to sumaform) are currently named that way.

juliogonzalez: Oh, god, this is going to confuse people @moio.

Now you have one but not the other.

rjmateus: This is what is already happening with ubuntu 20 image, we only have support for `ubuntu2004o`

juliogonzalez: Yeah, Silvio told me at rocket.chat. So now I added `opensuse152o` only, but keep in mind this is going to confuse people, as for now nobody is using those images for ubuntu16, etc, and I am pretty sure everyone expects having images without the `o` suffix.

We really need to do the switch to only official images soon, and make things standard.

backend_modules/null/host/variables.tf
```diff
@@ -86,7 +86,7 @@ variable "connect_to_additional_network" {
 }
 
 variable "image" {
-  description = "One of: opensuse150, opensuse151, sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles15, centos7"
+  description = "One of: opensuse150, opensuse151, opensuse152, sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles15, centos7"
```
rjmateus, line 89: should be `opensuse152o`

modules/base/variables.tf
```diff
@@ -60,6 +60,6 @@ variable "provider_settings" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
-  default     = ["centos7", "opensuse150", "opensuse151", "sles15", "sles15sp1", "sles15sp2", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4", "ubuntu1804"]
+  default     = ["centos7", "opensuse150", "opensuse151", "opensuse152", "sles15", "sles15sp1", "sles15sp2", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4", "ubuntu1804"]
```
rjmateus, line 63: should be `opensuse152o`

salt/mirror/minima.yaml
```diff
@@ -7,66 +7,7 @@ scc:
   password: {{ grains.get("cc_password") }}
   repo_names:
     # required by testsuite
-    - SLES11-SP1-Pool
```
rjmateus, line 10: Do you really want to delete all these repos?

salt/mirror/mirror.sh.conf
```diff
@@ -2,6 +2,7 @@ SCC_CREDS="{{ grains.get('cc_username') }}:{{ grains.get('cc_password') }}"
 IMAGES="https://github.com/moio/sumaform-images/releases/download/4.3.0/centos7.qcow2
 https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse150.x86_64.qcow2
 https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse151.x86_64.qcow2
+https://download.opensuse.org/distribution/leap/15.2/appliances/openSUSE-Leap-15.2-JeOS.x86_64-kvm-and-xen.qcow2
```
rjmateus, line 5: on libvirt base we are defining the direct URL to the image and didn't use the mirror. Should this image be download to the mirror?

salt/mirror/minima.yaml
juliogonzalez: crap!!!!!!!

That's what I used for testing.

salt/mirror/mirror.sh.conf
juliogonzalez: Ok, then for now I will remove it from here.

But we should really support using the official images from the mirror as well :-)

modules/base/variables.tf
juliogonzalez: Fixed, but I'd say you need to add all the other images you mentioned, such as centos6o, sle12sp5o, etc, right? :-D

backend_modules/null/host/variables.tf
juliogonzalez: Solved.

salt/mirror/minima.yaml
juliogonzalez: Solved.

modules/base/variables.tf
juliogonzalez: Solved.

rjmateus: Good point, I'm not sure. Maybe a work to be done when we "promote" the official images.

salt/mirror/mirror.sh.conf
rjmateus: makes totally sense. I think we can do it when we "promote" the official images.

juliogonzalez:
Ok, merging, merging susemanager-ci PR... and let's hope nothing will break :-D

PR #725, opened 2020-06-30
Support testing of preinstalled SUSE Manager images

jgleissner:
## What does this PR change?

This PR enables sumaform to test images with preinstalled SUSE Manager, for instance the ones that SUSE releases in the public clouds.

The PR adds an option to skip package installation, which allows the server to be bootstrapped without a Manager repository and also makes sure that the existing packages are not altered.

It also adds an option to skip creating the first user without disabling the state completely so states depending on that state can still be used (like creating the sample channel). This is useful with the public cloud images as they have a hook to automatically create the first user. There is also a sub-option to change the password of the first user, as the first user's password in the public cloud images is based on the instance ID which would required a bit of fiddling to use in the salt bootstrap, and since (to my possibly outdated knowledge) the test suite uses a hardcoded admin password, it seemed like a useful feature anyway.

I used this code base to successfully test SUSE Manager 4.1 images in AWS, so I thought it may be useful to merge it into upstream. Please let me know what you think.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -229,6 +229,8 @@ By default, sumaform deploys hosts with a range of tweaked settings for convenie
    * `auto_accept`: whether to automatically accept minion keys. Set to `false` to manually accept
    * `create_first_user`: whether to automatically create the first user (the SUSE Manager Admin)
      * `server_username` and `server_password`: define credentials for the first user, admin/admin by default
+     * `first_user_preset`: whether the system already has a first user (e.g. when using public cloud images)
+     * `first_user_change_password`: change password for pre-existing first user. User name and password set in `server_username` and `server_password`
    * `disable_firewall`: disables the firewall making all ports available to any host. Set to `false` to only have typical SUSE Manager ports open
    * `allow_postgres_connections`: configure Postgres to accept connections from external hosts. Set to `false` to only allow localhost connections
    * `unsafe_postgres`: use PostgreSQL settings that improve performance by worsening durability. Set to `false` to ensure durability
@@ -240,6 +242,7 @@ By default, sumaform deploys hosts with a range of tweaked settings for convenie
    * `create_sample_bootstrap_script`: whether to create a sample bootstrap script for traditional clients. Requires `create_sample_activation_key`
    * `publish_private_ssl_key`: copies the private SSL key in /pub for Proxies to copy automatically. Set to `false` for manual distribution
    * `disable_download_tokens`: disable package token download checks. Set to `false` to enable checking
+   * `skip_install`: do not install server packages. Intended to be used with images that have all required packages pre-installed
 
 
 ## Adding channels to SUSE Manager Servers
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -240,7 +240,6 @@ By default, sumaform deploys hosts with a range of tweaked settings for convenie
    * `create_sample_bootstrap_script`: whether to create a sample bootstrap script for traditional clients. Requires `create_sample_activation_key`
    * `publish_private_ssl_key`: copies the private SSL key in /pub for Proxies to copy automatically. Set to `false` for manual distribution
    * `disable_download_tokens`: disable package token download checks. Set to `false` to enable checking
-   * `skip_install`: do not install server packages. Intended to be used with images that have all required packages pre-installed
 
 
 ## Adding channels to SUSE Manager Servers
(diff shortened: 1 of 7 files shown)


README_ADVANCED.md
```diff
@@ -214,6 +214,8 @@ By default, sumaform deploys hosts with a range of tweaked settings for convenie
    * `auto_accept`: whether to automatically accept minion keys. Set to `false` to manually accept
    * `create_first_user`: whether to automatically create the first user (the SUSE Manager Admin)
      * `server_username` and `server_password`: define credentials for the first user, admin/admin by default
+     * `first_user_preset`: whether the system already has a first user (e.g. when using public cloud images)
```
moio, line 217: What happens setting `create_first_user` on `master` sumaform and giving the already existing username and password? My expectation would be that the state would see that the user is already there (with the right password) and just move on.

If the above is correct I would suggest not to add those two new parameters, rather adapt the state to change the password in case the user is already there (with a wrong password).

At that point `create_first_user` would act like a state, to ensure that the first user is there with the right password.

Does this make sense to you?

salt/server/init.sls
```diff
@@ -16,6 +16,7 @@ include:
   - server.tcpdump
 
 server_packages:
+  {% if not grains.get('skip_install') %}
```
moio, line 19: I would have expected `pkg.latest` to just do nothing if the right packages are already in place. What happens instead, and why is that a problem?

salt/default/network.sls
```diff
@@ -62,9 +62,11 @@ ipv6_disable_all:
 
 {% endif %}
 
+{% if not grains.get('skip_install') %}
```
moio, line 65: This should definitely be governed by a separate "don't mess with the firewall" flag IMO, as it could be useful in other contexts.

README_ADVANCED.md
jgleissner: Good idea. It would also make it a bit simpler for the user. :)

salt/server/init.sls
jgleissner: I think it errors out if there are no repositories. And if there are repositories there's always the possibility that something gets installed or updated, and if you want to specifically test an image, you don't want the installed packages to be touched, as this may mask issues in the image.

salt/default/network.sls
jgleissner: Possibly, but again if you want to test an image you don't want modifications so you don't really want to have that state executed. Especially since it removes SUSE Manager packages, as firewalld is a dependency of the server pattern.

moio: I am not arguing against this check, just mentioning it should be under a different name/switch. We can also take care of that if you prefer.

salt/server/init.sls
moio: @mcalmer do you perhaps remember a good reason why we needed `pkg.latest` for the pattern instead of `pkg.installed`?

On a current read I am not sure we need the updating at all.

salt/default/network.sls
jgleissner: It could be a different switch, but my point is that `remove_firewall_packages` should not be executed in case you don't want to touch the SUSE Manager packages, as `susemanager` requires `firewalld` so that state removes not only `firewalld` but also some Manager packages.

Considering that installing SUSE Manager will also install `firewalld`, does `remove_firewall_packages` even make sense?

moio: That is a good point and most likely the result of accumulated cruft that was never cleaned up properly. I promise to give it a more thorough look and come back to you with some better answer.

moio: OK this is definitely wrong, we already have a `firewall.sls` file that does the right thing (install a firewall, disable it if we have the existing `disable_firewall` flag set and otherwise open up essential ports) - but only for the `server` module.

Of course we should extend the approach to all other modules and have some graceful default to fall back for the generic host. That will come in a separate PR.

Thanks for bringing this up

moio: https://github.com/uyuni-project/sumaform/pull/736 should take care of this, please rebase on top of it once merged.

salt/server/init.sls
moio: OK, changed in https://github.com/uyuni-project/sumaform/pull/743, to be merged soon. My understanding is that `skip_install` can be removed altogether with that change in - please let us know if that is not the case.

modules/controller/main.tf
```diff
@@ -32,7 +32,7 @@ module "controller" {
     git_password = var.git_password
     git_repo     = var.git_repo
     branch       = var.branch == "default" ? var.testsuite-branch[var.server_configuration["product_version"]] : var.branch
-    mirror       = var.base_configuration["mirror"]
+    mirror       = var.no_mirror == true ? null :  var.base_configuration["mirror"]
```
moio, line 35: I am probably missing something - isn't `base_configuration["mirror"]` already optional? What is the reason why it can't be set to null without a new variable?

moio: Forget it, I figured it out, you selectively don't want to use a mirror for the controller only. Sorry for the noise.

jgleissner:
So do you think this PR is useful? Any concerns with any of the changes?

moio (approved):
Sure, we just forgot pressing a button and went on vacation!

Sorry and thanks again!

PR #729, opened 2020-07-14
Install NTP on most hosts

Bischoff:
## What does this PR change?

Experimentation proved we could have up to 5 minutes offset between the different machines deployed by sumaform. It's not only a problem for the test suite, but could hit the operation of SUSE Manager itself.

Corresponding issue: https://github.com/SUSE/spacewalk/issues/11934
Test suite part: uyuni-project/uyuni#2413

Implementation details:
* there's a new option, `use_ntp`, that is `true` by default, and that enables you to disable this code.
* the old `timezone` stuff moves to the new `time.sls` states file.
* CentOS and Ubuntu are not handled because they are already correct, as far as I can see

Concerns with current code:
* what happens with vanilla hosts? I guess we will get this stuff (?) as I call it from `minimal.sls`. Do we really want that?
* probably too many `use_ntp` here and there, I threw them a bit at random

### DIFF
--- a/backend_modules/aws/host/main.tf
+++ b/backend_modules/aws/host/main.tf
@@ -145,6 +145,8 @@ resource "null_resource" "host_salt_configuration" {
     domain_id      = length(aws_instance.instance) == var.quantity ? aws_instance.instance[count.index].id : null
     grains_subset = yamlencode(
       {
+        timezone              = var.base_configuration["timezone"]
+        use_ntp               = var.base_configuration["use_ntp"]
         testsuite             = var.base_configuration["testsuite"]
         roles                 = var.roles
         additional_repos      = var.additional_repos
@@ -189,6 +191,7 @@ resource "null_resource" "host_salt_configuration" {
         use_avahi : false
 
         timezone                  = var.base_configuration["timezone"]
+        use_ntp                   = var.base_configuration["use_ntp"]
         testsuite                 = var.base_configuration["testsuite"]
         roles                     = var.roles
         use_os_released_updates   = var.use_os_released_updates
--- a/backend_modules/ssh/host/main.tf
+++ b/backend_modules/ssh/host/main.tf
@@ -22,6 +22,9 @@ resource "null_resource" "provisioning" {
     base_configuration = yamlencode(var.base_configuration)
     grains_subset = yamlencode(
       {
+        timezone                  = var.base_configuration["timezone"]
+        use_ntp                   = var.base_configuration["use_ntp"]
+        testsuite                 = var.base_configuration["testsuite"]
         hostname                  = "${local.resource_name_prefix}${var.quantity > 1 ? "-${count.index + 1}" : ""}"
         roles                     = var.roles
         use_os_released_updates   = var.use_os_released_updates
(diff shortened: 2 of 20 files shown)


moio:
Thanks for taking care of this, it was _much needed_!

Bischoff:
Tested to work.

juliogonzalez:
It seems tests are not happy :-)

backend_modules/aws/host/main.tf
```diff
@@ -189,6 +189,7 @@ resource "null_resource" "host_salt_configuration" {
         use_avahi : false
 
         timezone                  = var.base_configuration["timezone"]
+        use_ntp                   = var.base_configuration["use_ntp"]
```
rjmateus, line 192: Should we add this variable also to the triggers?

backend_modules/ssh/host/main.tf
```diff
@@ -74,6 +74,7 @@ resource "null_resource" "provisioning" {
         use_avahi                 = var.base_configuration["use_avahi"]
         additional_network        = var.base_configuration["additional_network"]
         timezone                  = var.base_configuration["timezone"]
+        use_ntp                   = var.base_configuration["use_ntp"]
```
rjmateus, line 77: shouldn't we add this variable also to triggers?

salt/default/time.sls
```diff
@@ -0,0 +1,55 @@
...
+timezone_setting:
+  timezone.system:
+    - name: {{ grains['timezone'] }}
+    - utc: True
+    - require:
+      - file: timezone_symlink
+
+{% if grains['use_ntp'] %}
+
+{% if grains['osfullname'] == 'SLES' %}
```
rjmateus, line 26: service is installed only for sles and leap. Shouldn't we be doing it on all OS, or at least the ones used by the testsuite?

backend_modules/aws/host/main.tf
Bischoff: Let's discuss in chat. That's my problem, I don't understand the architecture of tyhses new backends, and I added those veriables quasi at random.

salt/default/time.sls
Bischoff: That's explained in the PR description: it seems it's already correct on CentOS and Ubuntu.

backend_modules/ssh/host/main.tf
Bischoff: same, no idea, i'm lost

Bischoff:
> LGTM... as soon as tests are passing :-D

Yes, I'm on it. I'm starting to fear the tests are incorrect, but let's see.

Bischoff:
> Seems the tests are in fact passing and the error is about something unrelated, so LGTM.

For the record, the new travis error is:
```
$ coala --ci
[ERROR][15:40:56] There is a conflict in the version of a dependency you have installed and the requirements of coala. This may be resolved by creating a separate virtual environment for coala or running `pip install "importlib-metadata>=1.7.0; python_version < "3.8""`. Be aware that the latter solution might break other python packages that depend on the currently installed version.
```

PR #730, opened 2020-07-15
force installed version of importlib-metadata

rjmateus:

## What does this PR change?

    force installed version of importlib-metadata
    
    This library is a transitive dependency from coala.
    We are getting some conflicts on version number
    since another transitive library is depending on
    this one (stevedore)


Bischoff:
It solves the issues i've been seing recently.

PR #731, opened 2020-07-15
Add a configuration file for chrony too

Bischoff:
## What does this PR change?

Follow-up of #729.

It looks like `chronyd` needs to be properly configured too.


Bischoff:
Local tests passed, merging.

Issue #732, opened 2020-07-17
Have the test suite module work with several clients, minions, SSH minions, etc.


Bischoff:
## problem 1

Identifiers for the different client types lead to absurdities. Instead of
```
    cli-sles12sp4 = {
      name = "client"
      image = "opensuse152o"
    }
```
we should have something like:
```
    suse_client = {
      name = "client"
      image = "opensuse152o"
    }
```

## problem 2

Testsuite module accepts 0 or 1 of each client, it should be 0 to _N_, especially for the QAM test suite:
```
  suse_clients = [
       client1 = {
          name = "client-leap-15.2"
          image = "opensuse152o"
    }, client2 = {
          name = "client-sles-15.1"
          image = "sles15sp1"
    }
   ]
```

## problem 3

For big setups like the QAM test suite, and for the uyuni test suite, we have to be able to specify different providers in base module. It could be several libvirt hosts, or even a mix of libvirt and cloud hosts.

This is currently possible when not using the cucumber test suite module, with things like that:
```
module "cli-sles15sp1" {
  providers = {
    libvirt = libvirt.classic181
  }
  source             = "./modules/client"
  base_configuration = module.base3.configuration
  ...
}

module "base3" {
  providers = {
    libvirt = libvirt.classic181
  }
  source = "./modules/base"
  provider_settings = {
    pool        = "default"
    bridge      = "br0"
    additional_network = "192.168.40.0/24"
  }
  ...
}
```

I'm not sure if we can achieve this with current test suite module.

moio:
I see problems 1 and 2 as related, and strictly dependent on the testsuite code: it was structured the way it is to follow what the testsuite needed. If the testsuite becomes more flexible in accepting VMs, my opinion is that we should follow that flexibility in the same way in sumaform accordingly.

Question is: does the Cucumber testsuite accept only one "SUSE client" and the version is a variable, or should it accept an array of clients with a bunch of OS's? I tend to think the latter is preferable. So one idea could be to expect an array of clients, each of which will have an OS and some flags or a label that indicates its "role" (traditional client, minion, image build host...).

About problem 3, there is the possibility of defining multiple providers in the top-level main.tf file and then pass a reference to a certain provider down in the modules. It requires some surgery but it can be done, and I do not think there's much of an alternative in this case (unless we want multiple separated main.tfs, which is probably worse).

Thanks for bringing this up

Bischoff:
>  it was structured the way it is to follow what the testsuite needed.

[...]

PR #734, opened 2020-07-20
Use better names for hosts in test suite module

Bischoff:
## What does this PR change?

Addresses #732

 * `present = true` is unnecessary
     (and `present = false` would not work)
* Use more generic names for hosts
    Stop specifying version numbers, and refer to generic operating system "families"
    (suse, redhat, and debian), that can map to e.g. openSUSE, CentOS, or Ubuntu.

susemanager-ci: SUSE/susemanager-ci#102
terracumber: https://gitlab.suse.de/juliogonzalezgil/terracumber/-/merge_requests/7


rjmateus:
The present flag was only used to avoid the usage of empty map. This flag doesn't have any impact, so Eric point one big problem: if soneone set the flag to false, it will take no effect. So removing the flag and just use empty map

Bischoff:
> The present flag was only used to avoid the usage of empty map. This flag doesn't have any impact, so Eric point one big problem: if soneone set the flag to false, it will take no effect. So removing the flag and just use empty map

The empty set is also a set... I don't think that anyone with computing background will be disturbed by contructs like
```
  client = {  }
```

Bischoff:
Impossible to have a variable number of clients, reducing the focus.

README_ADVANCED.md
```diff
@@ -615,12 +615,12 @@ module "server" {
...
 
 ```hcl
-module "minsles12sp1" {
+module "suse-minion" {
   source = "./modules/minion"
 
-  name = "minsles12sp1"
-  image = "sles12sp1"
+  name = "minsles12sp4"
```
calancha, line 622: Could we have `min-sles12sp4` and `cli-sles12sp4`. ie., with a '-' separator, in analogy with `min-sles12sp4`

README_TESTING.md
```diff
@@ -97,31 +97,22 @@ Several test hosts are optional and can be activated via a host_settings block l
 ```hcl
 host_settings = {
   pxy = {
-    present = true
   }
   cli-sles12sp4 = {
-    present = true
   }
   min-sles12sp4 = {
```
calancha: We should update the doc; here the modules show the old naming.

README_ADVANCED.md
Bischoff: You spotted it... I also did, but was too lazy to fix. I will now.

README_TESTING.md
Bischoff: Uh I missed it?

modules/controller/main.tf
```diff
@@ -39,8 +39,8 @@ module "controller" {
     client        = length(var.client_configuration["hostnames"]) > 0 ? var.client_configuration["hostnames"][0] : null
```
rjmateus, line 39: should those names also be changed to suse_client` and `suse_minion`?

modules/controller/main.tf
```diff
@@ -39,8 +39,8 @@ module "controller" {
...
     minion        = length(var.minion_configuration["hostnames"]) > 0 ? var.minion_configuration["hostnames"][0] : null
     build_host    = length(var.buildhost_configuration["hostnames"]) > 0 ? var.buildhost_configuration["hostnames"][0] : null
-    centos_minion = length(var.centos_configuration["hostnames"]) > 0 ? var.centos_configuration["hostnames"][0] : null
-    ubuntu_minion = length(var.ubuntu_configuration["hostnames"]) > 0 ? var.ubuntu_configuration["hostnames"][0] : null
+    redhat_minion = length(var.redhat_configuration["hostnames"]) > 0 ? var.redhat_configuration["hostnames"][0] : null
+    debian_minion = length(var.debian_configuration["hostnames"]) > 0 ? var.debian_configuration["hostnames"][0] : null
```
rjmateus, line 43: I'm not surre if we should change ubuntu to debian name. We also have a debian distribution which we may need to test in the future. Or the goal is just to test one debian base distribution in this context?

README_ADVANCED.md
```diff
@@ -26,22 +26,22 @@ Note: on clients and minions only, the version number can be omitted to take the
...
 ```hcl
-module "minsles12sp1" {
+module "suse-minion" {
   source = "./modules/minion"
 
-  name = "minsles12sp1"
-  image = "sles12sp1"
+  name = "min-sles12sp4"
+  image = "sles12sp4"
```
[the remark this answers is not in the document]
Bischoff: it's just examples but why not. I just randomly bumped version to something less antique 😸 .

README_ADVANCED.md
```diff
@@ -26,22 +26,22 @@ Note: on clients and minions only, the version number can be omitted to take the
 A libvirt example follows:
 
 ```hcl
-module "minsles12sp1" {
+module "suse-minion" {
```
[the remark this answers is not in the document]
Bischoff: It could. It's just to be consistent.

modules/controller/main.tf
Bischoff: Yes, the goal is to have, at least in regular testsuite, "generic" test targets: SUSE (no matter wether SLES or openSUSE), RedHat (no matter wether Fedora, Centos, RHEL, SLES ES, etc.), Debian (no matter wether Ubuntu, Debian, or other).

The QAM test suite is another story and has no "generic" targets.

modules/controller/main.tf
Bischoff: It would indeed be nice, but I tried to keep the changes minimal. I would say it's not a big deal here.

README_ADVANCED.md
```diff
@@ -60,9 +60,9 @@ module "server" {
   source = "./modules/server"
   base_configuration = module.base.configuration
 
-  image = "sles12sp3"
+  image = "sles12sp5"
```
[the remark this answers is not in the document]
Bischoff: Will change too.

PR #736, opened 2020-07-24
Do not remove firewall packages but disable it

rjmateus:
## What does this PR change?

Instead of removing firewall packages which will be needed
by susemanager we now disable the firewall with a parameter
if it is needed.

PR relates to: https://github.com/uyuni-project/sumaform/pull/725

Solves: https://github.com/SUSE/spacewalk/issues/12012

## TODO
- [ ] Run CI with this branch and check results

### DIFF
--- a/modules/client/variables.tf
+++ b/modules/client/variables.tf
@@ -31,6 +31,11 @@ variable "use_os_unreleased_updates" {
   default     = false
 }
 
+variable "disable_firewall" {
+  description = "whether to disable the built-in firewall, opening up all ports"
+  default     = true
+}
+
 variable "additional_repos" {
   description = "extra repositories in the form {label = url}, see README_ADVANCED.md"
   default     = {}
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -12,7 +12,6 @@ module "host" {
   additional_certs              = var.additional_certs
   additional_packages           = var.additional_packages
   quantity                      = var.quantity
-  grains                        = var.grains
   swap_file_size                = var.swap_file_size
   ssh_key_path                  = var.ssh_key_path
   gpg_keys                      = var.gpg_keys
@@ -24,6 +23,9 @@ module "host" {
   provider_settings             = var.provider_settings
   additional_disk_size          = var.additional_disk_size
   volume_provider_settings      = var.volume_provider_settings
+
+  grains = merge({ disable_firewall = var.disable_firewall },
+  var.grains)
 }
 
 output "configuration" {
--- a/salt/default/firewall.sls
+++ b/salt/default/firewall.sls
@@ -1,0 +1,12 @@
+
+{% if grains.get('disable_firewall') | default(true, true)  %}
+
+disable_firewall:
+  service.dead:
+{% if grains.get('osmajorrelease', None)|int() == 15 %}
+    - name: firewalld
+{% else %}
+    - name: SuSEfirewall2
+{% endif %}
+    - enable: False
+{% endif %}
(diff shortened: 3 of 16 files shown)


rjmateus:
I have run testsuite one time with: proxy, traditional client, minion and ssh minion.
Result:
```
Failing Scenarios:
cucumber features/secondary/trad_migrate_to_minion.feature:95 # Scenario: Cleanup: register minion again as traditional client
cucumber features/secondary/trad_migrate_to_minion.feature:102 # Scenario: Cleanup: check that this is again a traditional client
cucumber features/secondary/trad_migrate_to_minion.feature:107 # Scenario: Cleanup: check that we have again the old activation key
cucumber features/secondary/trad_migrate_to_sshminion.feature:9 # Scenario: Install a package before migration to Salt SSH minion
cucumber features/secondary/trad_migrate_to_sshminion.feature:77 # Scenario: Check that events history is still the same after migration to Salt SSH
cucumber features/secondary/trad_migrate_to_sshminion.feature:106 # Scenario: Cleanup: remove packages from the migrated SSH minion
cucumber features/secondary/srv_delete_channel_with_tool.feature:33 # Scenario: Delete channel with one clone
cucumber features/secondary/srv_delete_channel_with_tool.feature:41 # Scenario: Delete base channel and clone

419 scenarios (8 failed, 113 skipped, 298 passed)
2968 steps (8 failed, 836 skipped, 2124 passed)
```

rjmateus:
cc @jgleissner

salt/default/firewall.sls
```diff
@@ -24,6 +22,7 @@ disable_firewall:
 
 {% else %}
 
+{% if 'server' in grains.get('roles') %}
```
Bischoff, line 25: only for server ? I'm missing why, here...

modules/client/variables.tf
```diff
@@ -31,6 +31,11 @@ variable "use_os_unreleased_updates" {
   default     = false
 }
 
+variable "disable_firewall" {
+  description = "whether to disable the built-in firewall, opening up all ports"
```
Bischoff, line 35: could it be you forgot to document this new setting in some README ?

salt/default/firewall.sls
rjmateus: The previous behaviour when running testsuite is configure correctly the firewall on server, and disable it on all other roles. I try to keep it as it was, and only configure the firewall for server

modules/client/variables.tf
rjmateus: Thanks. I just add it to readme advanced

salt/default/firewall.sls
Bischoff: ok, thanks for the explanation.

We also configure a firewall on the proxy, but this time from suse manager itself, using a retail formula.

modules/client/variables.tf
Bischoff: Thanks.

modules/host/main.tf
```diff
@@ -24,6 +23,9 @@ module "host" {
   provider_settings             = var.provider_settings
   additional_disk_size          = var.additional_disk_size
   volume_provider_settings      = var.volume_provider_settings
+
+  grains = merge({ disable_firewall = var.disable_firewall },
```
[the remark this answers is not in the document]
rjmateus: My plan is to move all other variables which are independent from backend to here.
I add this in this way to test the structure/approach, and after it get merged, move all other var. On the TODO list :+1:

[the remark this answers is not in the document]
moio: Hmm, that would count as a "disagree and support" from my part ;-)

Go ahead!

rjmateus:
I have tested the deploy locally and it worked.

PR #741, opened 2020-08-10
Default to official images

calancha:
## What does this PR change?
Use by default only official images
    
As before, the value set at the main.tf has precedence.
This covers the first step in https://github.com/SUSE/spacewalk/issues/11911


backend_modules/libvirt/base/main.tf
```diff
@@ -22,6 +22,7 @@ locals {
     sles12sp1    = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp1.x86_64.qcow2"
     sles12sp2    = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp2.x86_64.qcow2"
     sles12sp3    = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp3.x86_64.qcow2"
+    sles12sp3o   = "http://schnell.suse.de/SLE12/SLE-12-SP3-JeOS-GM/SLES12-SP3-JeOS-for-OpenStack-Cloud.x86_64-GM.qcow2"
```
ktsamis, line 25: Not sure that the cloud image is the right one here

modules/cucumber_testsuite/variables.tf
```diff
@@ -31,7 +31,7 @@ variable "name_prefix" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
-  default     = ["centos7", "opensuse150", "opensuse151", "opensuse152o", "sles15", "sles15sp1", "sles15sp2o", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4", "ubuntu1804"]
+  default     = ["centos7", "centos7o", "opensuse150", "opensuse150o", "opensuse151", "opensuse152o", "sles15", "sles15sp1", "sles15sp1o", "sles15sp2o", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp3o", "sles12sp4", "ubuntu1804", "ubuntu1804o"]
```
rjmateus, line 34: if you changed the images to the official ones, you should not download by default the older images.
For example, downloading `centos7` and `centos7o` by default is not needed.

modules/base/variables.tf
```diff
@@ -65,6 +65,6 @@ variable "provider_settings" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
-  default     = ["centos7", "opensuse150", "opensuse151", "opensuse152o", "sles15", "sles15sp1", "sles15sp2o", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4", "ubuntu1804"]
+  default     = ["centos7", "opensuse150", "opensuse151", "opensuse152o", "sles15", "sles15sp1", "sles15sp2o", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp3o", "sles12sp4", "ubuntu1804"]
```
ktsamis, line 68: should we maybe start removing the ones that won't be used, all the ones without "o"?

modules/cucumber_testsuite/variables.tf
ktsamis, line 34: Same here maybe we remove the unneeded ones?

modules/server/main.tf
```diff
@@ -1,14 +1,14 @@
 variable "images" {
   default = {
-    "3.2-released"   = "sles12sp4"
-    "3.2-nightly"    = "sles12sp4"
-    "4.0-released"   = "sles15sp1"
-    "4.0-nightly"    = "sles15sp1"
+    "3.2-released"   = "sles12sp4o"
+    "3.2-nightly"    = "sles12sp4o"
```
rjmateus, line 4: if you changed these variables the default value for `images` variable in base module should also be updated.
For exmaple, by default the images `sles12sp4o`, `sles15sp1o` and `opensuse151o` will not be downloaded, meaning the default configuration will return an error.

modules/host/variables.tf
```diff
@@ -86,7 +86,7 @@ variable "connect_to_additional_network" {
 }
 
 variable "image" {
-  description = "One of: opensuse150, opensuse151, opensuse152o, sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles15, centos7"
+  description = "One of: opensuse150, opensuse151, opensuse152o, sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp3o, sles12sp4, sles15, centos7"
```
ktsamis, line 89: why not change here all of them to official?

modules/minion/variables.tf
```diff
@@ -101,7 +101,7 @@ variable "additional_grains" {
 
 
 variable "image" {
-  description = "One of: sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles15, centos7"
+  description = "One of: sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp3o, sles12sp4, sles15, centos7"
```
ktsamis, line 104: same why not add all official ones here?

modules/pxe_boot/variables.tf
```diff
@@ -13,7 +13,7 @@ variable "quantity" {
 }
 
 variable "image" {
-  description = "One of: sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles15, centos7"
+  description = "One of: sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp3o, sles12sp4, sles15, centos7"
```
ktsamis, line 16: same why not use official ones here for all?

modules/sshminion/variables.tf
```diff
@@ -76,7 +76,7 @@ variable "connect_to_additional_network" {
 }
 
 variable "image" {
-  description = "One of: opensuse150, opensuse151, opensuse152o, sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles15, centos7"
+  description = "One of: opensuse150, opensuse151, opensuse152o, sles11sp4, sles12, sles12sp1, sles12sp2, sles12sp3, sles12sp3o, sles12sp4, sles15, centos7"
```
ktsamis, line 79: same why not use official ones here for all?

backend_modules/libvirt/base/main.tf
srbarrios: Interesting, there is a reason behind it?

modules/base/variables.tf
srbarrios: I would say replace all of them by its official image ending in o, right?

backend_modules/libvirt/base/main.tf
calancha: > Interesting, there is a reason behind it?

It seems to be the default image for the pxeboot-minion module.

calancha: > Not sure that the cloud image is the right one here

We are using a similar image for sles15sp2o,

moio: Please note SLES 12 SP3 never really had an officially released JeOS, therefore we should not use this internal "beta" image.

See https://github.com/uyuni-project/sumaform/pull/676 "Peculiarities"

I would recommend to just continue using our custom image for this particular OS.

modules/proxy/main.tf
```diff
@@ -1,14 +1,14 @@
 variable "images" {
   default = {
-    "3.2-released"   = "sles12sp4"
-    "3.2-nightly"    = "sles12sp4"
-    "4.0-released"   = "sles15sp1"
-    "4.0-nightly"    = "sles15sp1"
+    "3.2-released"   = "sles12sp4o"
+    "3.2-nightly"    = "sles12sp4o"
```
moio, line 4: Have these been tested - even just a one-off "smoke test"?

calancha: I have deployed several times 3.2 with sles12sp4o using my local machine:
- deploy OK
- core features OK
- secondary features: ~ 30 failures
I have no way to test it in a full environment like our test suite, though; that's i would like to
follow an incremental approach in this task.

calancha:
> When you revert the sles12sp3o change, the rest LGTM.

Why do you want to revert sles12sp3o? It is an official image.
it is deploying well in my test machine and passsing core features.

modules/proxy/main.tf
moio: More than sufficient FMPOV

calancha:
- This test run (4.1) fails deployment
https://ci.suse.de/view/Manager/view/Manager-4.1/job/manager-4.1-cucumber-NUE/25/console
- This run (4.1) passes core features:
https://ci.suse.de/view/Manager/view/Manager-4.1/job/manager-4.1-cucumber-NUE/28/
the only difference in setup between these 2 runs is that the former uses sles15sp1o and the latter do not.

calancha:
@moio I am wondering why do we use for sles15sp1o this image:

https://download.suse.de/install/SLE-15-SP1-JeOS-QU2/SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-QU2.qcow2

instead of the following one:

https://download.suse.de/install/SLE-15-SP1-JeOS-GM/SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-GM.qcow2

which follows the same name convention as:
sles15sp2o   = "https://download.suse.de/install/SLE-15-SP2-JeOS-GM/SLES15-SP2-JeOS.x86_64-15.2-OpenStack-Cloud-GM.qcow2"
sles12sp4o   = "http://download.suse.de/install/SLE-12-SP4-JeOS-GM/SLES12-SP4-JeOS.x86_64-12.4-OpenStack-Cloud-GM.qcow2"
sles12sp5o   = "https://download.suse.de/install/SLE-12-SP5-JeOS-GM/SLES12-SP5-JeOS.x86_64-12.5-OpenStack-Cloud-GM.qcow2"

is there a special reason to prefer that SLE-15-SP1-JeOS-QU2?

mcalmer:
QU  == Quaterly Update

Installation images get a quarterly refresh to get the newest kernel and installation tools.
This applies also to JeOS. And with famous grub2 boothole bug older images might not boot anymore in the near future.
So I think it is a good idea to test with the latest available images.

calancha:
@srbarrios See in my 2nd commit your request.
I cannot access right now to the mirror to check if the paths are OK: I assume we create in the mirror directories as needed
to patch the URL structure.
Let me know if I must update `salt/mirror/mirror.sh.conf` as well.

srbarrios:
> @srbarrios See in my 2nd commit your request.
> I cannot access right now to the mirror to check if the paths are OK: I assume we create in the mirror directories as needed
> to patch the URL structure.
> Let me know if I must update `salt/mirror/mirror.sh.conf` as well.

Honeslty I'm not sure at what it serve this mirror.sh.conf on that repo. But for sure is not the same that we are using in our QAM Minima Mirror and I think it also don't match the mirror used in galaxy infrastructure. So, with this commit, it should be enough.

PR #744, opened 2020-08-14
Prevent from unbeatable defaults

calancha:
## What does this PR change?
Prevent from automatically overriding a value set in the main.tf.

`image` is initialized as 'default' for both server and proxy at
modules/server/variables.tf
modules/proxy/variables.tf

I don't see a need to add extra check for the product version:
if image == 'default' we will get the image matching the product version.

The problem with the extra check is as follows:
if the user sets the image as
image = 'foo' ; maybe for a quick test
and if product_version is 'head', then we will ignore the user value 'foo', taking
the default instead.

### DIFF
--- a/modules/proxy/main.tf
+++ b/modules/proxy/main.tf
@@ -46,7 +46,7 @@ module "proxy" {
     repository_disk_size      = var.repository_disk_size
   }
 
-  image                    = var.image == "default" || var.product_version == "head" ? var.images[var.product_version] : var.image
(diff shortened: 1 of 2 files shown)


moio:
Reason why it was coded that way is that head can only be built on exactly one OS, and any other one will fail (or could fail, but that would be unsupported). So we got confused users, and we decided to hardcode the OS for that particular version, exactly in order to avoid user errors.

What is your use case exactly? I am in favor of this patch if there is at least a concrete case when switching OSs really makes sense.

PR #748, opened 2020-08-18
Fix the minion repositories for SLE15 systems (all SPs)

meaksh:
## What does this PR change?

This PR fixes few issues on the repository definition for SLE15 (all SPs) minions:

- Set the right SP repository for the SLE modules (not always SP1)
- Add "SLE-Modules-Desktop-Applications" repository which is required by already added "SLE-Module-Development-Tools"

This PR should fix issues due a mix of SP1 and SP2 repos on SLE15SP2 minions, and also the issues on the Uyuni:Master testsuite regarding "applying highstate on build host".

### DIFF
--- a/salt/repos/minion.sls
+++ b/salt/repos/minion.sls
@@ -49,7 +49,7 @@ desktop_updates_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Desktop-Applications/{{ sle_version_path }}/x86_64/update/
 
-{% if '3.2-nightly' in grains.get('product_version') or '4.0-nightly' in grains.get('product_version') %}
+{% if '3.2-nightly' in grains.get('product_version') or '4.0-nightly' in grains.get('product_version') or '4.1-nightly' in grains.get('product_version') %}
 python2_pool_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Python2/{{ sle_version_path }}/x86_64/product/
--- a/salt/repos/minion.sls
+++ b/salt/repos/minion.sls
@@ -39,10 +39,12 @@ devel_updates_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Development-Tools/{{ sle_version_path }}/x86_64/update/
 
+{# The following "SLE-Module-Desktop-Applications" channel is required by "SLE-Module-Development-Tools" module #}
 desktop_pool_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Desktop-Applications/{{ sle_version_path }}/x86_64/product/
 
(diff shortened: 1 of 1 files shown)


salt/repos/minion.sls
```diff
@@ -12,30 +12,49 @@ containers_updates_repo:
...
+
+desktop_pool_repo:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Desktop-Applications/{{ sle_version_path }}/x86_64/product/
+
+desktop_updates_repo:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Desktop-Applications/{{ sle_version_path }}/x86_64/update/
 
 {% if '3.2-nightly' in grains.get('product_version') or '4.0-nightly' in grains.get('product_version') %}
```
juliogonzalez, line 50: Unrelated to this PR but... why is this only for 3.2 and 4.0 and only for nightly? :-?

salt/repos/minion.sls
```diff
@@ -12,30 +12,49 @@ containers_updates_repo:
...
   pkgrepo.managed:
-    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Development-Tools/15-SP1/x86_64/product/
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Development-Tools/{{ sle_version_path }}/x86_64/product/
 
 devel_updates_repo:
   pkgrepo.managed:
-    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Development-Tools/15-SP1/x86_64/update/
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Development-Tools/{{ sle_version_path }}/x86_64/update/
+
+desktop_pool_repo:
```
juliogonzalez, line 42: I suggest adding a comment telling that this is required for the development tools repos, in case someone else has the same question we had today.

salt/repos/minion.sls
meaksh: Not sure, but I guess it would make sense to add that it for "4.1-nightly" as well.

meaksh: Added

salt/repos/minion.sls
meaksh: Added

salt/repos/minion.sls
juliogonzalez: I guess so, but I fail to understand why we have the if for `nightlies` only.

IIRC this is added because the buildhost needs python2 to build images based on python2, but I fail to see why this is not applied always, as the other repositories.

meaksh: Hmmm, according to @Bischoff comment at https://github.com/uyuni-project/sumaform/commit/7fc0d47378efeedabb83f7dd32a6617e8db49010 , it seems this won't be required for 4.1 at all. Why this is only for nightly version? I don't know.

Maybe @Bischoff can bring some light here since he is the author of that commit.

In any case, since it seems there has been no issue yet on that. I'll revert the workaround for 4.1 to merge this already and will wait until we have some feedback from Eric.

moio:
Thanks!

PR #749, opened 2020-08-18
Cleanup repo metadata after repo changes in CentOS

moio:
## What does this PR change?

Differently than zypper, yum does not invalidate cached repo metadata if
a repo's URL changes, this can lead to situations in which package
downloading is attempted from the wrong URL.

This has been reproduced on 4.0-released traditional clients with the
centos7o image.


moio:

srbarrios:
I will just merge, it will not hurt anyhow, and it we can just test all together.

PR #750, opened 2020-08-18
Add RES8 repos to mirror

moio:
## What does this PR change?

Adds missing repos to support CentOS 8 minions/clients using the `centos8o` image with a mirror.


moio:

PR #755, opened 2020-08-19
Testsuite: openSCAP is only supported for Ubuntu minions from 18.04 on

moio:
Make conditions more precise to match current reality.


moio:

srbarrios:
It seems fixed, the error on Ubuntu16.04 about ssg-debderived disappeared.

srbarrios:
Oups! We missed to delete one `endif` tag.
I will fix it in a PR.

PR #756, opened 2020-08-19
Start qemu-ga at cloud-init stage on CentOS 6

moio:
Apparently qemu-ga is installed but not started by default at least on CentOS 6, leading to terraform-provider-libvirt errors at deploy time.

This is a tentative fix.


moio:
@srbarrios can you please test this?

srbarrios:
Centos6 is deployed now, so it seems it fixed the issue :)

moio:
Let me know if you find others that fail in a similar way, so that I can show you what the root cause is and how to fix it ;-)

moio:
Thanks a lot of course for all the efforts

srbarrios:
I'm not sure how to identify the problem on each case, but those are the clients not deployed atm:
``` 
sles15-sshminion
ubuntu1604-sshminion
ubuntu2004-minion
controller  --> It might fail because the others are failing
```

PR #759, opened 2020-08-20
Fix openscap on Ubuntu16

srbarrios:
## What does this PR change?

The previous if the statement was wrong.
As if the host was a Ubuntu and higher version than 16.04 it was installing something, but in the `else` case, if you had an Ubuntu with 16.04 it went through the else statement.
What it does now is, only install `openscap-utils` package if it is not a Ubuntu, and only  `libopenscap8` ,`ssg-debderived` if the version of Ubuntu is higher than 16.04.

### DIFF
--- a/salt/minion/testsuite.sls
+++ b/salt/minion/testsuite.sls
@@ -8,9 +8,11 @@ minion_cucumber_requisites:
   pkg.installed:
     - pkgs:
       - salt-minion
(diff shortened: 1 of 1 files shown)


rjmateus (approved):

Bischoff:
Not sure it is allowed in Jinja to have a block that resolves to empty between `if` and `else`. But maybe I'm wrong.

PR #760, opened 2020-08-20
Remove man and man-db from testsuite dependencies

srbarrios:
## What does this PR change?

We noticed that man and man-db are not used anymore in our cucumber testsuite.
So we can safely remove them from the dependencies.


moio (approved):

rjmateus (approved):

Bischoff (approved):
LGTM, thanks a lot

Removal from test suite was 
```
commit 948ffca821852bd035ca66594c4758c1d271251d
Author: Eric Bischoff <ebischoff@suse.com>
Date:   Mon Mar 11 14:37:23 2019 +0100

    "man" does not exist on CentOS, use "sed" instead
```

PR #766, opened 2020-08-26
Use modern script name

moio:
## What does this PR change?

It uses the modern name for former `migration.sh` script


mcalmer (approved):
right. Even 3.2 has already that new name.

PR #771, opened 2020-09-15
Fix: add section to troubleshoot Avahi and mDNS

mbologna:
## What does this PR change?

I was having resolution problems with Avahi on an openSUSE 15.2 host trying to resolve a sumaformed Uyuni server:

```
opensuse # avahi-resolve -n uyuni.tf.local
uyuni.tf.local 192.168.122.154
# ping uyuni.tf.local
ping: uyuni.tf.local: Name or service not known
# ping 192.168.122.154
PING 192.168.122.154 (192.168.122.154) 56(84) bytes of data.
64 bytes from 192.168.122.154: icmp_seq=1 ttl=64 time=0.245 ms
```

The problem disappears when I change Avahi configuration in the Uyuni host to use the first-level domain `local` (instead of `tf.local`).

Turns out that in the newest version of `nss-mdns` (shipping with OpenSUSE 15.2), second-level `.local` domains must be explicitly allowed to be resolved in `/etc/mdns.allow` AND `mdns` (not `mdns_minimal`) must be used in `nsswitch.conf`.
The latter is already covered in the troubleshooting, I took care of adding the former information into the docs.

Upstream issue: https://github.com/lathiat/nss-mdns/issues/64
Upstream resolution (linked in sumaform docs): https://github.com/lathiat/nss-mdns/blob/master/README.md#etcmdnsallow

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -30,6 +30,7 @@ Check that:
   - on SUSE systems check YaST -> Security and Users -> Firewall -> Zones -> public, "mdns" should appear on the list on the right
  - avahi is installed and running
    - typically you can check this via systemd: `systemctl status avahi-daemon`
+ - Avahi is capable of resolving the host you are trying to reach: `avahi-resolve -n mbologna-uyuni.tf.local`
  - mdns is configured in glibc's Name Server Switch configuration file
 
 In `/etc/nsswitch.conf` you should see a `hosts:` line that looks like the following:
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -39,6 +39,15 @@ hosts:          files mdns [NOTFOUND=return] dns
 ```
 `mdns` (optionally suffixed with `4` for IPv4-only or `6` for IPv6-only) should be present in this line. If it is not, add it.
 
+Starting with `nss-mdns` version 0.14.1, you also need to populate `/etc/mdns.allow` with:
+
+```
+.local.
+.local
+```
+
+`mdns.allow` is required to [force all .local domains to be resolved regardless of label count or unicast SOA records](https://github.com/lathiat/nss-mdns/blob/master/README.md#etcmdnsallow).
+
 ## Q: how can I work around slowness in resolution of `tf.local` mDNS/Zeroconf/Bonjour/Avahi names?
 
 If there is a 5-second delay on any name resolution (or ping) between Avahi hosts, a likely cause is that ipv6 is enabled on the VMs (that is the default setting) but the network is blocking ipv6 traffic.
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -28,9 +28,9 @@ Could not resolve hostname server.tf.local: Name or service not known
 Check that:
  - your firewall is not blocking UDP port 5353
   - on SUSE systems check YaST -> Security and Users -> Firewall -> Zones -> public, "mdns" should appear on the list on the right
- - avahi is installed and running
+ - Avahi is installed and running
    - typically you can check this via systemd: `systemctl status avahi-daemon`
- - Avahi is capable of resolving the host you are trying to reach: `avahi-resolve -n mbologna-uyuni.tf.local`
+ - Avahi is capable of resolving the host you are trying to reach: `avahi-resolve -n uyuni.tf.local`
  - mdns is configured in glibc's Name Server Switch configuration file
 
 In `/etc/nsswitch.conf` you should see a `hosts:` line that looks like the following:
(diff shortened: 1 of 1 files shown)


witekest (commented):
I also observe this behavior on openSUSE 15.2. Populating `/etc/mdns.allow` fixes the problem as described.

TROUBLESHOOTING.md
```diff
@@ -30,6 +30,7 @@ Check that:
   - on SUSE systems check YaST -> Security and Users -> Firewall -> Zones -> public, "mdns" should appear on the list on the right
  - avahi is installed and running
    - typically you can check this via systemd: `systemctl status avahi-daemon`
+ - Avahi is capable of resolving the host you are trying to reach: `avahi-resolve -n mbologna-uyuni.tf.local`
```
rjmateus, line 33: `avahi-resolve` came instaled by default or should we point that `avahi-utils` package should be installed?

mbologna: The actual package name containing `avahi-resolve` might depend on the distribution though

Bischoff: mbologna-uyuni -> something more neutral

TROUBLESHOOTING.md
```diff
@@ -30,6 +30,7 @@ Check that:
   - on SUSE systems check YaST -> Security and Users -> Firewall -> Zones -> public, "mdns" should appear on the list on the right
  - avahi is installed and running
```
Bischoff, line 31: avahi -> Avahi

TROUBLESHOOTING.md
```diff
@@ -38,6 +39,15 @@ hosts:          files mdns [NOTFOUND=return] dns
 ```
 `mdns` (optionally suffixed with `4` for IPv4-only or `6` for IPv6-only) should be present in this line. If it is not, add it.
 
+Starting with `nss-mdns` version 0.14.1, you also need to populate `/etc/mdns.allow` with:
+
+```
+.local.
+.local
```
Bischoff, line 46: Rather this
```
.local
.local.tf
```

TROUBLESHOOTING.md
```diff
@@ -38,6 +39,15 @@ hosts:          files mdns [NOTFOUND=return] dns
...
+```
+
+`mdns.allow` is required to [force all .local domains to be resolved regardless of label count or unicast SOA records](https://github.com/lathiat/nss-mdns/blob/master/README.md#etcmdnsallow).
```
Bischoff, line 49: all .local domains -> all multicast DNS domains
(because we use it for `.local.tf`, not for `.local`)

TROUBLESHOOTING.md
agraul: In Debian (10) the package is also called `avahi-utils`, in Fedora (32) it is called `avahi-tools`. We could mention that `avahi-resolve` is typically in an additional package that is commonly suffixed `-tools` or `-utils`, depends on how throughout we want to be here.

Bischoff (changes requested):
some changes requested (a few nitpicks, but also something important i suggest using same contents for `/etc/mdns.allow` that sumaform uses :smile_cat: )

TROUBLESHOOTING.md
mbologna: I would not go into the details of which package contains avahi-resolve. It is also true that `cnf` (`command-not-found`) is taking care of suggesting the right package in case someone launches a command and the command is not found.

TROUBLESHOOTING.md
mbologna: Fixed

TROUBLESHOOTING.md
mbologna: Shouldn't that be:

`.local`
`.tf.local`

instead?

TROUBLESHOOTING.md
mbologna: Fixed

TROUBLESHOOTING.md
moio: FMPOV this can be merged as soon as this discussion is resolved, thanks @mbologna

TROUBLESHOOTING.md
agraul: Fair enough, we don't need to mention package names. JFYI `command-not-found` is not that great of a general solution, it is not pre-installed on all systems (at least not on the Debian and Fedora VMs I'm running).

Bischoff: What about my remark about not mentioning mbologna in tge example ?

TROUBLESHOOTING.md
Bischoff: er yes, sorry i inverted

TROUBLESHOOTING.md
Bischoff: I don't see the fix... but after all, `.tf.local` is also a `.local` domain... :-)

TROUBLESHOOTING.md
mbologna: I was waiting for https://github.com/uyuni-project/sumaform/pull/771#discussion_r490060362 to be answered before pushes the fixes. The requested fix is now pushed.

Bischoff (approved):
LGTM now

PR #775, opened 2020-09-23
Update centos8o with the latest released image

calancha:
## What does this PR change?
Update centos8o with the latest stable image (8.2.2004).

- backend_modules/libvirt/base/main.tf (centos8o):
Point url to the latest CentOS 8 image.


moio:
Feel free to merge as soon as you are really-really sure this won't break our testsuites :wink: !

calancha:
> Feel free to merge as soon as you are really-really sure this won't break our testsuites wink !

Sorry, I haven't run the testsuite with this change yet.
My main motivation for this change is that some QA colleagues might be using sumaform to create
the clients for the MU 4.1.2 validation (we received instructions about how to do this safely); in such a case they won't be testing the latest CentOS 8 as we should.

Bischoff (commented):
LGTM, but leaving the decision to Oscar.

srbarrios (approved):

PR #776, opened 2020-09-30
WIP Use mirror in cloudinit too

Bischoff:
## What does this PR change?

This PR changes `user_data.yaml` to use the mirror in the case of libvirt backend, like we already do with aws backend.

### DIFF
--- a/backend_modules/libvirt/host/main.tf
+++ b/backend_modules/libvirt/host/main.tf
@@ -27,7 +27,9 @@ locals {
 data "template_file" "user_data" {
   template = file("${path.module}/user_data.yaml")
   vars = {
-    image = var.image
+    image             = var.image
+    use_mirror_images = var.base_configuration["use_mirror_images"]
+    mirror            = var.base_configuration["mirror"]
   }
 }
 
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -109,7 +109,7 @@ zypper:
   repos:
     - id: os_pool_repo
       name: os_pool_repo
-      baseurl: http://download.suse.de/ibs/SUSE/Products/SLE-SERVER/12-SP4/x86_64/product/
+      baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-SERVER/12-SP4/x86_64/product/
       enabled: 1
       autorefresh: 1
 
@@ -121,7 +121,7 @@ zypper:
   repos:
     - id: os_pool_repo
       name: os_pool_repo
-      baseurl: http://download.suse.de/ibs/SUSE/Products/SLE-SERVER/12-SP5/x86_64/product/
+      baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-SERVER/12-SP5/x86_64/product/
       enabled: 1
       autorefresh: 1
 
@@ -133,7 +133,7 @@ zypper:
   repos:
     - id: os_pool_repo
       name: os_pool_repo
-      baseurl: http://download.suse.de/ibs/SUSE/Products/SLE-Module-Basesystem/15/x86_64/product
+      baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-Module-Basesystem/15/x86_64/product
       enabled: 1
       autorefresh: 1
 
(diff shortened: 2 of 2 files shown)


backend_modules/libvirt/host/user_data.yaml
```diff
@@ -109,7 +109,7 @@ zypper:
   repos:
     - id: os_pool_repo
       name: os_pool_repo
-      baseurl: http://download.suse.de/ibs/SUSE/Products/SLE-SERVER/12-SP4/x86_64/product/
+      baseurl: http://${mirror_url}/SUSE/Products/SLE-SERVER/12-SP4/x86_64/product/
```
srbarrios, line 112: Are we going to far if we also use `var.use_mirror_images` ?

Bischoff: I asked myself the same question. I don't know. I took model on aws backend, that does not use it.

@rjmateus? Any insights?

Bischoff: I used this other variable. Maybe in AWS you always have a mirror.

srbarrios (approved):

backend_modules/libvirt/host/user_data.yaml
rjmateus: I think It can be implemented as @srbarrios is saying. But we will need to pass both variables.
In AWS the configuration only uses mirror for the `download.suse.de` since it is not available on the cloud.

Bischoff: @rjmateus is the way I access both variables look okay for you?

moio, line 112: Please double check if this is correct. In this diff you are removing:

`download.suse.de/ibs`

and substituting it with

`${mirror_url}`

But according to `host/main.tf`:

`mirror_url      = var.base_configuration["use_mirror_images"] ? var.base_configuration["mirror"] : "download.suse.de"`

Note that if `use_mirror_images` is false then `mirror_url` becomes `download.suse.de`, which lacks the `/ibs` part.

backend_modules/libvirt/host/main.tf
```diff
@@ -27,7 +27,8 @@ locals {
 data "template_file" "user_data" {
   template = file("${path.module}/user_data.yaml")
   vars = {
-    image = var.image
+    image           = var.image
+    mirror_url      = var.base_configuration["use_mirror_images"] ? var.base_configuration["mirror"] : "download.suse.de"
```
moio, line 31: What is the plan for distros that have repos other than `download.suse.de`?

moio:
Thanks for coming up with this effort, it was on my to-do list since official image support landed!

backend_modules/libvirt/host/main.tf
srbarrios: Right.
@Bischoff I would go with the same implementation made here:
https://github.com/uyuni-project/sumaform/pull/776#discussion_r497489647

rjmateus: I agree with Moio and Oscar, we should be passing both variables to the user_data.yaml template and use them inside, so we can mirror all repos corretly.

rjmateus (commented):
Another comment. We should use the mirror for non suse products? I think we can use, and in this was Oscar's suggestion could be a path to follow.
Thanks for bringing this

backend_modules/libvirt/host/user_data.yaml
Bischoff: "oops"

backend_modules/libvirt/host/main.tf
Bischoff: ok, I could answer we don't have that for the moment, and YAGNI, but ok.

Bischoff:
Comments addressed. Better?

backend_modules/libvirt/host/main.tf
```diff
@@ -27,7 +27,9 @@ locals {
 data "template_file" "user_data" {
   template = file("${path.module}/user_data.yaml")
   vars = {
-    image = var.image
+    image           = var.image
+    use_mirror_images = var.base_configuration["use_mirror_images"]
+    mirror          = var.base_configuration["mirror"]
```
rjmateus, line 32: just one nitpick, these lines are missing formating :)

Bischoff: the equal signs were aligned with the series above, but ok, i agree it does not look good.

Bischoff:
> Thanks for coming up with this effort, it was on my to-do list since official image support landed!

Most welcome, Silvio.

rjmateus (approved):
Thanks for the great contribution

Bischoff:
Now tested to work.

PR #778, opened 2020-10-05
Allow unauthenticated packages on ubuntu 16.04

Bischoff:
## What does this PR change?

This PR solves a problem with ubuntu 16.04 official image:
```
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
```


Bischoff:
Tested to work

srbarrios (approved):

srbarrios:
LGTM. Just consider if that could happen in Ubuntu20.04, as it still also not tested.

Bischoff:
> LGTM. Just consider if that could happen in Ubuntu20.04, as it still also not tested.

I doubt that, because it does not happen on 18.04. Should it happen all the same, I'd do another PR.

moio (approved):

PR #779, opened 2020-10-05
Upgrade controller image to opensuse152o

srbarrios:
## What does this PR change?

This PR upgrades the controller default image from opensuse150o toopensuse152o

It might need additional packages here:
https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/host/user_data.yaml#L100-L101

And upgrade versions here:
https://github.com/uyuni-project/sumaform/blob/master/salt/controller/init.sls#L60-L68

But let's give it a try first without any more change.

### DIFF
--- a/modules/controller/main.tf
+++ b/modules/controller/main.tf
@@ -83,7 +83,7 @@ module "controller" {
   }
 
 
-  image   = "opensuse150o"
+  image   = "opensuse152o"
   provider_settings = var.provider_settings
 }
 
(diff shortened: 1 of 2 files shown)


moio:
Thanks for taking care of this! Feel free to ping as soon as this is ready for review :+1:

calancha (approved):
As long as it has been tested to work, then it looks fine to me.

PR #783, opened 2020-10-16
Reduce the content checkout from github to the ctl and checkout last 100 commits

srbarrios:
## What does this PR change?

Inside the controller, we only make use of the `testsuite` folder, so there is no need to download the rest of the repository.

Also, we increase the depth of the shallow-clone to 100 commits, so we can compare and navigate from the last 100 commits.
We need that in order to extract, during our test suite pipeline the product commit changes between to test suite executions (Internal card: https://github.com/SUSE/spacewalk/issues/13630)

Our current shallow clone takes about 354M, with these changes the repository folder will size around 60M.

### DIFF
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -129,7 +129,12 @@ spacewalk_git_repository:
 {%-     set url = 'https://github.com/SUSE/spacewalk.git' %}
 {%-   endif %}
 {%- endif %}
-    - name: git clone --depth 1 {{ url }} -b {{ branch }} /root/spacewalk
+    - name: |
+        git clone --depth 100 --no-checkout --branch {{ branch }} {{ url }} /root/spacewalk
+        cd /root/spacewalk
+        git sparse-checkout init --cone
+        git sparse-checkout set testsuite
+        git checkout
     - creates: /root/spacewalk
     - require:
       - pkg: cucumber_requisites


ktsamis (approved):
makes sense to me to reduce only to the folder we need. Let's try it and see if anything else is needed. Looks great though

calancha:
This looks like a nice improvement.
First we need to check that the Git version installed in the controller supports those options.
That will likely happen when the ctl is opensuse152o.
For opensuse150o it seems not the case, because that OS ships with Git v**2.16.4,** but the `--filter` since to appear in v**2.170**.

srbarrios:
> This looks like a nice improvement.
> First we need to check that the Git version installed in the controller supports those options.
> That will likely happen when the ctl is opensuse152o.
> For opensuse150o it seems not the case, because that OS ships with Git v**2.16.4,** but the `--filter` since to appear in v**2.170**.

Very good catch. This PR will need to wait then.

Bischoff (changes requested):
As Tino found out, this will need to wait a bit because of required git version.

Please also rebase it on top of latest version of sumaform. It should be relatively straightforward, you just need to modify the new `git_helper` script.

salt/controller/init.sls
```diff
@@ -105,12 +105,21 @@ spacewalk_git_repository:
   cmd.run:
 {%- if grains.get("git_repo") == "default" %}
 {%- if grains.get("branch") == "master" %}
-    - name: git clone --depth 1 https://github.com/uyuni-project/uyuni.git -b master /root/spacewalk
+    - name: |
+        git clone --depth 1 --filter=blob:limit=50K --no-checkout https://github.com/uyuni-project/uyuni.git /root/spacewalk
```
moio, line 109: Will this mean our setup could break when we add files bigger than 50k?

juliogonzalez: This doesn't fail, this filters.

If one of the "not cloned" files is part of the testsuite, then of course the testsuite will have problems running.

juliogonzalez (approved):
Other than the comments about the PR being blocked until we update the controller image, this LGTM.

salt/controller/init.sls
moio: > This doesn't fail, this filters.

Does filter mean such big files would not be downloaded? Because in that case I would assume the teststuite to fail. Did I get that right?

srbarrios: It might be indeed not the best solution @moio, have you another proposal to reduce the amount of data downloaded from github every time we prepare a controller? I have in mind things like a sub-module... but I don't like too much that approach.
We can also make use of released zips containing only testsuite folder but thenn we'll not be able to to push/pulls inside the controller...  So that was a kind of trade-off solution.

lkotek (approved):
LGTM :-)

salt/controller/init.sls
juliogonzalez: > Does filter mean such big files would not be downloaded? Because in that case I would assume the teststuite to fail. Did I get that right?

Can you why should it?

I don't expect testsuite files to be more than 50K, they are just text files with cucumber/ruby code. If for some reason they exceed it, we can adjust it.

> We can also make use of released zips containing only testsuite folder but thenn we'll not be able to to push/pulls inside the controller... So that was a kind of trade-off solution.

Not sure what you mean. Released zips are only generated with you create releases, and contain the whole code for a given commit, not only the testsuite folder. And while we could automate the creation, I don't think we should create GitHub releases all the time for a CI.

What you could s getting the full zip from the branch you want. That's what Eric did on the other PR. You will download the same in the end, but of course maybe it's more efficient from GitHub's point of view.

moio:
@srbarrios can you try out the approach described here? 

https://medium.com/pinterest-engineering/how-a-one-line-change-decreased-our-build-times-by-99-b98453265370

It seems safer than arbitrarily putting a size limit on blobs, and I wonder if it could yield to even bigger savings.

srbarrios:
> @srbarrios can you try out the approach described here?
> 
> https://medium.com/pinterest-engineering/how-a-one-line-change-decreased-our-build-times-by-99-b98453265370
> 
> It seems safer than arbitrarily putting a size limit on blobs, and I wonder if it could yield to even bigger savings.

On vacation today, but I'm sure that even if you only check the size of our repository without the .git folder, it will be bigger than 250 megas. So if we want to download less content, this will not solve anything.

moio:
Then I might suggest a sparse checkout, where one only downloads a subset of the directories, instead of filtering by size.

Seems to be possible: https://stackoverflow.com/questions/4114887/is-it-possible-to-do-a-sparse-checkout-without-checking-out-the-whole-repository

srbarrios:
> Then I might suggest a sparse checkout, where one only downloads a subset of the directories, instead of filtering by size.
> 
> Seems to be possible: https://stackoverflow.com/questions/4114887/is-it-possible-to-do-a-sparse-checkout-without-checking-out-the-whole-repository

Actually... this PR is not just filtering by size, but also downloading only the testsuite folder.
But I never listened about the sparse checkout and I will be happy to try it.

srbarrios:
@moio @rjmateus I refactored to use `sparse-checkout` as Moio suggested.
I tested manually and works. Can I merge it?

srbarrios:
As I want that working for this weekend, I'm going to merge it.
If it cause some troubles, I will revert on Monday.

PR #784, opened 2020-10-16
Work around problems with github clone

Bischoff:
## What does this PR change?

See SUSE/spacewalk#12753

### DIFF
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -103,14 +103,20 @@ fix_cucumber_html_reporter_style:
 
 spacewalk_git_repository:
   cmd.run:
+    - cwd: /tmp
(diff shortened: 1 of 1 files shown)


juliogonzalez (commented):
Why don't we use the same solution for the other repos instead of `git clone`?

Bischoff:
> Why don't we use the same solution for the other repos instead of `git clone`?

Because we have not the data in two separate variables "product" and "branch", we just have "github_repo".

I could do some parsing, but not the time for a quick fix.

juliogonzalez:
ACK, LGTM to them, as you see it was not a "requires changes" :-)

juliogonzalez (approved):

PR #785, opened 2020-10-16
Copy in all cases

Bischoff:
## What does this PR change?

Address last case for the github clone problem.


Bischoff:
Tested to work.

PR #786, opened 2020-10-18
Fallback mechanism

Bischoff:
## What does this PR change?

This PR tries a `git clone` first, and if that fails, switches to `curl`

### DIFF
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -101,32 +101,62 @@ fix_cucumber_html_reporter_style:
     - require:
       - cmd: install_cucumber_html_reporter_via_npm
 
-spacewalk_git_repository:
-  cmd.run:
-    - cwd: /tmp
-    - name: |
+git_helper:
+  file.append:
+    - name: ~/git_helper.sh
+    - text: |
 {%- set url = grains.get('git_repo') | default('default', true) %}
 {%- set branch = grains.get('branch') | default('master', true) %}
 {%- if url == 'default' %}
 {%-   if branch == 'master' %}
-{%-     set project = 'uyuni-project' %}
-{%-     set repo = 'spacewalk' %}
+{%-     set url = 'https://github.com/uyuni-project/uyuni.git' %}
 {%-   else %}
-{%-     set project = 'SUSE' %}
-{%-     set repo = 'spacewalk' %}
+{%-     set url = 'https://github.com/SUSE/spacewalk.git' %}
 {%-   endif %}
-{%- else %}
-{%-   set list = url.split('/') %}
-{%-   set project = list[-2] %}
-{%-   set repo = list[-1].replace('.git', '') %}
 {%- endif %}
-        curl -L -n https://github.com/{{ project }}/{{ repo }}/archive/{{ branch }}.tar.gz --output archive.tar.gz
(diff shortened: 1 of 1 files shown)


salt/controller/init.sls
```diff
@@ -101,32 +101,62 @@ fix_cucumber_html_reporter_style:
...
-        tar xzf archive.tar.gz {{ repo }}-{{ branch }}/testsuite
-        mv {{ repo }}-{{ branch }} /root/spacewalk
+{%- set list = url.split('/') %}
+{%- set project = list[-2] %}
+{%- set repo = list[-1].replace('.git', '') %}
+        git clone --depth 1 {{ url }} -b {{ branch }} /root/spacewalk
+        if [ $? -ne 0 ]; then
+          cd /tmp
+          echo '********************************************' >&2
+          echo '*** PROBLEM WITH GIT CLONE, PLEASE DEBUG ***' >&2
```
moio, line 124: What is the advantage of having a fallback when the build needs to be debugged anyways? Are build results reliable in case we used curl?

moio: Additional question: is it correct that this mechanism could in principle be removed if/when the root cause of https://infra.nue.suse.com/SelfService/Display.html?id=178256 is completely addressed?

srbarrios: If you ask me, I would let the system fail until we fix the problem on the infrastructure side.
The first implementation was to hide the issue by using an implementation that downloads a zip of the repo, but I disliked that approach and moreover as a permanent solution, as I want to be able to make pull/push of code inside the controller. So, we arrive at the trade-off to use it only as a fall-back, so we can continue to see the issue, and hopefully, address the root cause of it.

Most important about it, is that's only related to our private network, everyone else from their machine shouldn't have problems with github clones.

moio: In that case I will ask at least to put a //HACK comment that clearly says when the hack can be removed, ie, when the issue linked above is closed. That is the bare acceptable minimum FMPOV.

One permanent "solution", that would probably make possible to remove several other hacks, is to package the testsuite as rpm as with all other components of the project. That would make sense from an overall product design perspective and would work around this specific issue - although it brings in other trade-offs of course.

srbarrios: No please, don't go on that direction :D 
I want to be able to execute git pulls inside my controller, when testing using a branch.

moio: >  I want to be able to execute git pulls inside my controller, when testing using a branch.

Fair enough.

@Bischoff, would you at least agree about the `// HACK` comment I mentioned above?

Bischoff: I do agree. But:

This was a kind of tradeoff between not breaking testsuite results when the issue appears and still being aware of the issue reappearing.

But now that the issue seems to be gone for good, I am starting to wonder if that was such a good idea after all.
First, if we don't pay active attention, we might miss if the suite starts failing again. Second, as you, I don't like to leave unnecessary crap behind.

What about the following? I remove the curl part, and if it fails, it fails, and we can restore it, but at least we will catch the issue.

moio: I agree @Bischoff, thanks for taking care

Bischoff: ok, we have a plan. Doing the PR now.

PR #787, opened 2020-10-20
Changing selinux to permissive in CentOS official images

srbarrios:
## What does this PR change?

Changing `selinux` from enforcing to permissive in CentOS official images.
As we were having problems with that, @hustodemon knows more about it.

Note: 
Still not tested, I just created the PR together with the branch :)

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -35,6 +35,14 @@ runcmd:
   - "sed -i -e's/without-password/yes/' /etc/ssh/sshd_config"
   - "service sshd restart"
   - "service qemu-ga start"
+  
+write_files:
+  - content: |
+    SELINUX=permissive
+    SELINUXTYPE=targeted
+  owner: root:root
+  path: /etc/sysconfig/selinux
+  permissions: '0644'
 %{ endif }
 
 %{ if image == "centos7o" }
@@ -58,6 +66,14 @@ yum_repos:
     name: epel
 
 packages: ["salt-minion", "avahi", "nss-mdns", "qemu-guest-agent"]
+
+write_files:
+  - content: |
+    SELINUX=permissive
+    SELINUXTYPE=targeted
+  owner: root:root
+  path: /etc/sysconfig/selinux
+  permissions: '0644'
 %{ endif }
 
 %{ if image == "centos8o" }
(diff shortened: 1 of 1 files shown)


srbarrios:
Internal bug: https://bugzilla.suse.com/show_bug.cgi?id=1177928

srbarrios:
Testing it here:
https://ci.suse.de/view/Manager/view/Manager-4.1/job/manager-4.1-cucumber-PRV/425/console

It might very well not work, I never did changes on cloud-init so not sure if I did it using the right methods

moio:
@srbarrios, @hustodemon: reading the bug description it sounds like this seems to be a genuine product defect that could be hit by customers using the default configuration.

Do we really want to add a workaround here, thereby hiding the bug which is really a product bug?

Moreover, is there a particular reason why you are trying to change the selinux policy at cloud-init stage? Could it be changed later (Salt state or even testsuite code)?

hustodemon:
@moio

- this PR brings the SELinux state to what we had previously (older images did have SELinux **disabled** (not even enabled & permissive))
- this should be only a temporary workaround to make the testsuite stable for 4.1 release and to verify the hypothesis, that SELinux is causing trouble (i'm almost certain it does). if we support SELinux on clients, we should revert the change and fix the weird behavior (linking downstream bug here: https://bugzilla.suse.com/show_bug.cgi?id=1177928
- i think changing at cloud init lvl is not necessary: it should be enough to disable it with `setenforce 0` (and also changing it in `/etc/sysconfig/selinux`, if the configuration should survive reboot).

srbarrios:
> @srbarrios, @hustodemon: reading the bug description it sounds like this seems to be a genuine product defect that could be hit by customers using the default configuration.
> 
> Do we really want to add a workaround here, thereby hiding the bug which is really a product bug?
> 
> Moreover, is there a particular reason why you are trying to change the selinux policy at cloud-init stage? Could it be changed later (Salt state or even testsuite code)?



> @moio
> 
> * this PR brings the SELinux state to what we had previously (older images did have SELinux **disabled** (not even enabled & permissive))
> * this should be only a temporary workaround to make the testsuite stable for 4.1 release and to verify the hypothesis, that SELinux is causing trouble (i'm almost certain it does). if we support SELinux on clients, we should revert the change and fix the weird behavior (link to the downstream bug is above)
> * i think changing at cloud init lvl is not necessary: it should be enough to disable it with `setenforce 0` (and also changing it in `/etc/sysconfig/selinux`, if the configuration should survive reboot).

As far as I understood you confirmed already the hipothesis. If we have a confirmation that SElinux must be supported. Maybe we should increase the bug priority as this is blocking our CI. And forget about this workaround?

In any case, I would avoid moving the workaround at testsuite level, as in such case if another user of sumaform try migrate a traditional client to a minion, it will have the same issue than us.

Also, I would avoid using setenforce as we trigger several reboots in this system.

srbarrios:
If @mcalmer push his fix on mgr-daemon package, I would say we can close this Draft.

PR #791, opened 2020-11-04
Use latest release of minima on server

Bischoff:
## What does this PR change?

The SUSE Manager server uses minima to get a copy of the test packages

moio -> uyuni-project
v0.4 -> v0.11

Note: it would be better to get test packages directly, as already stated in SUSE/spacewalk#9429 . This would remove the minima dependency as a side effect.


srbarrios (approved):

moio (approved):

PR #792, opened 2020-11-04
Add support for SUSE Manager 4.2 (HEAD) and SLES15SP3

juliogonzalez:
## What does this PR change?

Add support for SUSE Manager 4.2 (HEAD) and SLES15SP3

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -169,6 +169,18 @@ zypper:
 packages: ["avahi", "nss-mdns", "qemu-guest-agent"]
 %{ endif }
 
+%{ if image == "sles15sp3o" }
+zypper:
+  repos:
+    - id: os_pool_repo
+      name: os_pool_repo
+      baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-Module-Basesystem/15-SP3/x86_64/product
+      enabled: 1
+      autorefresh: 1
+
+packages: ["avahi", "nss-mdns", "qemu-guest-agent"]
+%{ endif }
+
 %{ if image == "ubuntu1604o" }
 apt:
   conf: |
--- a/salt/mirror/init.sls
+++ b/salt/mirror/init.sls
@@ -51,8 +51,8 @@ tar:
 jdupes:
   archive.extracted:
     - name: /
-    - source: https://github.com/jbruchon/jdupes/releases/download/v1.13.2/jdupes-1.13.2-linux64-static.tar.xz
-    - source_hash: 078531eee63b434f3b8383367050a55d18fdd671c4d7800df82da9630a35ea62fc755e6337bd703d6526d3ae265f4d37b218aa1045629809498056a59bda3922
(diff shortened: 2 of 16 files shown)


moio:
Thanks for taking care of this! Please also check https://github.com/uyuni-project/sumaform/blob/master/salt/mirror/minima.yaml

juliogonzalez:
> Thanks for taking care of this! Please also check https://github.com/uyuni-project/sumaform/blob/master/salt/mirror/minima.yaml

It's one of the changes on the PR already... or am I missing anything :-?

moio:
That's right, I overlooked it somehow, apologies

juliogonzalez:
Creating the minima mirror and creating a cucumber environment worked locally.

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -169,6 +169,18 @@ zypper:
 packages: ["avahi", "nss-mdns", "qemu-guest-agent"]
 %{ endif }
 
+%{ if image == "sles15sp3o" }
+zypper:
```
juliogonzalez, line 173: Should we add the moving target repo here as well?

salt/mirror/init.sls
```diff
@@ -51,8 +51,8 @@ tar:
 jdupes:
   archive.extracted:
     - name: /
-    - source: https://github.com/jbruchon/jdupes/releases/download/v1.13.2/jdupes-1.13.2-linux64-static.tar.xz
-    - source_hash: 078531eee63b434f3b8383367050a55d18fdd671c4d7800df82da9630a35ea62fc755e6337bd703d6526d3ae265f4d37b218aa1045629809498056a59bda3922
+    - source: https://github.com/jbruchon/jdupes/releases/download/v1.18.1/jdupes_1.18.1-x86-64.pkg.tar.xz
```
juliogonzalez, line 54: This is needed because the mirror creation is basically broken without it. `1.13.2` is not available anymore as the author removed it because of a security vulnerability.

backend_modules/libvirt/host/user_data.yaml
moio: If such repo is necessary in order to install packages below (which I would expect) then yes, there is no other way really.

salt/mirror/init.sls
moio: I am totally fine with that.

moio (approved):
I did not notice anything strange here, feel free to move on after testing.

Thanks for all the work here!

calancha (approved):
It looks also OK to me.  Thank you very much.

backend_modules/libvirt/host/user_data.yaml
juliogonzalez: It seems not really, the deployment worked fine.

AFAIK this means you will get the versions released, and not those from the moving target.

But since such packages are not using for testing, we should be fine, right?

moio: Not really. The QEMU agent is needed depending on the DHCP configuration of your network, and deployment will fail in certain situations, same for avahi which is optional.

Please log into a deployed machines and see which of these three packages are there and which are missing. If any are missing, then we really need the repos which provide them (any version really).

PR #795, opened 2020-11-07
Reduce noise in the sumaform log

calancha:
## What does this PR change?
This prevents from lines as the following one in the output:
... minion.module.host.null_resource.provisioning[0] (remote-exec): /root/salt/wait_for_salt.sh: line 7: salt-call: command not found

- salt/wait_for_salt.sh:
Redirect both, stdout and stderr to /dev/null.


moio (approved):
lgtm, thanks!

PR #796, opened 2020-11-07
Fix some mirror repo urls

calancha:
## What does this PR change?

Fix some mirror repo urls.
- salt/repos/proxy.sls
- salt/repos/server.sls
- salt/repos/virthost.sls
- salt/repos/default.sls:
The prefix `/ibs/` is also part of the url in the mirror for those urls as `/ibs/SUSE:/...`

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -318,7 +318,7 @@ os_update_repo:
 # Moving target, only until SLE15SP3 GA is ready
 os_movingtarget_repo:
   pkgrepo.managed:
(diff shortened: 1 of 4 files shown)


juliogonzalez (commented):
I think something is not right here, it could even be a bug on minima or something.

Howe comes this is is different only for SLE15SP3? If you check the sls files, you will see the `/ibs` part is present *always*.

Maybe whoever added the config for the PRV mirror did something wrong?

juliogonzalez (approved):
Ignore my previous comment, I understood everything the other way around.

In the end this seems to be my fault. I don't know why I added `/ibs` only for `download.opensuse.org`.

Thanks a lot for the fix. LGTM.

PR #797, opened 2020-11-16
Support mirroring OBS packages in our cloudinit

srbarrios:
## What does this PR change?

Add support to mirror OBS repositories.

What this PR doesn't do is to suppport mirror for other repos like:
```
baseurl: http://download.fedoraproject.org/pub/epel/6/$basearch
mirrorlist: https://mirrors.fedoraproject.org/metalink?repo=epel-6&arch=$basearch
```


Bischoff (approved):
LGTM

PR #799, opened 2020-11-18
Create /srv/mirror in mirror.sh script

jordimassaguerpla:
## What does this PR change?

Create /srv/mirror in mirror.sh script, otherwise, the "chmod -r 777" gets executed in / !!!


moio:
Shouldn't this be already handled by https://github.com/uyuni-project/sumaform/blob/master/salt/mirror/init.sls#L126-L134 ?

juliogonzalez:
@moio yes, but what if you use mirror.sh without salt or using other states?

That's what happened at the internal infra. They used it at an instance that didn't include that on the salt states, and caused a disaster.

So besides fixing that on the salt state, having this doesn't hurt.

juliogonzalez (approved):

PR #801, opened 2020-11-21
Let the test suite install OpenSCAP packages

Bischoff:
## What does this PR change?

The openSCAP packages were installed half in sumaform, half in the test suite, leading to a huge mess.

This situation is due to the fact that the QAM test suite does not use sumaform's `testsuite` module. But besides this immediate reason, it has always been wrong to let sumaform prepare ulterior processings. sumaform should prepare machines as pristine as possible.

This PR lets the test suite do all the job. See uyuni-project/uyuni#2880 for implementation on test suite side.

### DIFF
--- a/salt/client/testsuite.sls
+++ b/salt/client/testsuite.sls
@@ -9,22 +9,17 @@ client_cucumber_requisites:
     - pkgs:
       - spacewalk-client-setup
       - spacewalk-check
-      - spacewalk-oscap
-      - openscap-utils
       - mgr-cfg-actions
       - wget
     - require:
       - sls: default
 
-{% if grains['os'] == 'SUSE' %}
+{% if grains['os'] == 'SUSE' and '12' in grains['osrelease'] %}
 
 suse_client_cucumber_requisites:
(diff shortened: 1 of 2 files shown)


srbarrios (commented):
LGTM. We should even remove more packages

Bischoff:
> LGTM. We should even remove more packages

Completly agree. Created a card:  SUSE/spacewalk#13199

salt/client/testsuite.sls
```diff
@@ -9,22 +9,17 @@ client_cucumber_requisites:
...
       - spacewalk-check
-      - spacewalk-oscap
-      - openscap-utils
       - mgr-cfg-actions
       - wget
     - require:
       - sls: default
 
-{% if grains['os'] == 'SUSE' %}
+{% if grains['os'] == 'SUSE' and '12' in grains['osrelease'] %}
```
moio, line 17: Is `aaa_base-extras` something exclusively for SLE 12? Ie. not needed/present on SLE 15?

PR #804, opened 2020-11-25
Add 4.2-beta to sumaform

juliogonzalez:
## What does this PR change?

Adds something similar to what `4.2-released` will be, but using beta channels for the clients.

So if you specify `4.2-beta`, you get:

- Server: SLE15SP3 pool + update repos (for whatever milestone is out) + 4.2 beta pool + update repos  (for whatever milestone is out)
- Proxy: SLE15SP3 pool + update repos  (for whatever milestone is out) + 4.2 beta pool + update repos  (for whatever milestone is out)
- Client tools: The usual OS repositories + pool and/or beta repos (depending on the OS, for whatever milestone is out)

Requested by QA so they can use it for the QAM testsuite during milestone validations (https://github.com/SUSE/spacewalk/issues/13202)

Also:
- Simplify Ubuntu repo management

Tests:
In all cases I used a `tf` file that defined: Server, Proxy, SLE11 Traditional Cllient, SLE15 salt client, Ubuntu 18.04 salt client, CentOS7 salt client.

- [x] Provision `4.1-nightly` and inspect repositories on the instances
- [x] Provision `4.1-released` and inspect repositories on the instances
- [x] Provision `4.2-beta` and inspect repositories on the instances (NOTE: `4.2-beta` server alone, since deployment fails but it's enough to check repos, and then `4.1-released` + rest of `4.2-beta` instances.)

NOTE: I could not complete the `4.1-nightly` and `4.1-released` deployments because it seems I have some problems with the DNS resolution even after enabling avahi (so the proxy and clients can't resolve the server hostname). I am pretty sure this is related to my local dnsmasq. Will try to fix it other day. For now the inspection of repositories should be enough.

NOTE: Adding Beta for subsequent versions of SUSE Manager will be easy. Nothing to do for the clients, and we just need to add the final channel names for the next version at minima, and the final URLs at the states for server and proxy. Remember: the beta channels for server and proxy become the "release channels" on GA day. No changes to the names or the URLs. In other words, this PR is pretty much also adding everything that will be needed for `4.2-released` and part of what will be needed for `4.2-nightly`.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -10,6 +10,7 @@ Legal values for released software are:
  * `3.2-released`   (latest released Maintenance Update for SUSE Manager 3.2 and Tools)
  * `4.0-released`   (latest released Maintenance Update for SUSE Manager 4.0 and Tools)
  * `4.1-released`   (latest released Maintenance Update for SUSE Manager 4.1 and Tools)
+ * `4.2-beta`       (beta releases for SUSE Manager 4.2 and Tools)
  * `uyuni-released` (latest released version for Uyuni Server, Proxy and Tools, from systemsmanagement:Uyuni:Stable)
 
 Legal values for work-in-progress software are:
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -68,20 +68,26 @@ test_update_repo:
 tools_pool_repo:
   pkgrepo.managed:
     {% if grains.get('mirror') %}
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/SLES11-SP4-SUSE-Manager-Tools-BETA/sle-11-x86_64/
+    {% else %}
     - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/SLES11-SP4-SUSE-Manager-Tools/sle-11-x86_64/
+    {% endif %}
+    {% else %}
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://euklid.nue.suse.com/mirror/SuSE/build-ncc.suse.de/SUSE/Updates/SLE-SERVER/11-SP4-CLIENT-TOOLS-BETA/x86_64/update/
     {% else %}
     - baseurl: http://euklid.nue.suse.com/mirror/SuSE/build-ncc.suse.de/SUSE/Updates/SLE-SERVER/11-SP4-CLIENT-TOOLS/x86_64/update/
     {% endif %}
+    {% endif %}
 
 {% if 'nightly' in grains.get('product_version') | default('', true) %}
-
 tools_additional_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/Devel:/Galaxy:/Manager:/4.1:/SLE11-SUSE-Manager-Tools/images/repo/SLE-11-SP4-CLIENT-TOOLS-ia64-ppc64-s390x-x86_64-Media1/suse/
     - priority: 98
 
 {% elif 'head' in grains.get('product_version') | default('', true) %}
-
 tools_additional_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/Devel:/Galaxy:/Manager:/Head:/SLE11-SUSE-Manager-Tools/images/repo/SLE-11-SP4-CLIENT-TOOLS-ia64-ppc64-s390x-x86_64-Media1/suse/
@@ -205,11 +211,19 @@ test_update_repo:
 {% if not grains.get('product_version') or not grains.get('product_version').startswith('uyuni-') %}
 tools_pool_repo:
   pkgrepo.managed:
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Manager-Tools/12-BETA/x86_64/product/
+    {% else %}
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Manager-Tools/12/x86_64/product/
+    {% endif %}
 
 tools_update_repo:
   pkgrepo.managed:
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Manager-Tools/12-BETA/x86_64/update/
+    {% else %}
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Manager-Tools/12/x86_64/update/
+    {% endif %}
 {% else %}
 tools_pool_repo:
   pkgrepo.managed:
@@ -245,11 +259,19 @@ tools_additional_repo:
 {% if not grains.get('product_version') or not grains.get('product_version').startswith('uyuni-') %}
 tools_pool_repo:
   pkgrepo.managed:
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Manager-Tools/15-BETA/x86_64/product/
+    {% else %}
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Manager-Tools/15/x86_64/product/
+    {% endif %}
 
 tools_update_repo:
   pkgrepo.managed:
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Manager-Tools/15-BETA/x86_64/update/
+    {% else %}
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Manager-Tools/15/x86_64/update/
+    {% endif %}
 {% else %}
 tools_pool_repo:
   pkgrepo.managed:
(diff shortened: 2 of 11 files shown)


README_ADVANCED.md
```diff
@@ -10,6 +10,7 @@ Legal values for released software are:
  * `3.2-released`   (latest released Maintenance Update for SUSE Manager 3.2 and Tools)
  * `4.0-released`   (latest released Maintenance Update for SUSE Manager 4.0 and Tools)
  * `4.1-released`   (latest released Maintenance Update for SUSE Manager 4.1 and Tools)
+ * `4.2-beta`       (beta releases for SUSE Manager 4.2 and Tools)
```
moio, line 13: Do you see any pros/cons with using `4.2-nightly`?

juliogonzalez: Yes, that this is intended to be based on the beta releases and not on the Devel Projects.

We'll need this for further versions, so we need to add support for both `released` and `beta`

BTW: We should really get rid of `nightly` and call it `devel`. Such builds have nothing to do with `nighly`byulds at all :-D

moio: > BTW: We should really get rid of nightly and call it devel. Such builds have nothing to do with nighlybyulds at all :-D

Yep, this dates back to pre-Terraform days where the then-called suminator could deploy Spacewalk. In that context, `nightly` did make sense.

Then I never had the courage to break syntax for all users any more!

juliogonzalez: TBH: I always tell this, and I always afraid of the change. Maybe one day I will just grep 'nightly', see if I can do a sed, prepare, test one version, create PR and then announce a breaking change :-D

But same story as always: need to find time, and in the end this is not the most important thing in the world.

juliogonzalez:
@moio, as far as I could see during my tests, the repositories on the instances are correct. I hope the tests are enough.

README_ADVANCED.md
moio: > Yes, that this is intended to be based on the beta releases and not on the Devel Projects.

Then how about using `-released`? In the end, they are released. IIRC this is how we went in the past.

Eventually the released version becomes the publicly released one.

juliogonzalez: No, we need `-beta`. The client tools use different channels for beta (but not server and proxxy), and with this I can looks for `beta` to use them, and I don't need to change it for future SUSE Manager versions.

Otherwise I'd need to look for `4.2-released`, and then change it for SUSE Manager 5.0.

moio: > Otherwise I'd need to look for 4.2-released, and then change it for SUSE Manager 5.0.

Yeah, that's the suggestion. Won't we have a 5.0-released in any case in future?

juliogonzalez: Yes, but I don't want to change the code for the client tools for the next 5.0 beta. Doing it one time was enough nightmare.

When 5.0 beta arrives I will only need to add it to the tf files, default repos (so the new SLE is added), server and proxy, and that's it.

Client tools will just work.

Because, again:

> Eventually the released version becomes the publicly released one.

Half true. For Server and Proxy it's true, for the client tools it's not (we have separate channels for beta)

moio: Hmm, okay. Go ahead.

salt/repos/default.sls
```diff
@@ -68,20 +68,26 @@ test_update_repo:
...
+    {% else %}
     - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/SLES11-SP4-SUSE-Manager-Tools/sle-11-x86_64/
+    {% endif %}
+    {% else %}
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://euklid.nue.suse.com/mirror/SuSE/build-ncc.suse.de/SUSE/Updates/SLE-SERVER/11-SP4-CLIENT-TOOLS-BETA/x86_64/update/
     {% else %}
     - baseurl: http://euklid.nue.suse.com/mirror/SuSE/build-ncc.suse.de/SUSE/Updates/SLE-SERVER/11-SP4-CLIENT-TOOLS/x86_64/update/
     {% endif %}
+    {% endif %}
```
srbarrios, line 82: I suggest to simplify those if statements as we do just bellow in line 215:
```
    {% if 'beta' in grains.get('product_version') | default('', true) %}
    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Manager-Tools/12-BETA/x86_64/product/
    {% else %}
    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Manager-Tools/12/x86_64/product/
    {% endif %}
```

salt/repos/default.sls
```diff
@@ -384,20 +410,36 @@ tools_pool_repo:
...
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/RES/{{ release }}-CLIENT-TOOLS-BETA/x86_64/product/
+    {% else %}
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/RES/{{ release }}-CLIENT-TOOLS/x86_64/product/
+    {% endif %}
     {% elif grains.get('mirror') %}
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/RES{{ release }}-SUSE-Manager-Tools-BETA/x86_64/
+    {% else %}
     - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/RES{{ release }}-SUSE-Manager-Tools/x86_64/
```
srbarrios, line 422: I'm not sure if I understand it, is this folder pointing to the Updates as it does the else section?
I would say we can simplify it too. As we do above in line 414.
If the release is >= 8, then we use the `/SUSE/Product/RES/` else we use `SUSE/Updates/RES/` and we add the distinction of mirror inside the valirable passed like `{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}`

juliogonzalez: This can be refactored, I guess, but let's do it outside this PR as it will require further checks and time investment.

AFAIK this is related to how minima mirror works.

salt/repos/default.sls
juliogonzalez: This is SLE11, not 12 (line 215), and AFAIK this is the same magic we need for CentOS/RHEL < 8.

Let's have this done on a separate PR (to be done by someone with more experience about why we need separate URLs for the mirrros for SLE11 and CentOS/RHEL < 8.

srbarrios: The line was an example.
The refactor proposed is to use `{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}` instead of those embedded if statements

salt/repos/default.sls
srbarrios: So, you confirm me that there is no issue on NOT providing the updates repository path on lines 420 and 422?

salt/repos/default.sls
```diff
@@ -433,7 +475,11 @@ tools_pool_repo:
 tools_update_repo:
   pkgrepo.managed:
     - humanname: tools_update_repo
-    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/Devel:/Galaxy:/Manager:/Head:/RES{{ release }}-SUSE-Manager-Tools/SUSE_RES-{{ release }}_Update_standard/
+    {% if 'beta' in grains.get('product_version') | default('', true) %}
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/RES/{{ release }}-CLIENT-TOOLS-BETA/x86_64/update/
+    {% else %}
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/RES/{{ release }}-CLIENT-TOOLS/x86_64/update/
```
rjmateus, line 481: In this case shouldn't we use the Devel repository as we used to do before, for the case we are not in a beta version?
Also this code is inside an if that checks if we are using the 'head' product version, which will not contain also 'beta' in its name.

juliogonzalez: You are right. This is a bug.

juliogonzalez: Fixed. I didn't notice this problems during my tests because I didn't test head, as you can see at the description.

rjmateus (approved):

salt/repos/default.sls
srbarrios:
```suggestion
    - baseurl: http://{{ grains.get("mirror") }}/SUSE/Updates/RES/{{ release }}-CLIENT-TOOLS-BETA/x86_64/update/
    {% else %}
    - baseurl: http://{{ grains.get("mirror") }}/SUSE/Updates/RES/{{ release }}-CLIENT-TOOLS/x86_64/update/
```

salt/repos/default.sls
juliogonzalez: Agreed, but no more refactors for this PR. It's already big enough.

salt/repos/default.sls
juliogonzalez: To be honest, I think I didn't read correctly your original comment.

This part of the code is about the product/pool. My changes are only using the BETA product/pool if the product_version contains `beta`.

Updates are managed below (425-429) for <=7 and at 438 to 442 for RES8.

srbarrios: So the whole section `{% elif grains.get('mirror') %}` has the assumption of being minor than RES8, as the first `if` statement is about `{% if release >= 8 %}` what I'm pointing out here, is that we are using `product/pool` paths instead of `updates` paths, when we know that this section will not apply for RES8.
I got it now that if we jump inside the `'beta' in grains.get('product_version') ` we must use the `product/pool` path, but what about in the second case?

juliogonzalez: If it's the second  case, the update repository is managed below. But still keep in mind my PR is not changed how this was working before, and I am not even the original author for that.

All I am doing here is: if product contains "beta", use beta repositories, otherwise use the released repositories.

So if this is not working fine, we should fix it with another PR.

rjmateus: IMHO, PR should have a single purpose (if possible). In this case, refactoring the section about centos8, for me is out of the scope of these changes, which is focused on unable beta repos.
We should move forward with the changes in the beta channels, and a new PR opened to refactor this, if needed. 
If we do to much in one single PR, it will be much hard to track problems.
ps:  Sure we have exceptions, and in sumaform I have made several PR's that changed a lot, especially during sumaform refactoring modules.

srbarrios: Fine with it, just remarking again I didn't spotted only a possible refactor here but and issue on the repository provided if we have a mirror grain and RES version is lower than 8. It should be Updates, and we provide Pool.

juliogonzalez:
I will merge on Monday, just in case.

PR #805, opened 2020-11-26
Troubleshooting: Add tip about apparmor bug

ricardoasmarques:
Signed-off-by: Ricardo Marques <rimarques@suse.com>

## What does this PR change?

Add tip about apparmor bug

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -48,6 +48,12 @@ Starting with `nss-mdns` version 0.14.1, you also need to populate `/etc/mdns.al
 
 `mdns.allow` is required to [force all multicast DNS domains to be resolved regardless of label count or unicast SOA records](https://github.com/lathiat/nss-mdns/blob/master/README.md#etcmdnsallow).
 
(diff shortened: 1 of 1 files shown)


Bischoff (approved):
LGTM

rjmateus (approved):

ricardoasmarques:
Given the following includes that I have, by default, on my `Tumbleweed` host:

```
# grep 'include <abstractions/nameservice>' /etc/apparmor.d/usr.sbin.avahi-daemon | wc -l
1

# grep 'include <abstractions/mdns>' /etc/apparmor.d/abstractions/nameservice | wc -l
1

# grep '/etc/mdns.allow r' /etc/apparmor.d/abstractions/mdns | wc -l
1
```
it should not be needed to modify `/etc/apparmor.d/usr.sbin.avahi-daemon` as sugested in this PR.

To confirm that, I've reverted the `/etc/apparmor.d/usr.sbin.avahi-daemon` change on my host (the one where I had the issue), rebooted the host and I'm still able to resolve the VM by hostname, so I'm closing this PR for now, because apparently, it's not needed. 

(maybe the issue I had was resolved by a `reboot` and not by the change on `apparmor` configuration)

moio:
Thanks for the deep dive anyway! Let's keep this PR in mind in case somebody reports issues with Tumbleweed.

Bischoff:
Yes, then it was certainly fixed by the reboot, and we will never know what was wrong :-( .

PR #810, opened 2020-12-02
Fix 'sles15sp3o' image id

ricardoasmarques:
## What does this PR change?

This PR fixes the **SLES15SP3** image id introduced in https://github.com/uyuni-project/sumaform/pull/792.

The image id should be `sles15sp3o` instead of `sle15sp3o`, otherwise the following error is raised during `terraform apply`, when `images` is not explicitly set on `main.tf`:

Screenshot from 2020-12-02 14-37-10

### DIFF
--- a/modules/cucumber_testsuite/variables.tf
+++ b/modules/cucumber_testsuite/variables.tf
@@ -31,7 +31,7 @@ variable "name_prefix" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
-  default     = ["centos7o", "opensuse150o", "opensuse151o", "opensuse152o", "sles15o", "sles15sp1o", "sles15sp2o", "sle15sp3o", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4o", "ubuntu1804o"]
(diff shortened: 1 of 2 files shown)


moio (approved):

juliogonzalez (approved):
Looks good, but can you also change https://github.com/juliogonzalez/susemanager-ci/tree/terracumber/terracumber_config/tf_files  to use `sles15sp3o` as well (just use sed and create a PR)? Otherwise the cucumber and reference jenkins jobs will fail

ricardoasmarques:
> Looks good, but can you also change https://github.com/juliogonzalez/susemanager-ci/tree/terracumber/terracumber_config/tf_files to use `sles15sp3o` as well (just use sed and create a PR)? Otherwise the cucumber and reference jenkins jobs will fail

@juliogonzalez I don't see any reference that needs to be replaced in that repo/branch:

```
# cd susemanager-ci

# git checkout juliogonzalez/terracumber

# grep -r sle15sp3o | wc -l
0

# grep -r sles15sp3o | wc -l
0
```

Am I missing something? Thanks.

juliogonzalez:
We were lucky, and that's why we didn't have problems with the CI.

Now I remember we use SLE15SP3 only for the server and the proxy, and this fix is basically for pretty much everything else IIUC.

Feel free to merge the PR :-)

PR #812, opened 2020-12-07
Update variables with full list of images available

ktsamis:
## What does this PR change?

Update variables with full list of images available for libvirt and null provider. These are already available in https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/base/main.tf#L5, this PR is just documenting it


moio:
The omission of less-frequently-used images by default is on purpose - to save people's disk space unless they really need a certain image.

What need are you trying to address here, that can maybe addressed differently?

ktsamis:
I don't know what I was trying here but it makes no sense to download everything by default, closing

PR #814, opened 2020-12-08
Azure module

japelbaum:
## What does this PR change?

I have added the azure module so we can use sumaform to also provision an environment on azure.
I have built this with general public in mind as opposed to internal use.

TODO: add snapshot support for mirror image
TODO: Add README and advanced config info


japelbaum:
corrected main.tf in network module

backend_modules/azure/README.md
```diff
@@ -0,0 +1,138 @@
+ # AWS-specific configuration
```
rjmateus, line 1: This file is a copy from AWS and I think it doesn't apply the same way to azure integration.

backend_modules/azure/README_ADVANCED.md
```diff
@@ -0,0 +1,154 @@
+# Advanced configuration
```
rjmateus, line 1: I think this file is also a copy from AWS backend, I'm not sure if all is content makes sense on the azure world.

rjmateus:
Thanks for your contribution @japelbaum 
One missing piece. 
Base and host network should have a symbolic link for variables file (`variables.tf`) pointing to the variables file in the null backend_module. 
This file is used as a kind of interface to insure compatibility between all backends.
I'm planning to submit a full review during the current week

backend_modules/null/host/variables.tf
```diff
@@ -47,6 +47,14 @@ variable "quantity" {
   default     = 1
 }
 
+#The azure module needs this in order to use count in the creation of network cardsand public ips
+variable "public_instance" {
```
rjmateus, line 51: If this variable is needed for the azure environment it should be passed inside the `provider_settings` properties map.
It is also used in AWS, but is not something generic for all providers so it should leave inside provider-specific setting

backend_modules/azure/network/main.tf
```diff
@@ -0,0 +1,203 @@
...
+  name                        = "https"
+  resource_group_name         = "${azurerm_resource_group.suma-rg.name}"
+  network_security_group_name = "${azurerm_network_security_group.public-nsg[0].name}"
+  priority                    = 110
+  direction                   = "Inbound"
+  access                      = "Allow"
+  protocol                    = "Tcp"
+  source_port_range           = "*"
+  destination_port_range      = "443"
+  source_address_prefix       = "*"
```
rjmateus, line 123: shouldn't the `var.ssh_allowed_ips` be used also in here?

backend_modules/azure/network/main.tf
```diff
@@ -0,0 +1,203 @@
...
+  zones               = ["1"]
+}
+
+resource "azurerm_nat_gateway" "suma-ngw" {
+  count = var.create_network? 1: 0
+
+  name                = "${var.name_prefix}-ngw"
+  resource_group_name = "${azurerm_resource_group.suma-rg.name}"
+  location            = "${azurerm_resource_group.suma-rg.location}"
+  public_ip_address_ids   = ["${azurerm_public_ip.nat-pubIP[0].id}"]
```
rjmateus, line 181: this field is marked as deprecated and will be dropped in the future. It should be refactored to the replaced method: 
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/nat_gateway#public_ip_address_ids

backend_modules/azure/network/main.tf
```diff
@@ -0,0 +1,203 @@
...
+
+resource "azurerm_subnet_nat_gateway_association" "suma-ngw-association" {
+  count = var.create_network? 1: 0
+
+  subnet_id      = "${azurerm_subnet.private-sn[0].id}"
+  nat_gateway_id = "${azurerm_nat_gateway.suma-ngw[0].id}"
+}
+
+output "configuration" {
+  depends_on = [azurerm_subnet_route_table_association.private-rtas, azurerm_subnet_route_table_association.private_additional-rtas , azurerm_subnet_route_table_association.public-rtas]
```
rjmateus, line 192: the resource name is different from the one defined above. https://github.com/uyuni-project/sumaform/pull/814/files#diff-cb7f71c29f6b3f79b7e7e22c7796c9960887631cbedec34d98b96405e2c74827R80
```suggestion
  depends_on = [azurerm_subnet_route_table_association.private-rtas, azurerm_subnet_route_table_association.private-additional-rtas , azurerm_subnet_route_table_association.public-rtas]
```

backend_modules/azure/network/variables.tf
```diff
@@ -0,0 +1,25 @@
...
+  type        = string
+}
+
+variable "name_prefix" {
+  description = "resource name prefix"
+  type        = string
+}
+
+variable "create_network" {
+  description = "resource name prefix"
```
rjmateus, line 12:
```suggestion
  description = "defined if a new network should be created"
```

backend_modules/azure/network/variables.tf
```diff
@@ -0,0 +1,25 @@
+variable "location" {
+  description = "location where the instance is created"
+  type        = string
+}
+
+variable "name_prefix" {
+  description = "resource name prefix"
```
rjmateus, line 7: should this variable have a default value or is it mandatory?

backend_modules/azure/network/versions.tf
```diff
@@ -0,0 +1,6 @@
+terraform {
+  required_version = "~> 0.12.9"
+  required_providers {
+    aws = "~> 2.51"
```
rjmateus, line 4: this should be replaced by the needed azure provider version.

backend_modules/azure/host/versions.tf
```diff
@@ -0,0 +1,7 @@
+terraform {
+  required_version = "~> 0.12.9"
+  required_providers {
+    null = ">= 2.1.0"
+    aws  = "~> 2.51"
```
rjmateus, line 5: this should be replaced by the needed azure provider version.

salt/server/init.sls
```diff
@@ -20,7 +20,7 @@ server_packages:
     - refresh: True
     {% if grains['osfullname'] == 'Leap' %}
     - name: patterns-uyuni_server
-    {% else %}
+    {% elif grains.get('product_version', 'default') != 'suma41' %}
```
rjmateus, line 23: What is the goal of this change? Not install suma patter on product type suma41? 
IIUC you are adding/considering a new product_version named `suma41` which is a cloud image with suma already installed?
@moio in this case should we consider create a new product_version category, like `cloud-*` and do not install suma pattern if product starts with this prefix?

salt/default/minimal.sls
```diff
@@ -11,7 +11,9 @@ include:
 minimal_package_update:
   pkg.latest:
     - pkgs:
+{% if grains.get('product_version', 'default') != 'suma41' %}
```
rjmateus, line 14: The same question as below.

rjmateus (commented):
Thanks for your contribution. 
I'm dropping the first set of questions. Tomorrow I'm planning anew round of revisions. 
I don't have an azure account so I cannot test the execution. Did you know where I can get/ask for azure credentials to test the deployment?

rjmateus:
Can you please rebase again from master. We have merged a PR that should correct the errors on the github actions.

backend_modules/azure/host/main.tf
```diff
@@ -0,0 +1,240 @@
...
+  platform_image = lookup(lookup(var.base_configuration["platform_image_info"], var.image, {}), "platform_image", var.image)
+
+  provider_settings = merge({
+    public_key_location        = var.base_configuration["public_key_location"]
+    key_file                   = var.base_configuration["key_file"]
+    ssh_user                   = lookup(lookup(var.base_configuration["platform_image_info"], var.image, {}), "ssh_user", "azureuser")
+    public_instance = false
+    volume_size     = 50
+    bastion_host    = lookup(var.base_configuration, "bastion_host", null)
+    vm_size = lookup(var.base_configuration, "vm_size", "Standard_B1s") },
```
rjmateus, line 12: In this case you want to do a lookup on `base_configuration`? I was unable to find it as an output on the base configuration.

backend_modules/azure/host/main.tf
```diff
@@ -0,0 +1,240 @@
...
+    contains(var.roles, "server") ? { instance_type = "Standard_B2s" } : {},
```
rjmateus, line 13: you should be assigning `vm_size` instead of `instance_type`, otherwise, all machine flavors will have size `Standard_B1s` (I don't know if this is the goal, but if it is, then these lines should be removed).

backend_modules/azure/host/main.tf
```diff
@@ -0,0 +1,240 @@
...
+  public_subnet_id                     = var.base_configuration.public_subnet_id
+  private_subnet_id                    = var.base_configuration.private_subnet_id
+  private_additional_subnet_id         = var.base_configuration.private_additional_subnet_id
+  public_security_group_id             = var.base_configuration.public_security_group_id
+  private_security_group_id            = var.base_configuration.private_security_group_id
+  private_additional_security_group_id = var.base_configuration.private_additional_security_group_id
+  resource_group_name = var.base_configuration.resource_group_name
+  resource_name_prefix = "${var.base_configuration["name_prefix"]}${var.name}"
+  
+  public_instance = var.public_instance
```
rjmateus, line 33: should be using a provider specific variable.

backend_modules/azure/host/main.tf
```diff
@@ -0,0 +1,240 @@
...
+  
+  # availability_zone = var.base_configuration["availability_zone"]
```
rjmateus, line 35: If this line is not needed at azure, it can be remove

backend_modules/azure/host/user_data.yaml
```diff
@@ -0,0 +1,214 @@
...
+    repos:
+      - id: opensuse_pool_repo
+        name: opensuse_pool_repo
+        baseurl: http://download.opensuse.org/distribution/leap/15.2/repo/oss/
+        enabled: 1
+        autorefresh: 1
+
+packages: ["salt-minion"]
+
+  # runcmd:
```
rjmateus, line 25: The hacks are not needed for azure or you still need to test it more to be sure? If they are needed, then should be removed.

backend_modules/azure/host/main.tf
```diff
@@ -0,0 +1,240 @@
...
+  vars = {
+    image           = var.image
+    public_instance = local.public_instance
+    mirror_url      = var.base_configuration["mirror"]
+  }
+}
+
+resource "azurerm_public_ip" "suma-pubIP" {
+  count = local.public_instance ? var.quantity: 0
+  name                = "${var.base_configuration["name_prefix"]}${var.name}-pubIP"
```
rjmateus, line 51: should the name include count on it? Or a name collision doesn't matter in this case?

backend_modules/azure/host/main.tf
```diff
@@ -0,0 +1,240 @@
...
+  }
+}
+
+resource "azurerm_linux_virtual_machine" "instance" {
+  count                            = var.quantity
+  name                             = "${local.resource_name_prefix}${var.quantity > 1 ? "-${count.index + 1}" : ""}"
+
+  location                         = local.region
+  resource_group_name              = local.resource_group_name
+  network_interface_ids            = compact(["${azurerm_network_interface.suma-main-nic[count.index].id}","${var.connect_to_additional_network?azurerm_network_interface.suma-additional-nic[count.index].id:""}"])
```
rjmateus, line 96: the check for the additional NIC can be done by calling the length function on the created resource, since the check to create the resource is different from the one defined in here.

```suggestion
  network_interface_ids            = compact(["${azurerm_network_interface.suma-main-nic[count.index].id}","${length(azurerm_network_interface.suma-additional-nic) > 0?azurerm_network_interface.suma-additional-nic[count.index].id:""}"])
```

rjmateus (commented):
Another round of review. Thank you for the contribution

backend_modules/azure/README.md
japelbaum: Yes, Indeed I need to work on the readme. I initially copied the directory to make sure i use the same structure. 
In the latest commit i did i replaced the text with TODO

backend_modules/azure/README_ADVANCED.md
japelbaum: Correct. TODO. Will do it asap. Do you need both for this PR?

backend_modules/null/host/variables.tf
japelbaum: So the reason I have put it in there is that I use this variable when creating a public IP:
resource "azurerm_public_ip" "suma-pubIP" {
  count = local.public_instance ? var.quantity: 0

Now i cannot use the provider_setting variable there as this result in an error saying that the result can only be determined at runtime hence terraform cannot decide how many of these resources to create in the planning phase. Happy for any other suggestion on how to solve this.

backend_modules/azure/network/main.tf
japelbaum: Ah, good catch. I had added this rule to test access to SUMA externally.
We could consider only applying this rule when public instance=1?  Happy to remove this resource from terraform all together

backend_modules/azure/network/main.tf
japelbaum: Made the change and will include in the next commit. Thanks for noticing.

backend_modules/azure/network/variables.tf
japelbaum: name_prefix should be mandatory.

backend_modules/azure/network/versions.tf
japelbaum: Included in next commit

backend_modules/azure/host/versions.tf
japelbaum: included in next commit

salt/server/init.sls
japelbaum: Correct. he idea is not to install suma_pattern there.
Indeed, I added a new product_version suma41 for 2 reasons:
- We can use the latest (supported) suma BYOS images from the public cloud
- It is quicker

salt/default/minimal.sls
japelbaum: same answer :)

backend_modules/azure/base/main.tf
```diff
@@ -0,0 +1,112 @@
...
+}
+
+ module "bastion" {
+  source                        = "../host"
+  quantity                      = local.create_network ? 1 : 0
+  base_configuration            = local.configuration_output
+  image                         = "opensuse152"
+  name                          = "bastion"
+  connect_to_additional_network = true
+  public_instance = true
```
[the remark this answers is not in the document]
japelbaum: Agreed, as mentioned above, the reason i have taken it out of the provider setting as i use this one as a condition for creating a public IP resource.

backend_modules/azure/host/main.tf
japelbaum: Correct, this value is being set at the end in the base module on line 110

backend_modules/azure/host/main.tf
japelbaum: Good point,  this is another leftover from copying over the AWS module and trying to use the same structure. 
Included in next commit

backend_modules/azure/host/main.tf
japelbaum: removed

backend_modules/azure/host/user_data.yaml
japelbaum: So far, it looks like it is not needed. However we may want to check with the public cloud team.

backend_modules/azure/host/main.tf
japelbaum: Should include count in it.

backend_modules/azure/host/main.tf
japelbaum: Maybe we want to wait ill we decide what we do with the public_instance variable as commented before?

japelbaum:
@rjmateus to be honest i am using a trial account, which gives you 200$ of credit to try azure with for 30 days.
With regards to the symbolic links, Github doesnt allow me to commit symbolic links, not sure if there is a workaround?

salt/server/init.sls
moio: :wave: @japelbaum!

We faced the same question when the Public Cloud Team adapted sumaform to test AWS BYOS images, and the result was, it's not really necessary to avoid line 24.

Remember this is a Salt state, so it's supposed to express how configuration should look like, no matter the starting point. Indeed, BYOS images do come with this exact pattern preinstalled, so Salt will happily see that the work is already done and do nothing at all. In my experience, with absolutely negligible performance penalty.

So unless there are other very compelling reasons that may be escaping my attention, you could just leave line 23 alone and happily proceed :wink:

backend_modules/azure/host/main.tf
rjmateus: Sorry but I think I getting blind  :)
I cannot find it on the base module. Can you please past a link?
Does this value need to be set on the base, on can just be hardcoded in here?

rjmateus:
@japelbaum regarding the symbolic links, I just run the following command `ln -sf ../../null/host/variables.tf variables.tf``inside the directory I want to place the symbolic link, in this case `backend_modules/azure/host`.
Then running `git status` on the console show me the file to be added. If you have any problem I can help

backend_modules/azure/README.md
```diff
@@ -0,0 +1 @@
+TODO
```
rjmateus, line 1: https://github.com/uyuni-project/sumaform/pull/814#discussion_r544678032

I think is important to have some information. Wich provider settings are available and what should be passed on each value.

backend_modules/azure/network/main.tf
rjmateus: I think suse manager should not be open to the world. we can then configure a port forward and connect through bastion machine.

backend_modules/azure/network/main.tf
```diff
@@ -0,0 +1,208 @@
...
+resource "azurerm_nat_gateway" "suma-ngw" {
+  count = var.create_network? 1: 0
+
+  name                = "${var.name_prefix}-ngw"
+  resource_group_name = "${azurerm_resource_group.suma-rg.name}"
+  location            = "${azurerm_resource_group.suma-rg.location}"
+}
+
+resource "azurerm_nat_gateway_public_ip_association" "ngw-pubip-association" {
+  nat_gateway_id       = azurerm_nat_gateway.suma-ngw.id
```
rjmateus, line 184:
```suggestion
  count = var.create_network? 1: 0

  nat_gateway_id       = azurerm_nat_gateway.suma-ngw[0].id
```

backend_modules/azure/host/main.tf
rjmateus: I have tried locally with the following code and it look to work:
`public_instance = lookup(var.provider_settings, "public_instance", false)`

Let me know if it works for you

rjmateus (commented):
Another round of reviews 
I have tried the deployment but the suse manager is not deploying correctly. Are you able to deploy it in your environment?

backend_modules/azure/network/main.tf
japelbaum: Use main.tf to set ssh_allowed_ip, similar to AWS

backend_modules/null/host/variables.tf
japelbaum: Add public_instance = lookup(var.provider_settings, "public_instance", false) as local variable

backend_modules/azure/base/main.tf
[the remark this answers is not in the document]
japelbaum: as above added to the local variables.

backend_modules/azure/host/main.tf
japelbaum: you where right, should get this from provider settings

backend_modules/azure/host/user_data.yaml
japelbaum: seems this hack is not needed anymore

rjmateus (commented):
Hey Jona, Some more comments on the way. almost in the finish line :)
My plan is tomorrow test the deployment again.
Thanks for the hard work

.gitignore
```diff
@@ -17,4 +17,6 @@
 **/.*.sw*
 .*.sw*
 
-.idea
\ No newline at end of file
+.idea
+backend_modules/azure/base/variables.tf
```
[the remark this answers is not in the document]
japelbaum: Hey Ricardo

No particular reason. I was initially using windows with visual studio code. And windows has issues with symbolic links.
Will change

main.tf.azure.example
```diff
@@ -0,0 +1,102 @@
+locals {
+  name_prefix = "Tfs"
+  location = "francecentral"
+  key_file            = "Private Key File"
+  public_key_location = "Public Key File"
+}
+
+provider "azurerm" {
+
```
rjmateus, line 9: feature field is mandatory 
```suggestion
features {}
```

rjmateus:
@japelbaum I have tested the deployment of uyuni and that went really good :)
As Soon as the last comments are addressed I think we are good to go.
Thanks for the great work

japelbaum:
> @japelbaum I have tested the deployment of uyuni and that went really good :)
> As Soon as the last comments are addressed I think we are good to go.
> Thanks for the great work

That is awesome news :) Happy to hear. Is there value in writing an integration for gcp?

backend_modules/azure/network/main.tf
```diff
@@ -0,0 +1,197 @@
...
+  name                        = "https"
+  resource_group_name         = "${azurerm_resource_group.suma-rg.name}"
+  network_security_group_name = "${azurerm_network_security_group.public-nsg[0].name}"
+  priority                    = 110
+  direction                   = "Inbound"
+  access                      = "Allow"
+  protocol                    = "Tcp"
+  source_port_range           = "*"
+  destination_port_range      = "443"
+  source_address_prefix       = "*"
```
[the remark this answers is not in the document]
japelbaum: Agreed. I have changed it.

backend_modules/azure/host/user_data.yaml
```diff
@@ -0,0 +1,214 @@
...
+#disable_root: false
+disable_root: false
+ssh_pwauth: true
+chpasswd:
+  expire: false
+  list: |
+    root:linux
+%{ endif }
+
+%{ if image == "opensuse152" || image == "opensuse151" || image == "opensuse150"}
```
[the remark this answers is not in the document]
japelbaum: removed all the os versions that are not used.

PR #818, opened 2020-12-15
As we published the twopence gem in rubygems, we need to install all native extensions dependencies

srbarrios:
## What does this PR change?

This PR solve a problem starting when we published our twopence gem in rubygems, as now it will try to use the latest version which comes from rubygems.

For this, it needs all the native extensions, otherwise it fails on the `bundle install` step.


calancha (approved):
LGTM

Bischoff:
ok

PR #823, opened 2021-01-08
Adding QAM clients necessary for retail

lkotek:
## What does this PR change?

PR adds set of clients necessary for retail in case of QAM.

This **must** be merged before https://github.com/SUSE/susemanager-ci/pull/172.

### DIFF
--- a/modules/controller/main.tf
+++ b/modules/controller/main.tf
@@ -81,6 +81,12 @@ module "controller" {
     ubuntu1804_sshminion = length(var.ubuntu1804_sshminion_configuration["hostnames"]) > 0 ? var.ubuntu1804_sshminion_configuration["hostnames"][0] : null
     ubuntu2004_minion    = length(var.ubuntu2004_minion_configuration["hostnames"]) > 0 ? var.ubuntu2004_minion_configuration["hostnames"][0] : null
     ubuntu2004_sshminion = length(var.ubuntu2004_sshminion_configuration["hostnames"]) > 0 ? var.ubuntu2004_sshminion_configuration["hostnames"][0] : null
+    sle11sp4_buildhost   = length(var.sle11sp4_buildhost_configuration["hostnames"]) > 0 ? var.sle11sp4_buildhost_configuration["hostnames"][0] : null
+    sle11sp3_terminal    = length(var.terminal_configuration["hostnames"]) > 0 ? var.terminal_configuration["hostnames"][0] : null
+    sle12sp4_buildhost   = length(var.sle12sp4_buildhost_configuration["hostnames"]) > 0 ? var.sle12sp4_buildhost_configuration["hostnames"][0] : null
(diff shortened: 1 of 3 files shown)


modules/controller/main.tf
```diff
@@ -81,6 +81,12 @@ module "controller" {
     ubuntu1804_sshminion = length(var.ubuntu1804_sshminion_configuration["hostnames"]) > 0 ? var.ubuntu1804_sshminion_configuration["hostnames"][0] : null
     ubuntu2004_minion    = length(var.ubuntu2004_minion_configuration["hostnames"]) > 0 ? var.ubuntu2004_minion_configuration["hostnames"][0] : null
     ubuntu2004_sshminion = length(var.ubuntu2004_sshminion_configuration["hostnames"]) > 0 ? var.ubuntu2004_sshminion_configuration["hostnames"][0] : null
+    sle11sp4_buildhost   = length(var.sle11sp4_buildhost_configuration["hostnames"]) > 0 ? var.sle11sp4_buildhost_configuration["hostnames"][0] : null
+    sle11sp3_terminal    = length(var.sle11sp3_terminal_configuration["hostnames"]) > 0 ? var.sle11sp3_terminal_configuration["hostnames"][0] : null
```
srbarrios, line 85: This is by design I guess?
So sles11sp3 for terminal and sles11sp4 for buildhost?

modules/controller/main.tf
```diff
@@ -81,6 +81,12 @@ module "controller" {
...
+    sle12sp4_buildhost   = length(var.sle12sp4_buildhost_configuration["hostnames"]) > 0 ? var.sle12sp4_buildhost_configuration["hostnames"][0] : null
+    sle12sp4_terminal    = length(var.sle12sp4_terminal_configuration["hostnames"]) > 0 ? var.sle12sp4_terminal_configuration["hostnames"][0] : null
+    sle15sp2_buildhost   = length(var.sle15sp2_buildhost_configuration["hostnames"]) > 0 ? var.sle15sp2_buildhost_configuration["hostnames"][0] : null
+    sle15sp2_terminal    = length(var.sle15sp2_terminal_configuration["hostnames"]) > 0 ? var.sle15sp2_terminal_configuration["hostnames"][0] : null
```
srbarrios, line 89: I wonder if for the upcoming SUMA4.2 we need to also add sles15sp3 as a possibility here.
(Thinking on build validation)

modules/controller/main.tf
lkotek: Yes, exactly. We must use SLE11SP4 (x86_64) system to build SLE11SP3 (i586) image. (It it not possible to use SLE11SP3 for building - I opened bug report for it about a year ago and it was clarified then.)

modules/controller/main.tf
lkotek: I would say no (but I will ask retail developers to to sure). More general - we still use SLE12SP4 in the retail case for image building, not SP5 for example and the same about terminal - I think retail is more conservative in case of switching to newer versions.

srbarrios (approved):

PR #831, opened 2021-01-15
WIP: Add debian images and variables

ktsamis:
## What does this PR change?

Add debian images and variables, previous was reverted, this needs more fixing before merge


rjmateus:
@ktsamis @srbarrios 
Can you give some insides if you had to do some manual steps to use these images in QAM testsuite?
For example, manually install packages or start some services?
That would be very helpful to test and finish this PR.

ktsamis:
I did not use these for QAM testsuite. We don't have them enabled there yet. I downloaded manually the DVD and installed in another hypervisor and did the whole thing manually. I used the DVD which should be different than the images. I tried to deploy these images with this PR and they failed to get IPs so I think cloud-init is still not set. We need the user_data.yaml section for sure right?

rjmateus:
closes in favor of: https://github.com/uyuni-project/sumaform/pull/856

PR #840, opened 2021-01-25
Removing dead code

Bischoff:
## What does this PR change?

This code removes the creation of a file by sumaform that is not later used by the test suite.

(see issue SUSE/spacewalk#3464)

### DIFF
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -25,11 +25,6 @@ authorized_keys_controller:
     - source: salt://controller/id_rsa.pub
     - makedirs: True
 
-virtualhost:
-  file.managed:
-    - name: /tmp/virtualhostmanager.create.json
-    - source: salt://controller/virtualhostmanager.create.json
-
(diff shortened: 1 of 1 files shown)


Bischoff:
Note: the test suite does not use it, but there is support code for it.

https://github.com/uyuni-project/uyuni/blob/cbaa068a613556350b6c9930edfd781848816d0d/testsuite/features/step_definitions/xmlrpc_common.rb#L285-L338

and

https://github.com/uyuni-project/uyuni/blob/cbaa068a613556350b6c9930edfd781848816d0d/testsuite/features/support/xmlrpc_vhm.rb#L1-L31

Guys, should we also remove this support code from the test suite?

Ping @MalloZup for awareness.

ktsamis (approved):
If we don't use it and there is no plan to use it anymore, let's remove all of them, from sumaform and test suite

calancha (approved):

MalloZup (approved):

PR #841, opened 2021-01-25
remove gpg key hack and change repo name because or order

rjmateus:

## What does this PR change?

Remove gpg key hack for sles 15.
Also changes repo name to force correct repository order.

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -157,15 +157,12 @@ zypper:
       baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-Module-Basesystem/15/x86_64/product
       enabled: 1
       autorefresh: 1
-# HACK: repo metadata is not valid, disabling gpgcheck makes the image able to deploy
-      gpgcheck: 0
-    - id: module_server_applications_pool_repo
+      ## HACK name of server app repo needs to be alphabetically after the base system repo to force GPG key update
+    - id: server_applications_pool_repo
       name: module_server_applications_pool_repo
       baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-Module-Server-Applications/15/x86_64/product
       enabled: 1
       autorefresh: 1
-# HACK: repo metadata is not valid, disabling gpgcheck makes the image able to deploy
-      gpgcheck: 0
 
 packages: ["avahi", "nss-mdns", "qemu-guest-agent"]
 


ktsamis (approved):
Awesome, thank you very much for this :) :100:

mcalmer (approved):

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -157,15 +157,11 @@ zypper:
       baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-Module-Basesystem/15/x86_64/product
       enabled: 1
       autorefresh: 1
-# HACK: repo metadata is not valid, disabling gpgcheck makes the image able to deploy
-      gpgcheck: 0
-    - id: module_server_applications_pool_repo
+    - id: server_applications_pool_repo
```
srbarrios, line 160: Please, let's keep a "HACK: xxx" comment, explaining why we changed that, and what needs to be kept with this name.
Because it still a hack :)

rjmateus: I will add the hack comment, with a note about the maning.
Regarding the change on minion.sls and the order in the repos when applying the `zypper ref` to get the new GPG key it doesn't work in the exact way o mentioned. 
The first interaction of the new machine with zypper and those repos are at cloud-init level, which is completely independent from the salt part, and happens by the configurations on this file.
So at that time we only have 2 repositories at that time and when new repositories are added by salt, GPG key problem were already solved by cloud-init.

srbarrios: I think you don't get my point.
I'm not talking about deployment time, cloud-init or salt states will only run once.
I'm talking of an update of the gpg key on a running machine, with this setup, and suddenly you do a zypper refresh and it fails because the final order of all the repos included during the deployment has this trick related with the GPG only included in a repo that is not taken at first position.

rjmateus: That is not a direct sumaform problem. Sumaform is normally used for short life machines.
If one deploys a long-run machine by hand, will have the exact same problem and needs manual intervention.
That said, this is not the long-term solution, but since we are in the middle of a release we needed a solution that would allow QA to keep working on that.
In order to rename the repositories, we need time to access it and if it will have any impact on machine or sumform and QA execution.
I don't have that time now, and since we need to release a new version of suse manager, is not also the best time for refactorings on sumaform side.

srbarrios: Yep, as I said on my first comment, I'm just remarking an issue that can be done in a future card, when we decide to refactor this. Agree to don't go further at that moment. But is nice to have a issue number in the HACK comment, so we can get more info in the future.

Bischoff: I just saw this now @srbarrios. Yes we need some kind of keyword to remind us this is a hack / a workaround.

I just aligned all the hacks workaround with same tag we use for test suite https://github.com/uyuni-project/sumaform/pull/846

Apparently, Ricardo followed your advice, and there is still such a keyword here.

PR #845, opened 2021-01-26
fix workaround for bsc#1181283

jordimassaguerpla:
## What does this PR change?

The bug only affects SLE15SP3, meaning @includedir is only added in SLE15SP3
By having commented #includedir, I introduced a bug for previous
products based on SLE15SP2 or older, which was that there was no includedir
so sudo was not working on those products.

Let's comment the one that starts with @ so it only affects SP3.


calancha (approved):
Thank you, this must work.

Bischoff:
Thank you @jordimassaguerpla .

PR #848, opened 2021-01-27
Add module base system "moving target" repo

jordimassaguerpla:
## What does this PR change?

add TEST repo that contains postgresql13 and libpq5


jordimassaguerpla:
This is already in https://github.com/uyuni-project/sumaform/blob/master/salt/repos/default.sls#L345 . Sorry for the noise.

PR #850, opened 2021-02-01
Add build_validation variable

ktsamis:
## What does this PR change?

Add build_validation variable to differentiate between CI runs and QAM/Build validations. We need this to fix https://github.com/SUSE/spacewalk/issues/13849

Goes together with: https://github.com/SUSE/susemanager-ci/pull/183

### DIFF
--- a/salt/controller/bashrc
+++ b/salt/controller/bashrc
@@ -68,6 +68,9 @@ export AUTH_REGISTRY_CREDENTIALS="{{ grains.get('auth_registry_username') }}|{{ 
 {% if grains.get('server_http_proxy') | default(false, true) %}export SERVER_HTTP_PROXY="{{ grains.get('server_http_proxy') }}" {% else %}# no server HTTP proxy defined {% endif %}
 {% if grains.get('pxeboot_image') | default(false, true) %}export PXEBOOT_IMAGE="{{ grains.get('pxeboot_image') }}" {% else %}# no PXE boot image defined {% endif %}
 
+# Differentiate from build validation or CI run
+{% if grains.get('build_validation') | default(false, true) %}export BUILD_VALIDATION="true" {% else %}# normal CI {% endif %}
+
 #### Generate certificates for Google Chrome
 if [ ! -f /etc/pki/trust/anchors/$SERVER.cert ]; then
   wget http://$SERVER/pub/RHN-ORG-TRUSTED-SSL-CERT -O /etc/pki/trust/anchors/$SERVER.cert


salt/controller/bashrc
```diff
@@ -68,6 +68,9 @@ export AUTH_REGISTRY_CREDENTIALS="{{ grains.get('auth_registry_username') }}|{{
 {% if grains.get('server_http_proxy') | default(false, true) %}export SERVER_HTTP_PROXY="{{ grains.get('server_http_proxy') }}" {% else %}# no server HTTP proxy defined {% endif %}
 {% if grains.get('pxeboot_image') | default(false, true) %}export PXEBOOT_IMAGE="{{ grains.get('pxeboot_image') }}" {% else %}# no PXE boot image defined {% endif %}
 
+# Differentiate from build validation or CI run
+{% if grains.get('build_validation') | default(false, true) %}export BUILD_VALIDATION="true" {% else %}# normal CI {% endif %}
+
```
srbarrios, line 73: I personally don't like this approach.
I would like to keep sumaform agnostic from the kind of test we run.

Instead, I would handle this variable in Terracumber, as we do when we inject the custom_repositories.json in the controller through Terracumber.

Bischoff: No, no, no :-) . Please.

I want to be able to deploy myself (no jenkins) => sumaform.

And even when deployed by Jenkins, when I test manually, I do not want to need to do `export BUILD_VALIDATION=true` by hand every time i test.

Bischoff (approved):
LGTM. Please use sumaform approach.

srbarrios (changes requested):
Please. Don't merge this.
Discuss it first with @moio and @rjmateus , I think sumaform must be agnostic from the kind of test we will run. Also in my opinion, we are overengineering in that case.
We can handle this variable directly from the pipeline, and if you don't run it using Jenkins, or Terracumber, you can always go to the controller and add the variable in your .bashrc

Proposal: https://github.com/SUSE/susemanager-ci/pull/184

Bischoff (changes requested):
Nice, but incomplete : you need to define the variable, give it a default value, pass it on as a grain...

Bischoff:
> We can handle this variable directly from the pipeline, and if you don't run it using Jenkins, or Terracumber, you can always go to the controller and add the variable in your .bashrc

What if there is no pipeline, and I want to run sumaform as usual?

And no, i don't want to start adding variables by hand to .bashrc.

salt/controller/bashrc
```diff
@@ -68,6 +68,9 @@ export AUTH_REGISTRY_CREDENTIALS="{{ grains.get('auth_registry_username') }}|{{
 {% if grains.get('server_http_proxy') | default(false, true) %}export SERVER_HTTP_PROXY="{{ grains.get('server_http_proxy') }}" {% else %}# no server HTTP proxy defined {% endif %}
 {% if grains.get('pxeboot_image') | default(false, true) %}export PXEBOOT_IMAGE="{{ grains.get('pxeboot_image') }}" {% else %}# no PXE boot image defined {% endif %}
 
+# Differentiate from build validation or CI run
+{% if grains.get('build_validation') | default(false, true) %}export BUILD_VALIDATION="true" {% else %}# normal CI {% endif %}
```
rjmateus, line 72: Who is responsible for writing the grains `build_validation`. I don't see it in this PR.

Bischoff: yes, the PR is incomplete, see my review comment

ktsamis: This would be in the main.tf we use as a custom grain https://github.com/SUSE/susemanager-ci/pull/183 we can of course set it in the controller always via sumaform, I can modify this

ktsamis:
After discussions we will remove this completely. not needed anymore

PR #854, opened 2021-02-08
Update sles15sp3o image

calancha:
## What does this PR change?

The image at SLE-15-SP3-Full-TEST is updated quite often.
Instead, use the one at SLE-15-SP3-JeOS-Snapshot9.
- backend_modules/libvirt/base/main.tf (sles15sp3o):
Update image URL.


Bischoff (approved):
lgtm

PR #857, opened 2021-02-12
Change the 4.2 beta url to match mirrors without changing for default

ktsamis:
## What does this PR change?

Change the 4.2 beta url to match mirrors without changing for default
Because for download.suse.de the url doesn't change, but our mirror script put the folder of the product with the /ibs/ as well so we need that in the end result url.

### DIFF
--- a/salt/repos/proxy.sls
+++ b/salt/repos/proxy.sls
@@ -76,7 +76,7 @@ proxy_product_update_repo:
 {% if 'beta' in grains['product_version'] %}
 proxy_module_pool_repo:
   pkgrepo.managed:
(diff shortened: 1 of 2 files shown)


Bischoff (approved):
lgtm

srbarrios:
This doesn't look good to me @ktsamis 
This must be wrong for sure.
Check all the other repositories handled in Sumaform, and how we handle the mirror for all of them.

srbarrios:
This doesn't look good to me @ktsamis 
This must be wrong for sure.
Check all the other repositories handled in Sumaform, and how we handle the mirror for all of them.

PR #858, opened 2021-02-12
Beta urls

ktsamis:
## What does this PR change?

Correct res and sle11 beta urls for mirror from BETA to Beta. 
Also fix detection of release in RES 7 case for tools_pool_repo

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -69,7 +69,7 @@ tools_pool_repo:
   pkgrepo.managed:
     {% if grains.get('mirror') %}
     {% if 'beta' in grains.get('product_version') | default('', true) %}
-    - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/SLES11-SP4-SUSE-Manager-Tools-BETA/sle-11-x86_64/
+    - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/SLES11-SP4-SUSE-Manager-Tools-Beta/sle-11-x86_64/
     {% else %}
     - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/SLES11-SP4-SUSE-Manager-Tools/sle-11-x86_64/
     {% endif %}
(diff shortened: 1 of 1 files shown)


ktsamis:
@moio I put you as reviewer because I'm not sure about the commit about detection 937c1a7 it's been there for a while and I'm surprised we didn't see it before. Am I wrong in fixing it? Maybe it is correct?

salt/repos/default.sls
```diff
@@ -415,9 +415,9 @@ tools_pool_repo:
     {% else %}
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/RES/{{ release }}-CLIENT-TOOLS/x86_64/product/
     {% endif %}
-    {% elif grains.get('mirror') %}
+    {% elif release == 7 %}
```
juliogonzalez, line 418: Well... don't really understand why this change is here, but if you tested it, I guess it's OK

salt/repos/default.sls
```diff
@@ -436,7 +436,7 @@ tools_update_repo:
   pkgrepo.managed:
     - humanname: tools_update_repo
     {% if 'beta' in grains.get('product_version') | default('', true) %}
-    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/RES/{{ release }}-CLIENT-TOOLS-BETA/x86_64/update/
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/RES/{{ release }}-CLIENT-TOOLS-Beta/x86_64/update/
```
juliogonzalez, line 439: This is NOT Ok.

Check https://download.suse.de/ibs/SUSE/Updates/RES/ and you will see all beta client tools are `N-CLIENT-TOOLS-BETA` Could it be you have problems at your mirror?

juliogonzalez (changes requested):
Are you sure the changes from `BETA`  to `Beta` are needed in all cases? At least the last change is clearly wrong.

juliogonzalez (approved):
LGTM now.

srbarrios:
@juliogonzalez @ktsamis I think I already commented about these lines in the past. I think we should homogenize the way to handle the mirror in our repos. Instead of using this if statement that is hard to read, I would use the format we are using in other states by replace the domain directly inside the repo url.

Issue #859, opened 2021-02-15
Different switches for SUMA/Uyuni disk mirror implementation and the mirror used during deployment phase


srbarrios:
Currently, if we want to use a Minima mirror only to configure SUSE Manager to download the content from a mirror, so rhn.conf, we can't; we are forced to also mirror everything that sumaform needs during the automatic configuration phase, so during salt states in Sumaform, where tons of temporary repositories are set up.

I would like to have a different variable to use only for disk mirror feature on SUMA, when syncing products from SCC.

So, this salt states:
https://github.com/uyuni-project/sumaform/blob/master/salt/server/rhn.sls#L41-L69

srbarrios:
@rjmateus @moio if you both agreed on this need, I will find time to implement it.

moio:
What is the use case you are trying to achieve? Deploy Servers without a mirror but only consume content from a mirror?

If that is the case then yes, I am not against it.

srbarrios:
Exactly, so I want to avoid mainting in the mirror for QA Maintenance tests the repositories used by Sumaform, and only maintain the repositories from SCC products, to avoid consume too much network traffic in our QAM test suite runs, when SUMA server sync all the products supported by our test suite.

moio:
Feel free to go ahead

PR #861, opened 2021-02-18
New parameter: server_mounted_mirror (mirror only for product sync.)

srbarrios:
## What does this PR change?

Implements https://github.com/uyuni-project/sumaform/issues/859

Currently, if we want to use a Minima mirror only to configure SUSE Manager to download the content from a mirror, so rhn.conf, we can't; we are forced to also mirror everything that sumaform needs during the automatic configuration phase, so during salt states in Sumaform, where tons of temporary repositories are set up.

In this PR we use a different variable to use only for the disk mirror feature on SUMA, when syncing products from SCC.

### DIFF
--- a/.ci-travis/salt-server-validation
+++ b/.ci-travis/salt-server-validation
@@ -18,6 +18,7 @@ product_version: head
 cc_username: UC7
 cc_password: passme
 mirror: null
+server_mounted_mirror: null
 iss_master: null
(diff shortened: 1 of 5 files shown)


srbarrios:
Tested locally and works ;)

rjmateus (approved):
LGTM

PR #868, opened 2021-03-18
Add 3 needed packages in debian deployment for bootstrap

ktsamis:
## What does this PR change?

Adds 3 needed packages in debian deployment for bootstrap. `apt-transport-https", "python-apt", "python3-apt"` are needed, see https://bugzilla.suse.com/show_bug.cgi?id=1183649#c2 . Using this I deployed 2 debian hosts and it worked. So there is no concern about the order of the package installation (cloud-init vs salt)


Bischoff (approved):
lgtm (but why debian 10 then debian 9 ? my inner Sheldon is frustrated...)

PR #870, opened 2021-03-30
Do not always install proxy pattern

Bischoff:
## What does this PR change?

In the cucumber test suites, we already install retail pattern manually. We would like to do the same for the proxy pattern.

Benefits:
 - BV test suite: do not collide with salt cleanup scenario (it was removing `/etc/salt/broker` configuration file)
 - closer mimicking of what a customer does in the test suite
 - independance of sumaform

To achieve that, this PR introduces a new `install_proxy_pattern` configuration variable that defaults to `true` for compatibility.

See also uyuni-project/uyuni#3487 and SUSE/susemanager-ci#219


rjmateus (approved):
lgtm

srbarrios (approved):
LGTM.
Just a remark Eric, keep in mind that we are not using cucumber_testsuite module for BV terraform files. So if the default value is true, keep in mind to set this additional parameter in our tf files.

Bischoff:
> LGTM.
> Just a remark Eric, keep in mind that we are not using cucumber_testsuite module for BV terraform files. So if the default value is true, keep in mind to set this additional parameter in our tf files.

ah, right, will do that

PR #878, opened 2021-05-03
disable forwarding of registrations by default in sumaform

mcalmer:
## What does this PR change?

Disable forwarding registrations to SCC by default. This is to prevent dozen of dummy registration proxies to appear in SCC.

### DIFF
--- a/salt/server/rhn.sls
+++ b/salt/server/rhn.sls
@@ -91,6 +91,17 @@ rhn_conf_prometheus:
 
 {% endif %}
 
+{% if not grains.get('forward_registration') | default(false, true) %}
+
+rhn_conf_forward_reg:
+  file.append:
+    - name: /etc/rhn/rhn.conf
+    - text: server.susemanager.forward_registration = 0
+    - require:
+      - sls: server
+
+{% endif %}
+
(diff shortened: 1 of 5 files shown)


salt/server/rhn.sls
```diff
@@ -91,6 +91,17 @@ rhn_conf_prometheus:
 
 {% endif %}
 
+{% if not grains.get('forward_registration') | default(true, true) %}
```
rjmateus, line 94: I think default should be false, instead of true.

rjmateus (commented):
Just one thing to clarify.
But I have another question. Should we have a parameter for this or should we always disable it on machines created by sumaform?
Do we have a use case where we may want to register in SCC machines created by sumaform?

salt/server/rhn.sls
mcalmer: > But I have another question. Should we have a parameter for this or should we always disable it on machines created by sumaform?
> Do we have a use case where we may want to register in SCC machines created by sumaform?

Not really, but as in Uyuni the default is "true" I do not know if we should force disable it without the option to overwrite it and turn it on. Could be usefull if developer want to setup an environment where they want to develop/test this feature.

rjmateus (approved):

PR #882, opened 2021-05-07
Add opensuse 15.2 image for PR testing

jordimassaguerpla:
## What does this PR change?

This images includes uyuni server dependencies so that, when used for
testing, it takes 3 minutes to install, instead of 15.

With the plain JeOS, when installing uyuni server pattern, we need to
install more than 900 packages. With this new image, only 115 are
needed.

This goes with https://build.opensuse.org/request/show/891292

### DIFF
--- a/backend_modules/libvirt/base/main.tf
+++ b/backend_modules/libvirt/base/main.tf
@@ -10,7 +10,7 @@ locals {
     opensuse150o = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/distribution/leap/15.0/jeos/openSUSE-Leap-15.0-JeOS.x86_64-15.0.1-OpenStack-Cloud-Snapshot21.14.qcow2"
     opensuse151  = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse151.x86_64.qcow2"
     opensuse151o = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/distribution/leap/15.1/jeos/openSUSE-Leap-15.1-JeOS.x86_64-15.1.0-OpenStack-Cloud-Current.qcow2"
-    opensuse152  = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse152.x86_64.qcow2"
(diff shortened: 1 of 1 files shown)


rjmateus (approved):

moio:
@jordimassaguerpla thanks for this contribution.

I am a bit surprised about learning the difference in provisioning time is that big (see https://build.opensuse.org/request/show/891294), but I completely believe you with the measurements and trust you this is a good investment to trim some fat out of the process.

What I would still ask you is to name this image differently, for example with a suffix, in order to make it very explicit this is an special exception - sumaform images are by design not provided by this project, and even when that is the case they are by design minimal. The design trade-off we took was to maximize the flexibility at the (understood) cost of efficiency, and I still believe it's a good trade-off for most case. Now I completely understand it falls short in this particular case, and I wouldn't suggest to be too strict, but we should make the difference visible and explicit to the extent possible (as it violates the "principle of least surprise").

For reference, please see sumaform's design document in this regard: https://github.com/uyuni-project/sumaform/blob/master/DESIGN.md#base-disk-images

PS. I would feel even happier with a generic way to define an arbitrary image URL as part of one's main.tf - for any case in which the ones provided do not fit. I can definitely see similar cases in future. Nevertheless that's a nice-to-have from my perspective and not a requirement at this time.

Thanks!

jordimassaguerpla:
> @jordimassaguerpla thanks for this contribution.
> 
> I am a bit surprised about learning the difference in provisioning time is that big (see https://build.opensuse.org/request/show/891294), but I completely believe you with the measurements and trust you this is a good investment to trim some fat out of the process.
> 
> What I would still ask you is to name this image differently, for example with a suffix, in order to make it very explicit this is an special exception - sumaform images are by design not provided by this project, and even when that is the case they are by design minimal. The design trade-off we took was to maximize the flexibility at the (understood) cost of efficiency, and I still believe it's a good trade-off for most case. Now I completely understand it falls short in this particular case, and I wouldn't suggest to be too strict, but we should make the difference visible and explicit to the extent possible (as it violates the "principle of least surprise").
> 
> For reference, please see sumaform's design document in this regard: https://github.com/uyuni-project/sumaform/blob/master/DESIGN.md#base-disk-images
> 
> PS. I would feel even happier with a generic way to define an arbitrary image URL as part of one's main.tf - for any case in which the ones provided do not fit. I can definitely see similar cases in future. Nevertheless that's a nice-to-have from my perspective and not a requirement at this time.
> 
> Thanks!

New SR https://build.opensuse.org/request/show/892081 . I hope this addresses  your concerns.

thanks in advance

PR #886, opened 2021-05-19
Add ltss repos for 11sp4, 12sp4, 15, 15sp1

ktsamis:
## What does this PR change?

Add ltss repos for 11sp4, 12sp4, 15, 15sp1 and in advance 15sp2 and 15sp3. I didn't add all the 12sp* because we need to add them in the mirrors and there is no point since we won't be using them.

Adding the LTSS repos would fix openscap failures for older OS that only have the scap-security-guide in LTSS repos and not any others.

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -57,6 +57,14 @@ os_update_repo:
     - baseurl: http://euklid.nue.suse.com/mirror/SuSE/build-ncc.suse.de/SUSE/Updates/SLE-SERVER/11-SP4/x86_64/update/
     {% endif %}
 
+os_ltss_repo:
+  pkgrepo.managed:
+    {% if grains.get('mirror') %}
+    - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/SLES11-SP4-Updates/sle-11-x86_64/
+    {% else %}
+    - baseurl: http://euklid.nue.suse.com/mirror/SuSE/build-ncc.suse.de/SUSE/Updates/SLE-SERVER/11-SP4-LTSS/x86_64/update/
+    {% endif %}
+
 {% if grains.get('use_os_unreleased_updates') | default(False, true) %}
 test_update_repo:
   pkgrepo.managed:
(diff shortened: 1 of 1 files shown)


salt/repos/default.sls
```diff
@@ -197,6 +209,8 @@ os_update_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-SERVER/12-SP5/x86_64/update/
 
+# TODO: Add os_ltss_repo when it goes into LTSS
```
juliogonzalez, line 212: Ok, so in the end I see you decided not to add already the repos for SP2/SP3 even if they are empty?

Not complaining, just checking, as you told me otherwise :-)

salt/repos/default.sls
```diff
@@ -353,8 +382,13 @@ os_pool_repo:
...
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Basesystem/15-SP3/x86_64/update/
-{% endif %} {# '15.3' == grains['osrelease'] #}
 
+# Already made in advance but empty now:
+os_ltss_repo:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Product-SLES/15-SP3-LTSS/x86_64/update/
+
+{% endif %} {# '15.3' == grains['osrelease'] #}
```
juliogonzalez, line 391: Funny this line get's added and removed. Did you add trailing whitespace or something?

juliogonzalez (approved):
Looks good. Left some comments.

rjmateus (approved):
lgtm

salt/repos/default.sls
ktsamis: I didn't add the ones for 12sp3 and 12sp2 because then we need to add them in the mirrors and we don't use those in any testsuite.

salt/repos/default.sls
ktsamis: No just moved it a bit lower

ktsamis: maybe I removed automatically a whitespace but that's not highlighted so I think it's just the empty line before

salt/repos/default.sls
ktsamis: 15sp2 and 15sp3 were added lower

Issue #887, opened 2021-05-21
SLE11 SP4 image is broken


Bischoff:
In SLE11 SP4, the version of kernel (from initrd) and modules do not match:
```
ebi5-minion:~ # grep PRETTY /etc/os-release 
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"

ebi5-minion:~ # ls /lib/modules
3.0.101-108.126-default

ebi5-minion:~ # uname -a
Linux ebi5-minion 3.0.101-63-default #1 SMP Tue Jun 23 16:02:31 UTC 2015 (4b89d0c) x86_64 x86_64 x86_64 GNU/Linux

ebi5-minion:~ # depmod
WARNING: Couldn't open directory /lib/modules/3.0.101-63-default: No such file or directory
FATAL: Could not open /lib/modules/3.0.101-63-default/modules.dep.temp for writing: No such file or directory
```

This, under some circumstances, can result in DHCP not working after a reboot, or serial console not work, or other problems.

Root cause is the content of `/boot/grub/menu.lst`:
```
title sles11sp4_[_VMX_]
 root (hd0,0)
 kernel /boot/linux.vmx vga=0x314 splash=silent console=tty0 console=ttyS0,115200 root=/dev/vda1 disk=/dev/vda nomodeset elevator=noop showopts
 initrd /boot/initrd.vmx
 ```
If I use normal kernel and initrd instead of `linux.vmx` and `initrd.vmx`, there is no problem.

I will write a workaround for sumaform, but it would be better to fix the issue at image generation time.

Ping @moio for awareness.

moio:
Hmm, interesting.

There is something I am missing, why do you mention packer?

Bischoff:
Because I am lost in your images creation procedures... :-P

Bischoff:
I reverted my patch because apparently i missed something about the initial state about deployment, but we definitely need to fix this. I will look more calmly after GMC.

Bischoff:
Still not working, closing in favour of #936

PR #888, opened 2021-05-21
Work around initrd issue in sle11sp4

Bischoff:
## What does this PR change?

This PR makes sure SLES 11 SP4 uses same version for kernel and kernel modules after a reboot.

A better solution would be to generate correct images. For more details, see #887.

### DIFF
--- a/salt/default/grub.sls
+++ b/salt/default/grub.sls
@@ -1,0 +1,20 @@
+# HACK
+# This problem should be fixed when we build the image
+#   https://build.suse.de/package/show/Devel:Galaxy:Terraform:Images/sles11sp4
+{% if grains['osfullname'] == 'SLES' %}
+
+{% if grains['osrelease'] == '11.4' %}
+grub_hack:
+  file.replace:
+    - name: /boot/grub/menu.lst
+    - pattern: ^(title.*)$
+    - count: 1
+    - repl: |
+       title sles11sp4_fixed
+         root (hd0,0)
+         kernel /boot/vmlinuz-3.0.101-108.126-default vga=0x314 splash=silent console=tty0 console=ttyS0,115200 root=/dev/vda1 disk=/dev/vda nomodeset elevator=noop showopts
+         initrd /boot/initrd-3.0.101-108.126-default
+       \1
+{% endif %}
+
+{% endif %}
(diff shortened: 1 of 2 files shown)


moio:
Are we sure the grub menu is incorrect right after the image has been created?

Sources: https://build.suse.de/package/view_file/Devel:Galaxy:Terraform:Images/sles11sp4/sles11sp4.kiwi?expand=1

IOW, I wonder whether the image is actually wrong or if we are updating the kernel for whatever reason, and it's the update that's gone wrong.

srbarrios:
If we need to hack something in the current sles11sp4 custom image, I would do it at cloud-init stage.
So here: https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/host/user_data.yaml

Bischoff:
> Are we sure the grub menu is incorrect right after the image has been created?

Yes, I loop-mounted the freshly downloaded qcow2, and the problem is already there.

> Sources: https://build.suse.de/package/view_file/Devel:Galaxy:Terraform:Images/sles11sp4/sles11sp4.kiwi?expand=1

```        <type image="vmx" filesystem="ext3" boot="vmxboot/suse-SLES11" format="qcow2" bootloader="grub"  kernelcmdline="console=tty0 console=ttyS0,115200 root=/dev/vda1 disk=/dev/vda nomodeset elevator=noop" boottimeout="0">
```

It looks like the problem stems from this `vmx` thingie.

> IOW, I wonder whether the image is actually wrong or if we are updating the kernel for whatever reason, and it's the update that's gone wrong.

Nope, the image is already wrong.

Bischoff:
> If we need to hack something in the current sles11sp4 custom image, I would do it as earlier as possible, at the cloud-init stage.
> So here: https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/host/user_data.yaml

Why not, but an even better solution would be to directly fix image creation.

The problem is, I'd like to test GMC now with a SLES 11 SP4 that survives a reboot. So I would suggest to apply this PR and remove it once we have something better.

Bischoff:
I'm trying to modify directly the image

Bischoff:
Reverted, I get a different image in CI than in my local tests...

srbarrios:
Is our mirror updated?

Bischoff:
> Is our mirror updated?

Good question.

I will give it a go again after next BV deployment. It could have been the mirror, it could be sumaform does differently when using the testsuite module or not, or it could have been me missing some point.

Issue #834, opened 2021-01-19
SLES12SP4 python-cryptography dependency issue


srbarrios:
Related to https://github.com/SUSE/spacewalk/issues/13314

By default in our official image, from sumaform, we have pre-installed:
```suma-qam-41-min-sles12sp4:~ # rpm -qa| grep cryptography
python-cryptography-1.3.1-7.10.1.x86_64
```

That results on this issue, at boostrap:
```* registering
From cffi callback <function ssl_verify_callback at 0x7f3797639488>:
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/OpenSSL/SSL.py", line 214, in wrapper
    _lib.X509_up_ref(x509)
AttributeError: 'module' object has no attribute 'X509_up_ref'
An error has occurred:
rhn-plugin: The SSL certificate failed verification.
See /var/log/up2date for more information
```

But in our repos we have the proper version 2.1.4:
zypper actually display the correct version is also available 
```S | Name                 | Type       | Version      | Arch   | Repository    
--+----------------------+------------+--------------+--------+---------------
v | python-cryptography  | package    | 2.1.4-7.28.2 | x86_64 | os_update_repo
v | python-cryptography  | package    | 1.7.2-7.24.1 | x86_64 | os_update_repo
v | python-cryptography  | package    | 1.3.1-7.13.4 | x86_64 | os_update_repo
i | python-cryptography  | package    | 1.3.1-7.10.1 | x86_64 | os_pool_repo  
````

We need to figure out which is the package that depends on python-cryptography, to evaluate a proper solution.
For now, we did a [workaround](https://github.com/uyuni-project/sumaform/pull/832) to upgrade this package manually before we start the bootstrap.

- We need to see if we want to document this manual step in our official documentation
- We need to evalute the possibility to upgrade it on the boostrap stage, instead of manually.

**Comment by Bischoff** [comment]
The workaround does not work anymore:

```
[2021-05-25T17:45:11.354Z] module.sles12sp4-client.module.client.module.host.null_resource.provisioning[0] (remote-exec):     31m Comment: The following packages failed to update: python-cryptography`
￼```￼

Furthermore, it is not okay to hide a product problem in sumaform. I prefer it to fail in the test suite.

I'll do a PR to remove the hack.

Bischoff:
Not a sumaform issue, closing.

PR #889, opened 2021-05-25
Remove hack for sumaform issue #834

Bischoff:
## What does this PR change?

See #834.

1) we should not hide product bugs with sumaform hacks - I prefer it to fail in the test suite, we create a bug, developers fix the bug, etc.

2) the hack did not work anymore anyway :smile_cat: .

### DIFF
--- a/salt/client/testsuite.sls
+++ b/salt/client/testsuite.sls
@@ -23,15 +23,6 @@ suse_client_cucumber_requisites:
     - require:
       - sls: repos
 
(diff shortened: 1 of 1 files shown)


srbarrios:
@Bischoff please consider this comment https://github.com/SUSE/spacewalk/issues/13314#issuecomment-760884734

So you understand the reasons for that hack. As I shared and re-shared the issue in galaxy-devel and MC only agreed with me in doing this hack, but he did not decide next step.

moio:
Thanks

PR #896, opened 2021-06-03
Fix opensuse153's avahi

moio:
I tried to install Uyuni Master on libvirt with a mirror and one of the problems is that opensuse153 does not come with avahi preinstalled.

This patch fixes it.


moio:
@juliogonzalez any objections from your side?

juliogonzalez (approved):
No objections.

PR #897, opened 2021-06-03
Drop openSUSE 15.0 and 15.1

moio:
This drops the unsupported distro, to lighten up mirrors a bit.

Builds on top of https://github.com/uyuni-project/sumaform/pull/896 (is best merged after that one)

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -104,13 +104,6 @@ yum_repos:
 packages: ["salt-minion", "avahi", "nss-mdns", "qemu-guest-agent"]
 %{ endif }
 
-%{ if image == "opensuse151o" }
-packages: ["qemu-guest-agent"]
-
-runcmd:
-  - zypper removerepo --all
-%{ endif }
-
 %{ if image == "opensuse152o" }
 packages: ["qemu-guest-agent"]
 
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -104,13 +104,6 @@ yum_repos:
 packages: ["salt-minion", "avahi", "nss-mdns", "qemu-guest-agent"]
 %{ endif }
 
-%{ if image == "opensuse150o" }
-packages: ["salt-minion", "avahi", "nss-mdns", "qemu-guest-agent"]
-
-runcmd:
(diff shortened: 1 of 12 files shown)


juliogonzalez:
@moio maybe you should drop 15.1 as well? It's dead since January 31st already.

rjmateus (approved):
LGTM
One question. Should the following repos also be changed to not point to Leap 15.0?
https://github.com/uyuni-project/sumaform/blob/master/salt/repos/default.sls#L28-L35

juliogonzalez:
> LGTM
> One question. Should the following repos also be changed to not point to Leap 15.0?
> https://github.com/uyuni-project/sumaform/blob/master/salt/repos/default.sls#L28-L35

Nope! As for SUSE Manager (where SLE15 clients are always built on top off SLE15SP0), for Uyuni the openSUSE Leap 15.X client tools are built on top of Leap 15.0, so that URL never changes. I should have probable named it "15" instead of "15.0" but didn't remember when I created the structure, and now everyone is using it.

moio:
> @moio maybe you should drop 15.1 as well? It's dead since January 31st already.

Done

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -104,22 +104,15 @@ yum_repos:
...
+%{ if image == "opensuse152o" }
 packages: ["qemu-guest-agent"]
 
 runcmd:
   - zypper removerepo --all
 %{ endif }
 
-%{ if image == "opensuse152o" }
-packages: ["qemu-guest-agent"]
+%{ if image == "opensuse153o" }
```
juliogonzalez, line 114: Not sure if we still use 15.2 somewhere. Maybe this should be an `or`?

juliogonzalez (approved):
Minor comment. Other than that, LGTM.

backend_modules/libvirt/host/user_data.yaml
moio: Not sure I understand - line 107 covers 15.2 and it gets the same package as before: `qemu-guest-agent`. So no change expected for 15.2.

15.3 needs different packages (`avahi` and `nss-mdns`).

Maybe I am missing something?

juliogonzalez: Ah, right. I got confused by how the diff is showing the stuff.

LGTM.

PR #899, opened 2021-06-10
Include support for sle12sp5 and update it as default for sles12 clie…

srbarrios:
## What does this PR change?

Include support for sle12sp5 and update it as default for sles12 clients/minions


Bischoff (approved):
I did not see errors.

It was not needed to make it the default, but why not.

ktsamis (approved):

PR #903, opened 2021-06-10
Fix 15sp2 and sp3 LTSS repos when coming from mirror, as we still are…

srbarrios:
## What does this PR change?

Fix 15sp2 and sp3 LTSS repos when coming from mirror.
As we still are not able to sync them from scc, we need to sync them from the same ibs repo via http section in minima.


Bischoff (approved):
yeah this thingie is never ending

PR #904, opened 2021-06-21
Update sles15sp3o image to GM

calancha:
## What does this PR change?

Update sles15sp3o image to GM


srbarrios (approved):

srbarrios:
It should not affect Head or Uyuni as we don't use mirror, but remember to update the mirror for others

PR #914, opened 2021-07-06
Temporary replace the npm of cucumber-html-reporter

srbarrios:
## What does this PR change?

In order to support the latest version of cucumber-ruby gem, we need to fix cucumber-html-reporter.
As the new output from the internal reporter is slightly different, so the hooks we have in place that contains the screenshots and logs of a failed scenario are stored differently, so we need to parse them differently.

As the external reporter has not many updates, it will probably take long time to have a new release.
So, as temporary solution, I would like to use my personal fork that includes the fix.

### DIFF
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -81,9 +81,16 @@ install_npm:
     - name: npm8
 
 # https://github.com/gkushang/cucumber-html-reporter
-install_cucumber_html_reporter_via_npm:
+#install_cucumber_html_reporter_via_npm:
+#  cmd.run:
+#    - name: npm install cucumber-html-reporter@5.2.0 --save-dev
+#    - require:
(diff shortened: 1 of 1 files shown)


ktsamis (approved):

Bischoff (approved):

ncounter (approved):
This patch LGTM, and the fix itself from your fork looks good :+1:

Generally speaking, keep in mind to not update or change your fork as we will fetch it from here dynamically. As an alternative, to be *external-fork-repo independent* I'd even be fine with storing the *patched* version of the `lib/reporter.js` file here in `sumaform` and apply an additional salt state to replace the file after installing the upstream version. The result is the same, it's only a matter choice how to do it.

But as said, I'm fine with your current solution :+1:

PR #915, opened 2021-07-07
Support mirror in testsuite wrapper

Bischoff:
## What does this PR change?

So far, we could only use a mirror with old "modules" syntax.

This PR makes it possible to use the new "cucumber_testsuite" syntax.

It also switches the base image of mirrors from opensuse 15.1 to opensuse 15.2 "official".

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -92,6 +92,16 @@ locals {
   minimal_configuration = { hostname = contains(local.hosts, "proxy") ? local.proxy_full_name : local.server_full_name }
 }
 
+module "mirror" {
+  source = "../mirror"
+
+  quantity = contains(local.hosts, "mirror") ? 1 : 0
+  base_configuration = module.base.configuration
+
+  provider_settings        = lookup(local.provider_settings_by_host, "mirror", {})
+  volume_provider_settings = lookup(var.host_settings["mirror"], "volume_provider_settings", {})
+}
+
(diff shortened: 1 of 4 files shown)


Bischoff:
Only base tests carried out so far. Extended testing will only be done in conjunction with new "get rid of mirror.sh" PR.

modules/cucumber_testsuite/main.tf
```diff
@@ -92,6 +92,16 @@ locals {
   minimal_configuration = { hostname = contains(local.hosts, "proxy") ? local.proxy_full_name : local.server_full_name }
 }
 
+module "mirror" {
+  source = "../mirror"
+
+  quantity = contains(local.hosts, "mirror") ? 1 : 0
```
srbarrios, line 98: Could you explain where and how we use this variable please?

srbarrios (approved):

modules/cucumber_testsuite/main.tf
Bischoff: @srbarrios you can use it everywhere manually, for example if you want 10 SSH minions for whatever reason.

In this case it just means: create it if `cucumber_testsuite` module contains a
```
mirror = {
}
```
declaration, and do not create it otherwise.

Bischoff:
This PR makes no sense: when you deploy the test suite, you have no guarantee that the mirror is already fully synched.

You have to use two separate `main.tf` files, one for the mirror, one for the test suite.

Closing.

PR #917, opened 2021-07-08
Simplify creation of mirrors

Bischoff:
## What does this PR change?

`mirror.sh` is a wrapper script that is, in my opinion, horrible. It mixes a lot of different things under one roof, and then adds the same number of command line options to access those things separately again.

To make things even worse, in our mirrors, configuration files are sparse here and there, some of them being in dot-directories, i.e. hidden. Finally, we call it from a crontab attached to root user, a method which I would qualify as "hiding" too.

This PR removes this script, and then does the same that I did for minima-mirror-bv.mgr.prv.suse.net and minima-mirror-bv2.mgr.prv.suse.net:

* have one script in `/usr/local/bin` for each task we want to perform
* have one configuration file in `/etc` for each of the functions of the mirror
* have one log file in `/var/log` for each of these functions too
* have one symlink in `/etc/cron.xxx` for each task that has to be run periodically

This is both simple and obeying standard paths and practices.

This PR also upgrades the base image for mirrors to openSUSE 15.2 "official".


salt/mirror/cron.daily/apt-mirror.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
```
moio, line 1: Is there a particular reason why you chose cron.daily bash scripts instead of systemd units and timers?

Bischoff: yes, the fact that I know it better - we can still change that in a separate PR later on

juliogonzalez (approved):
Looks good, but I have a question: what if the user wants to have an extra execution that does not happen only once per day, such as we do for the SUSE Manager mirrors?

Bischoff:
> Looks good, but I have a question: what if the user wants to have an extra execution that does not happen only once per day, such as we do for the SUSE Manager mirrors?

He will just drop a script in `cron.hourly` for example.

The main crontab is still available in case of even more advanced need (not the root user's crontab, the system's crontab!).

juliogonzalez:
> > Looks good, but I have a question: what if the user wants to have an extra execution that does not happen only once per day, such as we do for the SUSE Manager mirrors?
> 
> He will just drop a script in `/etc/cron.hourly` for example.
> 
> The main crontab is still available in case of even more advanced need (not the root user's crontab, the system's crontab!).

Maybe I am missing something, but shouldn't then we maybe have all the scripts you did somewhere else, and link them or call them as needed (instead of placing them directly at the cron directories?

Bischoff:
> Maybe I am missing something, but shouldn't then we maybe have all the scripts you did somewhere else, and link them or call them as needed (instead of placing them directly at the cron directories?

I'm afraid we're starting to overengineer. It is common practice to drop scripts directly in crontab directories.

But if you insist, I'll create symlinks. I don't mind either.

moio:
> Maybe I am missing something, but shouldn't then we maybe have all the scripts you did somewhere else, and link them or call them as needed (instead of placing them directly at the cron directories?

At that point I'd really prefer systemd units and timers. Easy to start one-off, easy to override, take care of logs by themselves and (arguably from my perspective) easier to read.

Bischoff:
> > Maybe I am missing something, but shouldn't then we maybe have all the scripts you did somewhere else, and link them or call them as needed (instead of placing them directly at the cron directories?
> 
> At that point I'd really prefer systemd units and timers. Easy to start one-off, easy to override, take care of logs by themselves and (arguably from my perspective) easier to read.

Will do, but not now.

Bischoff:
> > > Looks good, but I have a question: what if the user wants to have an extra execution that does not happen only once per day, such as we do for the SUSE Manager mirrors?
> > 
> > 
> > He will just drop a script in `/etc/cron.hourly` for example.
> > The main crontab is still available in case of even more advanced need (not the root user's crontab, the system's crontab!).
> 
> Maybe I am missing something, but shouldn't then we maybe have all the scripts you did somewhere else, and link them or call them as needed (instead of placing them directly at the cron directories?

Done.

PR #919, opened 2021-07-13
Add scc support for aws deployment

maximenoel8:
## What does this PR change?

The use of internal repositories for public imply the use of mirror. The use of this mirror make the deployment very time consuming.

The idea of this PR is  to have the possibility to use SCC repositories when deploying on AWS. So when we want to test AWS deployment, we will only need to mirror repositories needed for the test and don't need to mirror all the default repositories.

The scc is an added behavior that change nothing  on default deployment. 

To use it, you need to add to module `base` the variable `server_registry_code` in your main.tf

The code has to be a Suse Manager Server activation code.

### DIFF
--- a/backend_modules/aws/base/main.tf
+++ b/backend_modules/aws/base/main.tf
@@ -3,6 +3,7 @@ locals {
   region            = lookup(var.provider_settings, "region", null)
   ssh_allowed_ips   = lookup(var.provider_settings, "ssh_allowed_ips", [])
   name_prefix       = var.name_prefix
+  server_registry_code = var.server_registry_code
 
   key_name = lookup(var.provider_settings, "key_name", null)
   key_file = lookup(var.provider_settings, "key_file", null)
@@ -61,7 +62,7 @@ data "aws_ami" "sles15" {
   }
 }
 
-data "aws_ami" "sles15sp1" {
+data "aws_ami" "sles15sp1o" {
   most_recent = true
   name_regex  = "^suse-sles-15-sp1-byos-v"
   owners      = ["013907871322"]
@@ -392,7 +393,7 @@ locals {
     ami_info = {
       opensuse152 = { ami = data.aws_ami.opensuse152.image_id },
       sles15      = { ami = data.aws_ami.sles15.image_id },
-      sles15sp1   = { ami = data.aws_ami.sles15sp1.image_id },
+      sles15sp1o   = { ami = data.aws_ami.sles15sp1o.image_id },
       sles15sp2o   = { ami = data.aws_ami.sles15sp2o.image_id },
       sles12sp5   = { ami = data.aws_ami.sles12sp5.image_id },
       sles12sp4   = { ami = data.aws_ami.sles12sp4.image_id },
--- a/backend_modules/aws/base/main.tf
+++ b/backend_modules/aws/base/main.tf
@@ -380,6 +380,7 @@ locals {
     name_prefix          = var.name_prefix
     use_shared_resources = var.use_shared_resources
     testsuite            = var.testsuite
+    server_registry_code = var.server_registry_code
 
     additional_network = local.additional_network
 
--- a/backend_modules/aws/base/main.tf
+++ b/backend_modules/aws/base/main.tf
@@ -3,7 +3,6 @@ locals {
   region            = lookup(var.provider_settings, "region", null)
   ssh_allowed_ips   = lookup(var.provider_settings, "ssh_allowed_ips", [])
   name_prefix       = var.name_prefix
-  server_registry_code = var.server_registry_code
 
   key_name = lookup(var.provider_settings, "key_name", null)
   key_file = lookup(var.provider_settings, "key_file", null)
@@ -381,7 +380,7 @@ locals {
     name_prefix          = var.name_prefix
     use_shared_resources = var.use_shared_resources
     testsuite            = var.testsuite
-    server_registry_code = var.server_registry_code
+    server_registration_code = var.server_registration_code
 
     additional_network = local.additional_network
 
--- a/backend_modules/aws/host/user_data.yaml
+++ b/backend_modules/aws/host/user_data.yaml
@@ -23,7 +23,6 @@ runcmd:
 %{ endif }
 
 %{ if image == "sles15"}
-
 zypper:
   repos:
     - id: os_pool_repo
@@ -34,6 +33,17 @@ zypper:
 
 packages: ["salt-minion"]
 %{ endif }
+
+%{ if image == "sles15sp2o"}
+runcmd:
+  - SUSEConnect --url https://scc.suse.com -r ${server_registry_code}
+  - SUSEConnect -p sle-module-python2/15.2/x86_64
+  - SUSEConnect -p sle-module-web-scripting/15.2/x86_64
+  - SUSEConnect -p sle-module-suse-manager-server/4.1/x86_64
+  - zypper in -y salt-minion
+
+%{ endif }
+
 %{ if image == "sles12sp3"}
 zypper:
   repos:
--- a/backend_modules/aws/host/user_data.yaml
+++ b/backend_modules/aws/host/user_data.yaml
@@ -34,13 +34,34 @@ zypper:
 packages: ["salt-minion"]
 %{ endif }
 
-%{ if image == "sles15sp2o"}
+%{ if server_registry_code != "null" }
+
+%{ if image == "sles15sp1o" }
+runcmd:
+  - SUSEConnect --url https://scc.suse.com -r ${server_registry_code}
+  - SUSEConnect -p sle-module-python2/15.1/x86_64
+  - SUSEConnect -p sle-module-web-scripting/15.1/x86_64
+  - SUSEConnect -p sle-module-suse-manager-server/4.0/x86_64
+  - zypper in -y salt-minion
+%{ endif }
+
+%{ if image == "sles15sp2o" }
 runcmd:
   - SUSEConnect --url https://scc.suse.com -r ${server_registry_code}
   - SUSEConnect -p sle-module-python2/15.2/x86_64
   - SUSEConnect -p sle-module-web-scripting/15.2/x86_64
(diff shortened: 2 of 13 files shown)


moio:
This would actually make sense across backends - unless I am missing something this is not an AWS-specific scenario.

rjmateus:
> This would actually make sense across backends - unless I am missing something this is not an AWS-specific scenario.

I Agree with Moio, and I would make sense for all backends. The implementation is also using the shared variables, so it should be implemented in all backends.
For backend-specific variables, we should pass data in the provider-settings map variable.

backend_modules/null/base/variables.tf
```diff
@@ -68,3 +68,8 @@ variable "images" {
   default     = ["centos6o", "centos7o", "centos8o", "opensuse150o", "opensuse151o", "opensuse152o", "opensuse153o", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4o", "sles12sp5o", "sles15o", "sles15sp1o", "sles15sp2o", "sles15sp3o", "ubuntu1604o", "ubuntu1804o", "ubuntu2004o"]
   type        = set(string)
 }
+
+variable "server_registry_code" {
+  description = "Use scc register"
```
Bischoff, line 73: SCC capital letters

do you mean "registration" instead of "register"?

modules/base/main.tf
```diff
@@ -15,6 +15,7 @@ module "base_backend" {
   testsuite            = var.testsuite
   provider_settings    = var.provider_settings
   images               = var.images
+  server_registry_code = var.server_registry_code
```
Bischoff, line 18: `registration` ?

what's this code ? if it's this `UC7` thing I think we already have it in another variable.

modules/base/variables.tf
```diff
@@ -68,3 +68,9 @@ variable "images" {
   default     = ["centos6o", "centos7o", "centos8o", "opensuse150o", "opensuse151o", "opensuse152o", "opensuse153o", "sles11sp4", "sles12", "sles12sp1", "sles12sp2", "sles12sp3", "sles12sp4o", "sles12sp5o", "sles15o", "sles15sp1o", "sles15sp2o", "sles15sp3o", "ubuntu1604o", "ubuntu1804o", "ubuntu2004o"]
   type        = set(string)
 }
+
+variable "server_registry_code" {
+  description = "Use scc register"
+  type        = string
+  default     = "null"
```
Bischoff, line 75: use empty string `""` as null value ?

salt/repos/server.sls
```diff
@@ -12,7 +12,7 @@ server_update_repo:
     - priority: 97
 {% endif %}
 
-{% if '4.0' in grains['product_version'] %}
+{% if '4.0' in grains['product_version'] and not grains.get('server_registry_code') %}
```
Bischoff, line 15: will this `not` work with the `'null'` value?

Bischoff (changes requested):
see inline comments, plus:

* same as Moio's and Ricardo's remarks
* it lacks explanation in `README_ADVANCED.md`
* beware that using SCC directly is a nice possibility, but it exposes us to:
    * failures at SCC
    * the case where a package is being updated at SCC right now
    * also, I'm not sure about bandwidth between the cloud and SCC

backend_modules/aws/base/main.tf
```diff
@@ -3,6 +3,7 @@ locals {
   region            = lookup(var.provider_settings, "region", null)
   ssh_allowed_ips   = lookup(var.provider_settings, "ssh_allowed_ips", [])
   name_prefix       = var.name_prefix
+  server_registry_code = var.server_registry_code
```
rjmateus, line 6: this variable is declared but never used.

backend_modules/aws/host/user_data.yaml
```diff
@@ -34,6 +33,38 @@ zypper:
 
 packages: ["salt-minion"]
 %{ endif }
+
+%{ if server_registry_code != "null" }
```
rjmateus, line 37: If we use the null type instead of null string, I'm not sure if this check will still work. 
Some tests are needed to make sure the test is behaving as expected.

backend_modules/aws/host/user_data.yaml
```diff
@@ -34,6 +33,38 @@ zypper:
...
+
+%{ if image == "sles15sp1o" }
+runcmd:
```
rjmateus, line 40: For readability, I think this if statements are goog.
But did you consider move the common parts to outside the if, and the condiction only generating the different connect statements?
Example:
```
runcmd:
  - SUSEConnect --url https://scc.suse.com -r ${server_registry_code}
%{ if image == "sles15sp1o" }
  - SUSEConnect -p sle-module-python2/15.1/x86_64
  - SUSEConnect -p sle-module-web-scripting/15.1/x86_64
  - SUSEConnect -p sle-module-suse-manager-server/4.0/x86_64
%{ endif }
  - zypper in -y salt-minion
```

modules/base/variables.tf
rjmateus: Or use the `null` type, as we have in line 28 of this file.

backend_modules/null/base/variables.tf
```diff
@@ -68,3 +68,8 @@ variable "images" {
...
+  default     = "null"
```
rjmateus, line 74: I think we should the `null` type in here instead of string null

salt/repos/default.sls
```diff
@@ -335,7 +335,7 @@ test_update_repo:
 {% endif %}
 {% endif %} {# '15' == grains['osrelease'] #}
 
-{% if '15.1' == grains['osrelease'] %}
+{% if '15.1' == grains['osrelease'] and grains['server_registry_code'] == 'null' %}
```
rjmateus, line 338: The comparation is made in two different ways. For example in file `salt/repos/server.sls` it is done in the form:
`... and not grains.get('server_registry_code') `

What should be the correct way to check if the varaible is empty? 
That will depend if we use the null type?

salt/repos/default.sls
```diff
@@ -350,7 +350,7 @@ os_ltss_repo:
 
 {% endif %} {# '15.1' == grains['osrelease'] #}
 
-{% if '15.2' == grains['osrelease'] %}
+{% if '15.2' == grains['osrelease'] and grains['server_registry_code'] == 'null' %}
```
rjmateus, line 353: some comment as above.

salt/server/init.sls
```diff
@@ -27,7 +27,7 @@ server_packages:
       - sls: repos
       - sls: server.firewall
 
-{% if '4' in grains['product_version'] and grains['osfullname'] != 'Leap' %}
+{% if '4' in grains['product_version'] and grains['osfullname'] != 'Leap' and grains['server_registry_code'] == 'null' %}
```
rjmateus, line 30: some comment as above

rjmateus (commented):
Thanks for the contribution.
I drop some comments.

modules/base/main.tf
maximenoel8: No, it's not the proxy code you need to enter but the SUMA SCC registration code.

modules/base/variables.tf
maximenoel8: I use "null" in quote because using null was breaking my code and I didn't find a better solution.

The issue is if `default = null` then `%{ if server_registration_code != "null" }` in user_data.yaml has an error because the value `server_registration_code` doesn't exist by default.

I will try using `""`, I don't know how the salt/grain will convert that value

If you have any better idea how to manage this I'm open to suggestion.

salt/repos/server.sls
maximenoel8: The value come from grains and not sumaform. It's converted as `null` by salt and this condition work.

maximenoel8:
I agree it should be added for the others providers. I will also need to add it for proxy.
I just want to see with you about this first PR and then I will add it to the others.

For the null variable. It was a big issue for me. I started by using `default = null` but when `server_registration_code` was not set, the cloud init will fail because the variable `server_registration_code` doesn't exist in this expression `%{ if server_registration_code != "null" }`

The null value is correctly converted in the salt/grain and so this expression work : 
`and grains['server_registration_code'] == 'null'`


To explain it simply, I want to skip all the scc register part if `server_registration_code` is not set ( cloud init description ). And skip all the internal repositories declaration if `server_registration_code` has a value ( salt execution ). If you have a better approach I'm open to it.

backend_modules/aws/host/user_data.yaml
maximenoel8: It looks good but I will also need to add a condition to `%{ if server_registration_code != "null" }` otherwise the registration and salt install will be executed for all the images if scc_registration_code is set.

Any idea how to write something like : 
`%{ if server_registration_code != "null" && image contains sles15sp }` ?

I didn't find a good documentation that describe how to use more complex condition in cloud-init .

modules/base/variables.tf
rjmateus: I think you should use null type as default value in here.
One possible solution is check the value before passing it to cloud init template. I will add a comment with a proposed solution.

backend_modules/aws/host/user_data.yaml
rjmateus: My idea was keep the outer if with the `%{ if server_registry_code != "null" }` and skip the all code block if the variable was not defined.
This code proposal could replace only the runcmd sections.

moio:
>     * the case where a package is being updated at SCC right now

At least this one is not a concern, infrastructure is there to make ample bandwidth available.

backend_modules/aws/host/user_data.yaml
maximenoel8: Yes I understand but the first runcmd outside of the image check will be executed for all the image if the SCC registration code is set.

rjmateus: Good point. in This case, I would say keep the code as it is right now :)

PR #926, opened 2021-07-21
Skip 'update-ca-certificates' run if the certs are updated automatically

cbbayburt:
In SLES15SP3, the directory `/etc/pki/trust/anchors` is watched via the `ca-certificates.path` systemd unit, and as soon as its contents changed, the system-wide certificates are updated automatically. In `certs` state, we run `update-ca-certificates` right after deploying certs. These two processes create a race condition that in rare cases leads to an error.

This PR updates the `certs` state to run `update-ca-certificates` only if this systemd unit does not exist or is not running.

See https://bugzilla.suse.com/1188500

### DIFF
--- a/salt/minion/testsuite.sls
+++ b/salt/minion/testsuite.sls
@@ -42,13 +42,17 @@ suse_certificate:
     - makedirs: True
 
 update_ca_truststore:
-  cmd.wait:
+  cmd.run:
     - name: /usr/sbin/update-ca-certificates
-    - watch:
+    - onchanges:
       - file: registry_certificate
       - file: suse_certificate
     - require:
       - pkg: suse_minion_cucumber_requisites
+    - unless:
+      - fun: service.status
+        args:
+          - ca-certificates.path
 
 {% endif %}
 
--- a/salt/repos/additional.sls
+++ b/salt/repos/additional.sls
@@ -42,7 +42,12 @@
 
 {% if grains['os'] == 'SUSE' %}
 update-ca-certificates:
-  cmd.run
+  cmd.run:
+    - name: /usr/sbin/update-ca-certificates
+    - unless:
+      - fun: service.status
+        args:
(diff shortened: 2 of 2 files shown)


rjmateus (approved):
looks good to me. I think this is the more natural solution we could get +1

Bischoff:
@cbbayburt beware this PR was missing two other calls to update-ca-certificates.

cbbayburt:
The testsuite error is caused by the `minion/testsuite.sls` only, but I'd expect the other state could cause the same problem when it's used (normally the testsuite doesn't use this state) so I'll update it as well. But I don't think we need to update `bashrc` at this moment. Even if we do, we need a different solution for that one.

Bischoff:
> The testsuite error is caused by the `minion/testsuite.sls` only, but I'd expect the other state could cause the same problem when it's used (normally the testsuite doesn't use this state) so I'll update it as well. But I don't think we need to update `bashrc` at this moment. Even if we do, we need a different solution for that one.

ok.

Rather than testing the presence of a service, i would rather simply test `sle15 sp2` grain and lower. Admitting this is needed.

cbbayburt:
> Rather than testing the presence of a service, i would rather simply test sle15 sp2 grain and lower. Admitting this is needed.

To give a bit more context: This issue also affects SUSE Manager production environment. We run a similar state during minion onboarding. Starting from that point, we discussed this in the dev team, and decided this would be the most natural solution.

The other reasonable option is of course filtering by distribution. But this wouldn't be as accurate as the other, simply because we don't know which other distributions/versions will support the auto-update or not. They might backport it, they might drop it in a future version, etc. but checking for the service is guaranteed to find out if the auto-update will happen.

cbbayburt:
I'm currently testing this with both SLES12 and SLES15 machines. When done, I'll open a new PR with the fix.

PR #929, opened 2021-07-23
Remove calls to update-ca-certificates

Bischoff:
## What does this PR change?

We currently hit this bug on SLE 15 SP3 minions:

https://bugzilla.suse.com/show_bug.cgi?id=1188500

From what I understand from https://www.mankier.com/8/update-ca-trust (but I am not a specialist), when you write files to `/etc/pki/trust/anchors/`, they are already at the right place, so you don't need to run `/usr/sbin/update-ca-certificates`. My impression is, our states to call it have always been unneeded.

This PR simply removes all calls to this utility.

@moio, opinion?

### DIFF
--- a/salt/controller/bashrc
+++ b/salt/controller/bashrc
@@ -78,6 +78,5 @@ export AUTH_REGISTRY_CREDENTIALS="{{ grains.get('auth_registry_username') }}|{{ 
 #### Generate certificates for Google Chrome
 if [ ! -f /etc/pki/trust/anchors/$SERVER.cert ]; then
   wget http://$SERVER/pub/RHN-ORG-TRUSTED-SSL-CERT -O /etc/pki/trust/anchors/$SERVER.cert
-  update-ca-certificates
   certutil -d sql:/root/.pki/nssdb -A -t TC -n "susemanager" -i  /etc/pki/trust/anchors/$SERVER.cert
 fi
--- a/salt/minion/testsuite.sls
+++ b/salt/minion/testsuite.sls
@@ -41,15 +41,6 @@ suse_certificate:
     - source: salt://minion/certs/SUSE_Trust_Root.crt.pem
     - makedirs: True
 
-update_ca_truststore:
-  cmd.wait:
-    - name: /usr/sbin/update-ca-certificates
-    - watch:
-      - file: registry_certificate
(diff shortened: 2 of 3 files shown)


calancha (approved):
We expect this will fix the problem at Uyuni/4.2.
Let's hope it's not going to hit us back at 4.1 as a side-effect.

cbbayburt:
> when you write files to /etc/pki/trust/anchors/, they are already at the right place, so you don't need to run /usr/sbin/update-ca-certificates. My impression is, our states to call it have always been unneeded.

@Bischoff This is only true for SLES15SP3 machines. For SLES12, we still need to run this state.

For a proper fix of this problem, I'll open this PR (https://github.com/uyuni-project/sumaform/pull/926) again soon with a small tweak

Bischoff:
> We expect this will fix the problem at Uyuni/4.2.
> Let's hope it's not going to hit us back at 4.1 as a side-effect.

I'm currently discussing with Fabian, he's very responsive. I'm trying to determine if i must remove it only for SLES 15 SP3, or for all versions.

Bischoff:
> > when you write files to /etc/pki/trust/anchors/, they are already at the right place, so you don't need to run /usr/sbin/update-ca-certificates. My impression is, our states to call it have always been unneeded.
> 
> @Bischoff This is only true for SLES15SP3 machines. For SLES12, we still need to run this state.
> 
> For a proper fix of this problem, I'll open this PR (#926) again soon with a small tweak

OK I did not know you already had a PR.

Beware your PR was missing two other calls to `update-ca-certificates`.

Are we really sure we still need it for SLE 12 ? What would the command exactly do?

cbbayburt:
> Are we really sure we still need it for SLE 12 ?

Yes, because AFAIK the auto-detection of the certificates is introduced with SLES15SP3.

> What would the command exactly do?

It updates the system-wide certificates with the contents of the `/etc/pki/trust/anchors` directory. Without that command, certs added here won't be used.

Bischoff:
> It updates the system-wide certificates with the contents of the `/etc/pki/trust/anchors` directory. Without that command, certs added here won't be used.

Thanks, you are perfectly right and I was wrong, it's in the man page, i just missed it!

Closing this in favour of your upcoming PR.

PR #930, opened 2021-07-23
Skip 'update-ca-certificates' run if the certs are updated automatically

cbbayburt:
In SLES15SP3, the directory `/etc/pki/trust/anchors` is watched via the `ca-certificates.path` systemd unit, and as soon as its contents changed, the system-wide certificates are updated automatically. In `certs` state, we run `update-ca-certificates` right after deploying certs. These two processes create a race condition that in rare cases leads to an error.

This PR updates the `certs` state to run `update-ca-certificates` only if this systemd unit does not exist or is not running.

See https://bugzilla.suse.com/1188500

### DIFF
--- a/salt/minion/testsuite.sls
+++ b/salt/minion/testsuite.sls
@@ -42,13 +42,19 @@ suse_certificate:
     - makedirs: True
 
 update_ca_truststore:
-  cmd.wait:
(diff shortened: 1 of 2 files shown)


Bischoff (approved):
lgtm

PR #934, opened 2021-08-02
 Rollback workaround for progress#95512

juliogonzalez:
## What does this PR change?

According to https://progress.opensuse.org/issues/95512, seems the replication problems for the images are fixed for all mirrors except one located in Indonesia.

It also seems that according to that ticker, we should always be using HTTP (which IMHO is wrong, the mirrordirector should point us to whatever protocol the mirror uses, regardless of us using HTTP or HTTPS for connecting to the mirrordirector). So I changed all references to download.opensuse.org (and download.suse.de just in case) to always use HTTP.


Bischoff (approved):
I agree using http sounds wrong to me, but if they say so... lgtm

calancha (approved):

PR #938, opened 2021-08-11
Fix: Use the right variable for the definition of terminal grains inside the controller

srbarrios:
## What does this PR change?

Fix: Use the right variable for the definition of terminal grains inside the controller


Bischoff (approved):

srbarrios:
Already fixed in a different PR.

PR #947, opened 2021-09-15
Add new product version uyuni-pr

jordimassaguerpla:
## What does this PR change?

We need this so we can not add the master repos, when testing Pull
Requests.

Repos will be added as additional repos, instead.

Each PR has a different repo, so it can not be added here, but instead
will be added to susemanager-ci sumaform config files.

I also changed the README to state "openSUSE Leap image" instead of "openSUSE Leap 15.2 image", because stating the version is outdated.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -19,7 +19,14 @@ Legal values for work-in-progress software are:
  * `4.1-nightly` (corresponds to the Build Service project Devel:Galaxy:Manager:4.1)
  * `4.2-nightly` (corresponds to the Build Service project Devel:Galaxy:Manager:4.2)
  * `head` (corresponds to the Build Service project Devel:Galaxy:Manager:Head, for `server` and `proxy`only works with SLE15SP3 image)
- * `uyuni-master` (corresponds to the Build Service project systemsmanagement:Uyuni:Master, for `server` and `proxy` only works with openSUSE Leap 15.1 image)
+ * `uyuni-master` (corresponds to the Build Service project systemsmanagement:Uyuni:Master, for `server` and `proxy` only works with openSUSE Leap image)
+
+Legal values for CI:
+
+`uyuni-pr` is a special product version used internally to test Pull Requests. Packages are under a subproject in systemsmanagement:Uyuni:Master:TEST and systemsmanagement:Uyuni:Master:PR.
+This is not meant to be used outside the Continous Integration system (CI).
+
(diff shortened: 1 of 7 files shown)


jordimassaguerpla:
Partial fix for https://github.com/SUSE/spacewalk/issues/15846

README_ADVANCED.md
```diff
@@ -19,7 +19,8 @@ Legal values for work-in-progress software are:
  * `4.1-nightly` (corresponds to the Build Service project Devel:Galaxy:Manager:4.1)
  * `4.2-nightly` (corresponds to the Build Service project Devel:Galaxy:Manager:4.2)
  * `head` (corresponds to the Build Service project Devel:Galaxy:Manager:Head, for `server` and `proxy`only works with SLE15SP3 image)
- * `uyuni-master` (corresponds to the Build Service project systemsmanagement:Uyuni:Master, for `server` and `proxy` only works with openSUSE Leap 15.1 image)
+ * `uyuni-master` (corresponds to the Build Service project systemsmanagement:Uyuni:Master, for `server` and `proxy` only works with openSUSE Leap image)
+ * `uyuni-pr` (corresponds to the Build Service projects under systemsmanagement:Uyuni:Master:TEST, for `server` and `proxy only works with openSUSE Leap image)
```
moio, line 23: Markdown broke here
```suggestion
 * `uyuni-pr` (corresponds to the Build Service projects under systemsmanagement:Uyuni:Master:TEST, for `server` and `proxy` only works with openSUSE Leap image)
```

moio, line 23: I am not enthusiastic with the name choice here (`uyuni-pr`) - I know your intent is to run tests on pull requests so that makes sense to me, but as a normal user I would not really understand.

How about `uyuni-test` given it is based on `TEST` projects?

jordimassaguerpla:
This should be merged after https://github.com/SUSE/susemanager-ci/pull/333

README_ADVANCED.md
jordimassaguerpla: `uyuni-test` can also be confusing because this product is for pull requests. The same way we have `uyuni-master`. What if we call it `uyuni-pull-request`or `uyuni-test-pr` ?

moio: OK, but then make clear in the documentation this is intended for the Uyuni CI only, and not really to be used by the casual README reader

jordimassaguerpla: ok. Done.

jordimassaguerpla: done.

PR #951, opened 2021-09-16
Use custom image for openSUSE clients when testing PRs

jordimassaguerpla:
## What does this PR change?

openSUSE official images are being constantly being updated. This
creates 2 problems:

1- The image can be different from build to build
2- Sometimes the image is not available

Given this is for testing PRs, we don't need to use the official one.
For master testing, yes, but for PRs, we are allowed to take some
shortcuts that add stability or velocity.

In this case, what we are doing is building the image once and disabling
the build. This way, once the image has been synced in all mirrors, there won't
be new builds.

Thus, by using our image, every build will have the same image, which
adds consistency and test stability.


jordimassaguerpla:
Partial fix for https://github.com/SUSE/spacewalk/issues/15846

rjmateus (approved):

PR #953, opened 2021-09-22
Fix user data for pr image

jordimassaguerpla:
## What does this PR change?

When changing the image for PRs to a custom one, we also need to change
the user data

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -111,7 +111,7 @@ runcmd:
   - zypper removerepo --all
 %{ endif }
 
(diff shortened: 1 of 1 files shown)


backend_modules/libvirt/host/user_data.yaml
```diff
@@ -111,7 +111,7 @@ runcmd:
   - zypper removerepo --all
 %{ endif }
 
-%{ if image == "opensuse153o" }
+%{ if image == "opensuse153o" || image == "opensuse153-ci-pr" || image = "opensuse153-ci-pr-client" }
```
srbarrios, line 114:
```suggestion
%{ if  "opensuse153" in image }
```

PR #956, opened 2021-09-23
Use a more stable image for the controller image

jordimassaguerpla:
## What does this PR change?

opensuse 15.2 images get updated very frequently. This goes to the
openSUSE mirrors. From time to time, images are not available in the
mirrors. This causes a failure when running sumaform.

Let's use a more stable image, that is not updated and, so it is available
most of the time, plus an image that is specific for the
controller, which its build is being disabled.

Moreover, this will make a run reproduceable. If we use an image that is
continuously being updated, runs are not reproduceable.


jordimassaguerpla:
partial fix for https://github.com/SUSE/spacewalk/issues/15846

juliogonzalez (commented):
I don't understand, why only for the controller? And why to to our own image?

In any case, mark this clearly as a workaround (grep for the word, as I don't remember the format of the comment). It seems there's some discussion and maybe the fix for this won't take months, so we need to be able to locate this to quickly revert when needed.

rjmateus:
I am not sure if we should do this. As Julio mentioned, the problem should be solved at opensuse level.
We start using the official images because we don't want to keep our own image builds since it means more work for our team.
One of our goals is to remove most of the currently existing self-made images.

Besides the problem at opensuse mirrors, we don't get much more benefits, since the machine doesn0t take to much time to be created.

In my opinion, one thing we should consider is to use the shared image feature from sumaform [1], and download the images only once. This can even be configured to be done periodically with a jenkins job.

[1] https://github.com/uyuni-project/sumaform/blob/master/README_ADVANCED.md#shared-resources-prefixing-sharing-virtual-hardware

jordimassaguerpla:
Let's use the opensuse official one and if we have problems again or it does not get fixed in the opensuse level, let's revisit this later.

jfyi, we use "terraform taint" to the vms, but not to the images, so images are not downloaded each time anyway.

For more info on why we are taking this decision, see https://github.com/SUSE/spacewalk/issues/15846#issuecomment-930143412

Thanks @rjmateus  and @juliogonzalez for your feedback anyway.

PR #957, opened 2021-09-24
Add support for ARM clients

Bischoff:
## What does this PR change?

This PR adds support for aarch64 clients.

See also https://github.com/SUSE/susemanager-ci/pull/347


backend_modules/libvirt/host/user_data.yaml
```diff
@@ -377,3 +377,9 @@ yum_repos:
 packages: ["avahi", "nss-mdns", "qemu-guest-agent", "salt-minion"]
 
 %{ endif }
+
+%{ if image == "opensuse153armo" }
+
+packages: ["salt-minion"]
```
rjmateus, line 383: this configuration is different from the one used in other opensuse 153 images:
https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/host/user_data.yaml#L114-L119
Doesn't it come with salt but have avahi and nss-mdns packages?

rjmateus (approved):
LGTM, just one small question

srbarrios (approved):

srbarrios:
If that grows for all versions and distros, maybe we want to split in different image_urls for each arch, or using maps or something similar, to organize it easily.

backend_modules/libvirt/host/user_data.yaml
Bischoff: I wondered the same. I am not sure about the answer. I think it is due to different base image: OpenStack image versus generic JeOS.

BTW, I missed the fact that openSUSE153o has this:
```
runcmd:
  - zypper removerepo --all
```
I think we need that too, I will add it now.

Bischoff:
> If that grows for all versions and distros, maybe we want to split in different image_urls for each arch, or using maps or something similar, to organize it easily.

I think that's a good idea indeed. But that means we need to introduce a new `arch` variable for the hosts.

Please notice that, on the disk, we need to differentiate the name anyway: `opensuse153o` has not the same contents as `opensuse153armo`.

PR #958, opened 2021-09-27
Don't include server_devel_repo in proxy vm

srbarrios:
## What does this PR change?

Recently we experience issues onboarding our Proxy in 4.2.
It seems related to this salt state https://github.com/SUSE/spacewalk/blob/Manager-4.2/susemanager-utils/susemanager-sls/salt/packages/init.sls#L2-L10 and having the server_devel_repo as repository in the proxy.

By removing this repo, the apply states works and this error is not shown anymore:
```
2021-09-27T11:45:47.638637+00:00 localhost systemd[1]: Started /usr/bin/zypper --non-interactive --no-refresh install --auto-agree-with-licenses --name sle-module-suse-manager-retail-branch-server-release sle-module-suse-manager-server-release.
2021-09-27T11:45:48.340157+00:00 localhost salt-minion[25376]: [ERROR   ] Command 'systemd-run' failed with return code: 4
2021-09-27T11:45:48.340354+00:00 localhost salt-minion[25376]: [ERROR   ] stdout: Loading repository data...
2021-09-27T11:45:48.340413+00:00 localhost salt-minion[25376]: Reading installed packages...
2021-09-27T11:45:48.340461+00:00 localhost salt-minion[25376]: Resolving package dependencies...
2021-09-27T11:45:48.340516+00:00 localhost salt-minion[25376]: 2 Problems:
2021-09-27T11:45:48.340586+00:00 localhost salt-minion[25376]: Problem: nothing provides 'product(sle-module-web-scripting) >= 15.3' needed by the to be installed product:sle-module-suse-manager-server-4.2-0.x86_64
2021-09-27T11:45:48.340705+00:00 localhost salt-minion[25376]: message repeated 2 times: [ Problem: nothing provides 'product(sle-module-web-scripting) >= 15.3' needed by the to be installed product:sle-module-suse-manager-server-4.2-0.x86_64]
2021-09-27T11:45:48.340749+00:00 localhost salt-minion[25376]:  Solution 1: do not install product:sle-module-suse-manager-server-4.2-0.x86_64
2021-09-27T11:45:48.340793+00:00 localhost salt-minion[25376]:  Solution 2: break sle-module-suse-manager-server-release-4.2-420.8.7.develHead.x86_64 by ignoring some of its dependencies
```

I'm not at all convinced of removing this repository, as we are not sure about other necessary packages inside this repository.
I would like the approval from @mcalmer to follow this approach and give an attempt to that solution.

For now, I will make a run using this branch.


juliogonzalez (commented):
The description of the PR is wrong. What you are removing is not the repository for the Server for development.

That's just the pool where all packages are building, common for proxy and server.

IIRC this repo was added in the past, because it contains some packages the developers need, that are NOT included at the images (which are repositories equivalent to what we have at the channel definitions).

Not sure if the testsuite needs any of such packages.

Since I didn't track this discussion either, I am neither approving or requesting changes.

srbarrios:
https://ci.suse.de/view/Manager/view/Manager-4.2/job/manager-4.2-dev-acceptance-tests-PRV/514/

srbarrios:
> That's just the pool where all packages are building, common for proxy and server.
> 

@juliogonzalez What do you mean by common for proxy and server? I only touch the file salt/repos/proxy.sls this is also used for the server at some point? :thinking:

juliogonzalez:
> @juliogonzalez What do you mean by common for proxy and server? I only touch the file salt/repos/proxy.sls this is also used for the server at some point? thinking

Talk to @mcalmer again, he was explaining it at Slack.

See also https://github.com/SUSE/spacewalk/wiki/OBS-repositories-and-packages,-and-more#projects-contents and https://github.com/SUSE/spacewalk/wiki/OBS-repositories-and-packages,-and-more#repository-examples

juliogonzalez:
> @juliogonzalez @mcalmer hmmm I think we could have a hidden problem if we define twice the same repos, as we do in the example lines I shared, but if you are very confident with this, then just omit my question :D

Didn't follow the conversation on Slack, so as I said, I not approving or asking for changes for the PR, as I am lacking context. But seems you were tracking this with @mcalmer, so if he approves, that should be enough.

srbarrios:
I created an issue in sumaform https://github.com/uyuni-project/sumaform/issues/959 to solve my question here. Which is related to my changes here, but not related with the reason why I proposed those changes.

mcalmer:
I changed proxy.sls to use the devel repos specific for "proxy" and not the global devel repo which hold all packages for Server, Proxy and RTBS in one repo. That repo create the mess as it also include the server and rtbs products.

mcalmer (approved):

jordimassaguerpla:
Why the proxy tries to install the server product?

mcalmer:
> Why the proxy tries to install the server product?

Because it is available. We have a SUMA state which install all available products. This is required for subscription matching.
Otherwise users can add a channel and only install the software, but not the product. This would lead into wrong results as we match against products.

mcalmer:
Btw. it worked, as the test run brought us one step forward. I would merge it.

ktsamis:
I will merge this and we can move to the next failure

juliogonzalez:
The title is still wrong, as `s/Devel:/Galaxy:/Manager:/X.Y/SLE_15_SPX/` is not a server repo, as I told above, but @meaksh just found one reason for having it: salt

Salt is NOT part of the images generated by the product definition, as it should come from SLE. So the way of making it available for server and proxy was adding that repository.

juliogonzalez:
In other words: without this repository, the proxy (and maybe the server) are not testing the package will be submitting as part of the next MU (and I think it happens the same with other packages such as node_exporter, basically for all packages that are part of SLE and not SUSE Manager)

srbarrios:
> In other words: without this repository, the proxy (and maybe the server) are not testing the package will be submitting as part of the next MU (and I think it happens the same with other packages such as node_exporter, basically for all packages that are part of SLE and not SUSE Manager)

As I said in the description, and several times in the chats... "I'm not at all convinced of removing this repository, as we are not sure about other necessary packages inside this repository." :)
But I will let @mcalmer continue with this discussion, see if you guys solve it, and I will go back to my cave to my other refactors ;)

PR #961, opened 2021-09-28
fix duplicate repos and wrong URLs

mcalmer:
## What does this PR change?

with the last change we had a duplicate repo introduced on head/uyuni.
For older versions we had a bit wrong directory names which should be fixed now.


ktsamis (approved):
Thanks very much for catching and fixing this

srbarrios (approved):

PR #962, opened 2021-09-30
Use an untrusted repo for a pull request testing

jordimassaguerpla:
The repo generated for pull request testing is not signed. Thus, for
this particular case, we need to trust the repo.

Partial fix for: https://github.com/SUSE/spacewalk/issues/15846

### DIFF
--- a/salt/repos/additional.sls
+++ b/salt/repos/additional.sls
@@ -4,9 +4,14 @@
   pkgrepo.managed:
     - humanname: {{ label }}
   {%- if grains['os_family'] == 'Debian' %}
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
The reason why the repos are not signed is because it was making things more complicated. It is not only to sign, but to store the secret key and how to distribute the key. I mean, if we are concerned that the package we try to install could have been "replaced"... the key could also be, if they are both downloaded... so it was adding some complications which look too much for a test system.

srbarrios (approved):

rjmateus (approved):

PR #964, opened 2021-10-01
Add "basesystem overlay" testing repo to the proxy

meaksh:
## What does this PR change?

This PR adds a new repository in the proxy systems in order to get devel packages for those that are not part of SUMA Proxy channels, like "salt".

These new repositories "Basesystem-Overlay" and "openSUSE-Overlay" currently contains the Salt package + some new dependencies which are not part of SLE.

This way, the deployed proxy will have the devel testing package for Salt and not the latest released Salt in SLE or Leap.

### DIFF
--- a/salt/repos/proxy.sls
+++ b/salt/repos/proxy.sls
@@ -125,6 +125,18 @@ server_devel_releasenotes_repo:
     - priority: 96
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


mcalmer (approved):

juliogonzalez (approved):
Looks good!

Remember to add 4.1 as well, next Monday.

PR #985, opened 2021-10-22
Add again empty repos for 15sp2 and 15sp3 LTSS in advance

ktsamis:
## What does this PR change?

Add again empty repos for 15sp2 and 15sp3 LTSS in advance because in the testsuite we are using these. Across different branches 4.1 and 4.2 different products are used and 15sp1 in on ltss but not the rest yet. We can't simply remove the os_ltss_repo from the steps that use it because this step is run against different products, some need the LTSS some don't. Also in the future when the level of support changes for 15sp2 we will bump into this again.

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -319,7 +319,6 @@ os_update_repo:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Basesystem/15-SP2/x86_64/update/
 
-uncomment when it goes LTSS
 os_ltss_repo:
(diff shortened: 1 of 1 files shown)


ktsamis:
In theory we don't need these yet and we can remove the os_ltss_repo from the steps that they are used in the testsuite. But the problem is then a difference between branches in 4.1 and 4.2 and that will cause confusion. Also we will have the same issue when 15sp2 becomes ltss. We could go that way as well, this was just easier to fix at the time and also less confusing I guess for people. It also has the advantage that it's more robust for the next year or so.

ktsamis:
Either way the removal of os_ltss_repos needs to go together with the removal of testsuite steps that use it, or we need to readd them

PR #989, opened 2021-11-08
Allow Jenkins VMs

juliogonzalez:
## What does this PR change?

Enable deployment of a Jenkins instance with sumaform that:
- Installs the latest LTS version from OBS
- Creates a user `admin` with a random password stored at `/var/lib/jenkins/secrets/initialAdminPassword`
- Enables default security for Jenkins (integrated DB)
- Installs the required plugin run an acceptance test pipeline, and swarm, so we can later add workers

# Workarounds

Since we don't have workers yet, I had to do some workarounds at Jenkins so it would also act as worker:
- Install sumaform
- Create `/home/jenkins`, `/home/jenkins/.ssh`, create a private/public keypair, and copy `/home/jenkins/.credentials` from the internal Jenkins-worker to the same location at the AWS Jenkins. This allowed us to have a pipeline that's no different from others
- Link `/home/jenkins/.ssh` at `/var/lib/jenkins` so sumaform doesn't complain, as it's the home for the `jenkins` user.
- Add a label `sumaform-cucumber` to the worker for Jenkins itself.
- Copy the public key to the controller, and create it as `uyuni-jenkins` at AWS, to be used by the pipeline later.

## For a a subsequent PR

- [ ] Configure Apache on Jenkins to use HTTPS and Let's encrypt certificates, redirect traffic from TCP 80 to TCP 443
- [ ] Configure Apache on Jenkins to work as a reverse http proxy
- [ ] Configure Jenkins itself to be able to work with the proxy
- [ ] Create a separate security group for Jenkins, and allow traffic for ports 80 and 443
- [ ] Provide Jenkins workers, with terraform installed. They should be able to start EC2 resources using a role, not access/secret keys. Consider creating a separate Jenkins user for the swarm onboarding.
- [ ] Needless to say, email notifications don't work. We could setup a SMTP server at Jenkins using `uyuni-project.org`, but still that's usually detected a a source of spam by email filters, so the best approach is using [AWS SES](https://aws.amazon.com/ses/).
- [ ] We should not pass though the controller, if we want to work from the AWS CI.

Related: https://github.com/SUSE/susemanager-ci/pull/370

### DIFF
--- a/modules/jenkins/main.tf
+++ b/modules/jenkins/main.tf
@@ -1,0 +1,28 @@
+
+module "jenkins" {
+  source = "../host"
+
+  base_configuration      = var.base_configuration
+  name                    = "jenkins"
+  use_os_released_updates = var.use_os_released_updates
+  additional_repos        = var.additional_repos
+  additional_packages     = var.additional_packages
+  swap_file_size          = var.swap_file_size
+  ssh_key_path            = var.ssh_key_path
+  roles                   = ["jenkins"]
+
+  grains = {
+    mirror                  = var.base_configuration["mirror"]
+    data_disk_fstype        = var.data_disk_fstype
+  }
+
+  image = var.image
+
+  provider_settings        = var.provider_settings
+  additional_disk_size     = var.data_disk_size
+  volume_provider_settings = var.volume_provider_settings
+}
+
+output "configuration" {
+  value = module.jenkins.configuration
+}
--- a/modules/jenkins/versions.tf
+++ b/modules/jenkins/versions.tf
@@ -1,0 +1,4 @@
+
+terraform {
+  required_version = ">= 1.0.10"
+}
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -116,6 +116,38 @@ module "suse-minion" {
 
 This will create 10 minions connected to the `server`.
 
+## Jenkins (PoC)
+
+If you are running `sumaform` for Uyuni and you want Continuous Integration, you can a setup a Jenkins instance using the `jenkins` module.
+
+For now, the module provides Jenkins with the following plugins enabled:
+
+- swarm
+- git
+- git-client
+- workflow-aggregator
+- extended-choice-parameter
+- timestamper
+- htmlpublisher
+- rebuild
+- http_request
+- ansicolor
+- greenballs
+
+Authentication is enabled, and a user `admin` is created. The password can be found at `/var/lib/jenkins/secrets/initialAdminPassword` at the Jenkins instance.
+
+To enable Jenkins, use the following definition.
+
+```hcl
+module "jenkins" {
+  source             = "./modules/jenkins"
+  base_configuration = module.base.configuration
+}
+```
+
+Usually you will want to use this on public clouds, but if you want to use this in libvirt, you are encouraged to use a separate pool, as explained for the mirror below.
+
+
 ## Mirror
 
 If you are using `sumaform` outside of the SUSE Nuremberg network you should use a special extra virtual machine named `mirror` that will cache packages downloaded from the SUSE engineering network for faster access and lower bandwidth consumption.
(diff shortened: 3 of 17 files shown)


modules/jenkins/main.tf
```diff
@@ -0,0 +1,29 @@
...
+  additional_repos        = var.additional_repos
+  additional_packages     = var.additional_packages
+  swap_file_size          = var.swap_file_size
+  ssh_key_path            = var.ssh_key_path
+  roles                   = ["jenkins"]
+
+  grains = {
+    cc_username             = var.base_configuration["cc_username"]
+    cc_password             = var.base_configuration["cc_password"]
+    data_disk_fstype        = var.data_disk_fstype
```
juliogonzalez, line 17: I wonder if we need `    mirror                 = var.base_configuration["mirror"]` here, so the mirror is used, if available?

juliogonzalez, line 17: I wonder if we need `mirror = var.base_configuration["mirror"]` here, so the mirror is used, if available?

rjmateus: That you help speedup OS updates for sure. Maybe a good option to add it, the same way we have it for other machine types.

modules/jenkins/main.tf
```diff
@@ -0,0 +1,29 @@
+
+module "jenkins" {
+  source = "../host"
+
+  base_configuration      = var.base_configuration
+  name                    = "jenkins"
```
rjmateus, line 6: Do you want a static name or should we allow user to define it and have "jenkins" as the default?

modules/jenkins/versions.tf
```diff
@@ -0,0 +1,4 @@
+
+terraform {
+  required_version = ">= 0.12"
```
rjmateus, line 3: after migration to terraform 1.0.10 this needs to be updated.

modules/jenkins/main.tf
```diff
@@ -0,0 +1,29 @@
...
+  use_os_released_updates = var.use_os_released_updates
+  additional_repos        = var.additional_repos
+  additional_packages     = var.additional_packages
+  swap_file_size          = var.swap_file_size
+  ssh_key_path            = var.ssh_key_path
+  roles                   = ["jenkins"]
+
+  grains = {
+    cc_username             = var.base_configuration["cc_username"]
```
rjmateus, line 15: Are these variables used?

rjmateus (commented):
LGTM, just some small comments

modules/jenkins/main.tf
juliogonzalez: I am not 100% sure, are they used for the repos or only for mirroring? If for the repos, they will be used if sle15 is used and no mirror is present. If not, then I guess a Jenkins based on SLE15 won't work without a mirror, so I better define the image as a constant, not as a variable.

modules/jenkins/main.tf
juliogonzalez: In this case I see it as the case for the mirror. Unlike minions or other VMs, there are no reasons to allow a name change, FMPOV (the mirror also has the name hardcoded)

modules/jenkins/versions.tf
juliogonzalez: Solved.

modules/jenkins/main.tf
juliogonzalez: Done.

modules/jenkins/main.tf
rjmateus: those variables are used by:
- mirror: To download the packages
- Server: to add them to server scc credentials
- Contoller: I don't know the usage in here

In this case, I think jenkins would need it to pass as properties for the terraform files it will deploy (the same way I think is being done now in the existing jenkins worker). But now in the code they are not in use.

juliogonzalez: I will remove them, and will hardcode the image. IIUC sumaform is not able to register SLE clients directly at SCC right now.

juliogonzalez: On a second thought, I will not hardcoded it. Even the mirror has the image as variable. I guess because that way it's easier to replace a leap version with another one :-)

juliogonzalez: Done.

PR #990, opened 2021-11-08
Allow refresing salt states for AWS and Azure

juliogonzalez:
## What does this PR change?

Before this:
```
module.jenkins.module.jenkins.module.host.null_resource.host_salt_configuration[0] (remote-exec): mv: cannot move '/tmp/salt' to '/root/salt': File exists
```
After this:
The states are refreshed at `/root/salt`


rjmateus (approved):
LGTM

PR #997, opened 2021-11-12
Add a http_proxy to Jenkins, required by the Uyuni testsuite

juliogonzalez:
## What does this PR change?

Add a http_proxy to Jenkins, required by the Uyuni testsuite

It uses a new container, defined at https://build.opensuse.org/package/show/systemsmanagement:Uyuni:Master:Docker/proxy, which is basically the same thing as the proxy we have for the tests at the internal infra.

I deployed this, and works, this is a test from the bastion to the http proxy at Jenkins:
```
ec2-user@ip-172-16-0-82:~> telnet 172.16.1.252 3128
Trying 172.16.1.252...
Connected to 172.16.1.252.
Escape character is '^]'.
^CConnection closed by foreign host.
```


rjmateus (approved):
LGTM

PR #998, opened 2021-11-15
Rewrite scc support on AWS

maximenoel8:
## What does this PR change?

In the goal to make the controller deployment compatible with AWS, I realize that I will need to handle SuseConnect registration for sles clients / minions. This new addition will have strongly complexify the cloud-init deployment.

To solve this issue, I redesign the cloud-init deployment to only install salt on sles and then handle the registration part with salt states. It will be easier to configure the different SCC registration needs.


salt/repos/init.sls
```diff
@@ -14,7 +14,7 @@ include:
 {% if grains['os'] == 'SUSE' %}
 refresh_repos:
   cmd.run:
-    - name: zypper --non-interactive --gpg-auto-import-keys refresh
+    - name: zypper --non-interactive --gpg-auto-import-keys refresh; exit 0
```
srbarrios, line 17: I would better apply this change in a different PR, as it doesn't seems specifically related to AWS:

srbarrios (approved):
LGTM. I would just move apart from this PR, a small change that applies not only to AWS deployments.

moio:
> I realize that I will need to handle SuseConnect registration for sles clients / minions

I am curious/might lack some parts of the picture: why is that required?

maximenoel8:
> > I realize that I will need to handle SuseConnect registration for sles clients / minions
> 
> I am curious/might lack some parts of the picture: why is that required?

To speed up AWS testing and avoid mirroring all the native repositories for SLES, I'm using SuseConnect method so I only need to synchronize the specific repositories. Currently, Suma Proxy and Server is supported but I would like to also support basic sles server.

PR #999, opened 2021-11-16
Fix empty string for terraform update

jordimassaguerpla:
[] is not the same as [""]


jordimassaguerpla:
it didnt' fix the issue

PR #1000, opened 2021-11-17
Install "venv-salt-minion" in all images for Uyuni Master

meaksh:


juliogonzalez (approved):
Looks good. Let's give it a try.

PR #1002, opened 2021-11-18
Add a container registry utility module

moio:
## What does this PR change?

It adds an utility VM to run a container registry.


moio:
CC @ncounter. Please note how the systemd unit file was generated (almost) automatically by podman - that's a pretty straightforward way to create services based on containers we will want to re-use.

PR #1006, opened 2021-11-19
Install Salt bundle based on 'install_salt_bundle' base config

meaksh:
## What does this PR change?

This PR fix the current issues caused by previous commit in order to enable the `venv-salt-minion` installation. 

This PR introduces a new configuration parameter for the "base" configuration named `install_salt_bundle` (default: false).


juliogonzalez (approved):
It's even better than the previous code, as now we can disable/enable the bundle.

Looks great. Let's merge it and try it.

PR #1008, opened 2021-11-19
Add "post_provisioning_cleanup" stage to cleanup Salt packages when using Bundle

meaksh:
## What does this PR change?

see title

### DIFF
--- a/salt/post_provisioning_cleanup.sh
+++ b/salt/post_provisioning_cleanup.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# If we are using Salt Bundle (venv-salt-minion), then we need to remove
+# the original Salt package installed on the instance.
+
+# Nothing to do in case "install_salt_bundle" grain is not true
+if ! grep -q '"install_salt_bundle": true' /etc/salt/grains; then
+    exit 0
+fi
+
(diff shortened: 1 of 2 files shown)


salt/post_provisioning_cleanup.sh
```diff
@@ -0,0 +1,30 @@
...
+    INSTALLER=zypper
+elif [ -x /usr/bin/yum ]; then
+    INSTALLER=yum
+elif [ -x /usr/bin/apt ]; then
+    INSTALLER=apt
+fi
+
+echo "Removing Salt packages, except Salt Bundle (venv-salt-minion) ..."
+if [ "$INSTALLER" == "zypper" ] || [ "$INSTALLER" == "dnf" ] || [ "$INSTALLER" == "yum" ]; then
+zypper -q --non-interactive remove salt salt-minion python3-salt python2-salt ||:
```
juliogonzalez, line 25: Shouldn't this be `${INSTALER}` and not `zypper`? Maybe you want to use it below as well.

juliogonzalez: Ok, right, not `${INSTALLER}` since yum/dnf and zypper syntax is not common, but this should work.

juliogonzalez:
Looks nice, let's give it a try.

PR #1009, opened 2021-11-22
Support dmacvicar/libvirt when defining multiple providers

srbarrios:
## What does this PR change?

Allow us to define multiple providers (with their aliases) and use the libvirt provider that suppports latest terraform.

This is only a temporary solution that must not be merged.
@rjmateus will come with a proper solution.


rjmateus:
closing in favor of: https://github.com/uyuni-project/sumaform/pull/1014

PR #1010, opened 2021-11-22
Fix propagation of "install_salt_bundle" after last refactor

meaksh:
## What does this PR change?

This PR fixes a problem with newly introduced `install_salt_bundle` which is not working properly after the last refactor.

This also copies the `/etc/salt/grains` file deployed by sumaform to `/etc/venv-salt-minion/grains` in case `install_salt_bundle` is set.


rjmateus (approved):
LGTM

PR #1014, opened 2021-11-24
version files as symbolic link to backend in use

rjmateus:

## What does this PR change?

If we use multiple libvirt providers definitions it needs each module that doesn't use the default.
Since each provider independent module doesn't define the provider namespace in use, terraform tries to resolve the same to `hashicorp/libvirt` instead of using the name space defined in the root main tf file.

With this change, all provider indepoendent module version file is now a symbolic link to the provider in use. With this approach we make sure all needed providers are defined for each module and we maintain compatibility with all backend providers.

@srbarrios ping

### DIFF
--- a/modules/base/versions.tf
+++ b/modules/base/versions.tf
@@ -1,3 +1,1 @@
-terraform {
-  required_version = "1.0.10"
-}
+../backend/base/versions.tf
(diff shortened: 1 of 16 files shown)


Bischoff (approved):
According to https://ci.suse.de/view/Manager/view/Manager-4.1/job/manager-4.1-qa-build-validation/126/console (run on branch `version_file_symblink`), it works.

PR #1015, opened 2021-11-26
Optional parameter to accept a greater range of SSL protocol versions

srbarrios:
## What does this PR change?

As our server and proxy has a default Apache configuration that doesn't allow old SSL Protocols, our SLES11SP4 or CentOS6 systems will require to manually that config. This parameter automated it for us.

See also https://github.com/SUSE/spacewalk/issues/15659

### DIFF
--- a/modules/proxy/variables.tf
+++ b/modules/proxy/variables.tf
@@ -138,3 +138,8 @@ variable "proxy_registration_code" {
   description = "SUMA SCC registration code to enable the SLES and SUMA repositories for proxy"
   default     = null
 }
+
+variable "accept_all_ssl_protocols" {
+  description = "Turn to true to force Apache to accept a greater range of protocol versions"
+  default     = false
+}
--- a/modules/server/variables.tf
+++ b/modules/server/variables.tf
@@ -254,3 +254,8 @@ variable "server_registration_code" {
   description = "SUMA SCC registration code to enable the SLES and SUMA repositories for server"
   default     = null
 }
+
+variable "accept_all_ssl_protocols" {
+  description = "Turn to true to force Apache to accept a greater range of protocol versions"
+  default     = false
+}
--- a/modules/proxy/main.tf
+++ b/modules/proxy/main.tf
@@ -51,6 +51,7 @@ module "proxy" {
     publish_private_ssl_key   = var.publish_private_ssl_key
     repository_disk_size      = var.repository_disk_size
     proxy_registration_code   = var.proxy_registration_code
+    accept_all_ssl_protocols  = var.accept_all_ssl_protocols
   }
 
   image                    = var.image == "default" || var.product_version == "head" ? var.images[var.product_version] : var.image
--- a/modules/server/main.tf
+++ b/modules/server/main.tf
@@ -75,6 +75,7 @@ module "server" {
     repository_disk_size           = var.repository_disk_size
     forward_registration           = var.forward_registration
     server_registration_code       = var.server_registration_code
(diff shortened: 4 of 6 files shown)


modules/proxy/variables.tf
```diff
@@ -138,3 +138,8 @@ variable "proxy_registration_code" {
   description = "SUMA SCC registration code to enable the SLES and SUMA repositories for proxy"
   default     = null
 }
+
+variable "accept_all_ssl_protocols" {
+  description = "turn to true to force Apache to accept a greater range of protocol versions"
```
Bischoff, line 143: Turn (capital)

modules/server/variables.tf
```diff
@@ -254,3 +254,8 @@ variable "server_registration_code" {
   description = "SUMA SCC registration code to enable the SLES and SUMA repositories for server"
   default     = null
 }
+
+variable "accept_all_ssl_protocols" {
+  description = "turn to true to force Apache to accept a greater range of protocol versions"
```
Bischoff, line 259: Same

Bischoff (approved):
LGTM. I was not able to check the Salt `requires:`.

2 nitpicks.

modules/proxy/variables.tf
srbarrios: Let's keep as other descriptions

moio:
I am probably lacking some context, I do not have the full picture.

Isn't CentOS 6 out of support right now?
Did we document anywhere that Apache configuration changes are needed to work with SLE 11, which is still supported?

modules/proxy/variables.tf
Bischoff: Precisely. They all use capitals.

Bischoff: Hmmm, the ones immediately above use capitals, but in fact there's a mixture. Crap.

srbarrios:
> I am probably lacking some context, I do not have the full picture.
> 
> Isn't CentOS 6 out of support right now?
> Did we document anywhere that Apache configuration changes are needed to work with SLE 11, which is still supported?

Yes, CentOS 6 is out :)
For SLES11 we have these configuration changes mentioned in the troubleshooying section.

This was working in our test automation only because the proxy was not secured, as the server was. But we did some recent changes in the proxy and now is secured too, so we need to force this concrete change.

It might also be interesting for @rjmateus in AWS envs, as I noticed he had also problems on that part.

modules/cucumber_testsuite/main.tf
```diff
@@ -55,6 +55,8 @@ module "server" {
   monitored                      = true
   use_os_released_updates        = true
   install_salt_bundle            = lookup(local.install_salt_bundle, "server", false)
+  accept_all_ssl_protocols       = lookup(local.accept_all_ssl_protocols, "server", false)
```
rjmateus, line 58: This local variable is not defined, so it cannot be use

rjmateus (commented):
One question about a variable

modules/cucumber_testsuite/main.tf
srbarrios: Aha, it seems I missed some line, thanks :)

modules/proxy/variables.tf
srbarrios: I made you happy, and changed by Turn this description, but I will not change others, feel free to change grammar of every other descriptions in another PR ;)

rjmateus (approved):
LGTM

PR #1016, opened 2021-11-29
AlmaLinux 8.4 - fix qcow url

hustodemon:
it seems the old url (the `/8/`directory) points to the version 8.5 now. Fixed by explicitly stating `8.4`.

### DIFF
--- a/backend_modules/libvirt/base/main.tf
+++ b/backend_modules/libvirt/base/main.tf
@@ -2,7 +2,7 @@
 locals {
   images_used = var.use_shared_resources ? [] : var.images
   image_urls = {
-    almalinux8o     = "${var.use_mirror_images ? "http://${var.mirror}" : "https://repo.almalinux.org"}/almalinux/8/cloud/x86_64/images/AlmaLinux-8-GenericCloud-8.4-20210616.x86_64.qcow2"
+    almalinux8o     = "${var.use_mirror_images ? "http://${var.mirror}" : "https://repo.almalinux.org"}/almalinux/8.4/cloud/x86_64/images/AlmaLinux-8-GenericCloud-8.4-20210616.x86_64.qcow2"
     centos6o        = "${var.use_mirror_images ? "http://${var.mirror}" : "https://cloud.centos.org"}/centos/6/images/CentOS-6-x86_64-GenericCloud.qcow2"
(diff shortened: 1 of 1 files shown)


hustodemon:
don't review yet. i'm still testing it.

backend_modules/libvirt/base/main.tf
```diff
@@ -2,7 +2,7 @@
 locals {
   images_used = var.use_shared_resources ? [] : var.images
   image_urls = {
-    almalinux8o     = "${var.use_mirror_images ? "http://${var.mirror}" : "https://repo.almalinux.org"}/almalinux/8/cloud/x86_64/images/AlmaLinux-8-GenericCloud-8.4-20210616.x86_64.qcow2"
```
rjmateus, line 5: I'm not sure if we should stick with version 8.4 or if we should use 8.5 instead since we want to test alma linux 8, and normally we use the latest one.
If move to version 8.5 we should change only the file name to `AlmaLinux-8-GenericCloud-8.5-20211119.x86_64.qcow2`
QA may have something to say about it.

rjmateus (commented):
One question

backend_modules/libvirt/base/main.tf
mcalmer: Why not use the "latest" name. This would make us independent from future version change.

`AlmaLinux-8-GenericCloud-latest.x86_64.qcow2 `

hustodemon: That's not up to me to decide. I only wanted to fix sumaform.

rjmateus: @ktsamis @srbarrios @Bischoff What are your thoughts about this? what version would make more sense here from the QA point of view?

meaksh: I just noticed this PR now, and in the meantime we merged this other one: https://github.com/uyuni-project/sumaform/pull/1019, which uses `latest` image.

Sorry for not checking in here before going with my own PR. In case you feel we should do it differentely, just let me know.

hustodemon: Don't worry. I'm glad it's fixed! Closing this one.

PR #1021, opened 2021-12-13
Terminals are known only by their MAC address

Bischoff:
## What does this PR change?

Correct generation of Retail information

See also uyuni-project/uyuni#4567

Tested to work on 4.1 BV.

This PR also removes build hosts and terminals for sles12sp4 and sles15sp2, forgotten in a previous PR. We now use sles12sp5 and sles15sp3.

### DIFF
--- a/modules/controller/variables.tf
+++ b/modules/controller/variables.tf
@@ -90,15 +90,6 @@ variable "pxeboot_configuration" {
   }
 }
 
-variable "terminal_configuration" {
-  description = "use module.<TERMINAL_NAME>.configuration, see main.tf.libvirt-testsuite.example"
-  default = {
-    hostnames = []
-    macaddr = null
-    image = null
-  }
-}
-
 variable "kvmhost_configuration" {
   description = "use module.<VIRTHOST_NAME>.configuration, see main.tf.libvirt-testsuite.example"
   default = {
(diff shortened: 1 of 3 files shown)


modules/controller/variables.tf
```diff
@@ -410,23 +401,7 @@ variable "sle15sp3_buildhost_configuration" {
...
-    hostnames = []
-  }
-}
-
-variable "sle15sp2_terminal_configuration" {
-  description = "use module.<SLE15SP2_TERMINAL>.configuration, see main.tf.libvirt-testsuite.example"
-  default = {
-    macaddr = null
-    image = null
+    macaddrs = []
```
ktsamis, line 404: images is not needed anymore? Also are you sure you are not changing the type of variable in macaddrs? It's a new variable right? Otherwise it's a typo from `macaddr` that it was before and `macaddrs` that it is now

Bischoff: It was possible to use `macaddr` instead of `macaddrs` because of a terraform module named `pxe_boot`  (in `sumaform/modules/pxe_boot` where you can set up both the mac address and the image.

In the Retail terminals case, Lukas just made them derive from `minion`s. I did not want to change that now. That means they can't have these image or mac address fields, and that the MAC addresses list directly inherited from the `host` module (of which `minion` module derives).

If we change this choice, we can have terminals derive from `pxe_boot` module too. But I will not do it unless it appears necessary.

PR #1017, opened 2021-11-29
update libvirt terraform provider

rjmateus:
Signed-off-by: Ricardo Mateus <rmateus@suse.com>

## What does this PR change?

use uyuni terraform libvirt proivder lateste version

PR #1023, opened 2021-12-27
update terraform libvirt provider ro 0.6.12

rjmateus:

## What does this PR change?

Updates sumaform to use terraform loibvirt provider latest provider from terraform registry.
We will be using the upstream provider since all patches were included in the latest released version.
We will keep the repository and provider for uyuni in case we need to add another patch in the future.

# How to upgrade
 - Un-install the provider's package `zypper rm terraform-provider-libvirt`
 - In you main.tf, change the provider version to `0.6.12`
 - If you want to keep the same state file, see `keeping state file`
 - Run `terraform init -upgrade`

## Keeping state file
- Open the`terraform.tfstate`
- Search for all `cpu` keys. The value should now be an array of objects. To change it just surround the existing content with square brackets (`[ {....} ]` )


juliogonzalez (approved):
Changes LGTM, but I fail to understand why you are asking to remove the old RPM instead of just updating. I guess it was already discussed with QA.

I think you will also need to send an email to the mailing lists about this, as it can crash environments for the developers.

juliogonzalez:
> @juliogonzalez we need to remove because the way the plugin is provided will change. With version 0.6.3 it was provided by a package built by us on OBS. For version 0.6.12 we are relying on the terraform registry service and getting the provider directly from the official location.

100% correct. I didn't remember we were working on this before the holidays :-D

> When this PR is merged, we need to:
> 
> * Send email to mailing list
> * Adapte CI terraform files to cope with the cahnge (PR already opened)
> * Clean CI environment to cope with the changes
> * Remove package from al lour CI worker machines.

PR LGTM. I am approving now. Feel free to merge as soon as QE is happy about it.

ktsamis (approved):
Looks good but let's schedule it please. We can do it on a Monday so we do the cleanups/removal/installation etc at the end of the day and start again all envs. That way we see on Tuesday if there is a problem.

srbarrios (approved):

SchoolGuy:
As per https://github.com/uyuni-project/sumaform/pull/1017#issuecomment-1031422221 we should update this once again if we don't want to get trouble with our SSH Servers.

nkrinner:
Some more context here: https://github.com/dmacvicar/terraform-provider-libvirt/issues/916 (provieded by @nodeg )

Background: The current provider will not work with an updated ssh (from version 8.8) and machines will not be accessible.

rjmateus:
@nkrinner I updated the PR to use the latest version of the provider (0.6.14), which should solve this problem according to the release notes [1].
Now I would say that we can merge as soon as QA gives the OK to merge it. I can prepare the release e-mail with the steps for all stakeholders to get a working sumaform with this changes.

[1] https://github.com/dmacvicar/terraform-provider-libvirt/releases/tag/v0.6.14

nodeg:
Since 4.3 FCS is coming soon, we can give this another try.
I found https://github.com/dmacvicar/terraform-provider-libvirt/issues/939 which seems a regression of `0.6.14`. I have not tried it myself but if we find the same issues we should switch to `0.6.13` instead of `0.6.14`.

srbarrios:
> Since 4.3 FCS is coming soon, we can give this another try. I found [dmacvicar/terraform-provider-libvirt#939](https://github.com/dmacvicar/terraform-provider-libvirt/issues/939) which seems a regression of `0.6.14`. I have not tried it myself but if we find the same issues we should switch to `0.6.13` instead of `0.6.14`.

After FCS 👍🏼

aaannz (approved):

rjmateus:
@nodeg CI PR is updated. We can merge it. I would recommend do it on monday, so we can react on any problem.

nodeg:
Thanks @rjmateus Sorry, I missed this notification until now. I just saw we need the updated package in `systemsmanagement/sumaform` in the OBS. There is still the [old 0.6.3 version there](https://build.opensuse.org/package/show/systemsmanagement:sumaform/terraform-provider-libvirt).

nodeg:
Ok so in our QE retro we discussed that we have to wait until the submissions of 4.3.1 before merging.

srbarrios:
> Ok so in our QE retro we discussed that we have to wait until the submissions of 4.3.1 before merging.

Feel free to merge it now. We can revert if it cause any trouble.

nodeg:
We should consider updating directly to [0.7.0](https://github.com/dmacvicar/terraform-provider-libvirt/releases/tag/v0.7.0) which was released in October.

rjmateus:
agree with @nodeg  we should update to the latest. After merge we should solve the problems and not revert it

nodeg:
Thank you @rjmateus  I agree, after merging we should solve potential issues and not just revert. If you need help, let me know.

srbarrios:
> agree with @nodeg  we should update to the latest. After merge we should solve the problems and not revert it

Sure. But please be sure you reserve time to fix it with urgency if it fails, avoid merging if you will not have time or you will be away.

mcalmer:
But I would suggest first try to run this with the on-PR testsuite. You can change nearly everything in that job and point to branches of susemanager-ci, terraform, sumaform, etc. You should be able to give it a try there first.

ktsamis:
Hey just a reminder that this needs some adjusting and testing again on PR testsuites. It would be the right time to merge these changes in the year otherwise we might delay it until next year again

PR #1029, opened 2022-01-24
Create module for build hosts

Bischoff:
## What does this PR change?

We don't need container module nor development tools module on normal minions. Worse, these modules are now creating problems during client migration.

Create a module of its own for build hosts.

## Links

Fixes SUSE/spacewalk#16770 .
See also SUSE/susemanager-ci#443

### DIFF
--- a/salt/repos/build_host.sls
+++ b/salt/repos/build_host.sls
@@ -1,0 +1,61 @@
+{% if 'build_host' in grains.get('roles') and grains.get('testsuite') | default(false, true) and grains['osfullname'] == 'SLES' %}
+
+{% if '12' in grains['osrelease'] %}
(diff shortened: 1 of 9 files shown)


Bischoff:
Tested to work.

maximenoel8 (approved):
LGTM

salt/repos/build_host.sls
```diff
@@ -0,0 +1,61 @@
+{% if 'build_host' in grains.get('roles') and grains.get('testsuite') | default(false, true) and grains['osfullname'] == 'SLES' %}
```
rjmateus, line 1: I don't we should only apply this if it's a testsuite, since we may want to create a build host outside the testsuite.

PR #1032, opened 2022-01-25
Not add moving target repos in 4.3-beta for server and proxy

srbarrios:
## What does this PR change?

Trying to deploy server and proxy using product version 4.3-beta, I have two issues:

1)
```
[1mmodule.proxy.module.proxy.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    Detected conflicting IDs, SLS IDs need to be globally unique.
```
Fixing this by renaming the salt state.

2)
```
[1mmodule.server.module.server.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    Rendering SLS 'base:repos.server' failed: while constructing a mapping
[1mmodule.server.module.server.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m  in "<unicode string>", line 10, column 1:
[1mmodule.server.module.server.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    server_product_pool_repo:
[1mmodule.server.module.server.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    ^
[1mmodule.server.module.server.module.host.null_resource.provisioning[0] (remote-exec):[0m [0mfound conflicting ID 'os_movingtarget_repo'
``` 
Fixing this by moving the movingtarget repos from a general statement in 4.3, to only available if we are not in beta.

PR #1033, opened 2022-01-26
Fix 43-beta os movingtarget repo

srbarrios:
## What does this PR change?

Fix 43beta os movingtarget repo


juliogonzalez (approved):
Looks OK, but I wonder why we didn't need a similar if for all the other repositories, such as server applications, that are required for proxy/server

srbarrios:
> Looks OK, but I wonder why we didn't need a similar if for all the other repositories, such as server applications, that are required for proxy/server

We need, I fixed it already in a previous PR https://github.com/uyuni-project/sumaform/pull/1032

PR #1035, opened 2022-02-07
Use SLE15SP4 Beta3 images, as Beta2 images are gone

juliogonzalez:
## What does this PR change?

Use SLE15SP4 Beta3 images, as Beta2 images are gone


juliogonzalez:
I tested the URLs with curl, and they work fine. Merging.

PR #1036, opened 2022-02-07
Do not run update-ca-certificates on SLE15

jordimassaguerpla:
## What does this PR change?

because there is a systemd unit that watches for changes and runs it
already:
   /usr/lib/systemd/system/ca-certificates.path

If we try to run it, it fails with an error because it runs twice
at the same time.


Bischoff (approved):
I would just remove lines 10 and 11, no need to be that verbose imho

PR #1037, opened 2022-02-07
Add SLE Micro 5.1 support

aaannz:
## What does this PR change?

Adds link for SLE Micro 5.1 build at [sumaform:images:microos project](https://build.opensuse.org/project/show/systemsmanagement:sumaform:images:microos)

MicroOS, being a transactional system, requires salt to skip creating new snapshots. Added `direct_call` module executor ensures that.

### DIFF
--- a/backend_modules/libvirt/base/main.tf
+++ b/backend_modules/libvirt/base/main.tf
@@ -34,7 +34,7 @@ locals {
     opensuse152-ci-pr        = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse152-ci-pr.x86_64.qcow2"
     opensuse153-ci-pr        = "http://minima-mirror.mgr.prv.suse.net/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse153-ci-pr.x86_64.qcow2"
     opensuse153-ci-pr-client = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse153-ci-pr-client.x86_64.qcow2"
-    slemicro51o     = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/microos/images_51/SUSE-MicroOS.x86_64-sumaform.qcow2"
+    slemicro51-ign  = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/microos/images_51/SUSE-MicroOS.x86_64-sumaform.qcow2"
   }
   pool               = lookup(var.provider_settings, "pool", "default")
   network_name       = lookup(var.provider_settings, "network_name", "default")
--- a/backend_modules/libvirt/base/main.tf
+++ b/backend_modules/libvirt/base/main.tf
@@ -34,6 +34,7 @@ locals {
     opensuse152-ci-pr        = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse152-ci-pr.x86_64.qcow2"
     opensuse153-ci-pr        = "http://minima-mirror.mgr.prv.suse.net/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse153-ci-pr.x86_64.qcow2"
     opensuse153-ci-pr-client = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse153-ci-pr-client.x86_64.qcow2"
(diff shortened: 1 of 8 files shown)


aaannz:
@meaksh I added `--module-executors='[direct_call]'` unconditionally. In my tests it works, but do you see any problems with that?

backend_modules/libvirt/base/main.tf
```diff
@@ -34,6 +34,7 @@ locals {
     opensuse152-ci-pr        = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse152-ci-pr.x86_64.qcow2"
     opensuse153-ci-pr        = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse153-ci-pr.x86_64.qcow2"
     opensuse153-ci-pr-client = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse153-ci-pr-client.x86_64.qcow2"
+    slemicro51o     = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/microos/images_51/SUSE-MicroOS.x86_64-sumaform.qcow2"
```
rjmateus, line 37: Please name the variable as `slemicro51`, without the `o`.
We are using the image variable suffix `o` to represent official images. The one's packaged should not this suffix.

rjmateus (changes requested):
thanks for the great contribution.
Just one nitpick

backend_modules/libvirt/base/main.tf
aaannz: Are you sure it is for official?
According to this [backend_modules/libvirt/host/main.tf#L23](https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/host/main.tf#L23) depending on o suffix it enables cloud_init disk. Or am I reading it wrong?

rjmateus: The goal was for images that are the official ones we need to apply cloud-init to bring them to a state that sumaform are able to work with. The one's package by us should be provided with the needed data during the package and no cloud-init is needed.
In this case, do you need to run anything on cloud-init for those images?

aaannz: Just basics - network config and user password. I can do that in kiwi image as well and remove cloud init. My preference was to do the least in kiwi and rest in unified way in sumaform.

rjmateus: ok, that makes sense to me. what are the kiwi doing?
Do you think we should adapt the if statement on the cloud-init to also include images starting with `slesmicro`.
Maybe in future we could use the official ones and had the work partially done

aaannz: Removed cloudinit completely and added ignition for the root password. Rest works out of the box. Image still cannot be one from upstream because it is missing qemu-ga.

rjmateus (approved):

rjmateus:
LGTM, thanks for all the hard work on this one

aaannz:
While working on unrelated L3 I discovered that unconditionally passing direct call executor does not fly with SLE11: `salt-call: error: no such option: --module-executors`

admd:
@aaannz what are our options here to make it work for testsuit integration?

aaannz:
I do not know how to solve this on salt level. IIUC salt transactional executor is masking quite a number of salt calls and it is not supported to have different executors per individual states.

I am working on combustion support for libvirt. Once that is finished we will have an option to template combustion script and install all needed package there. Salt installation would then be noops unless version changes.

Another option is simply to preinstall everything needed for test suite minion in kiwi file. That I can do straight away, but lacks flexibility.

rjmateus:
@aaannz Could be an option to check the OS version in the bash script and only run with the `--module-executors` for sles micro, or sles >= then 5?

admd:
> Another option is simply to preinstall everything needed for test suite minion in kiwi file. That I can do straight away, but lacks flexibility.

@aaannz let's do this^^ for now. Less flexible, sure, but at least we will start testing the core functionality.

aaannz:
- Changed first highstate run to force direct executors only for SLE Micro based machines.
- Added test suite packages to the kiwi file so salt states are expected pass without trying to install anything. I did not yet run this in testsuite and I do not expect to have time to add it to test suite anytime soon, but I will help with anything missing.

- During search for packages to install I also noticed some old 4.0 setups and decided to remove it here.

salt/repos/server.sls
```diff
@@ -1,57 +1,5 @@
 {% if 'server' in grains.get('roles') %}
 
-{% if '4.0' in grains['product_version'] and not grains.get('server_registration_code') %}
```
srbarrios, line 3: Please don't remove this!
Keep in mind that still have a card to automate tests for server upgrade from 4.0 to 4.x
https://github.com/SUSE/spacewalk/issues/11924

aaannz: Ok, no problem. But 4.0 is EOL since 30 Jun 2021. I think that card should be closed now.

srbarrios:
Somehow related, notice that we have this card https://github.com/SUSE/spacewalk/issues/16314 as some dependencies are not included in SLE Micro 5.1, so some features are not available.

rjmateus:
LGMT @aaannz 
I think we can merge it and try it on the fields. Thanks for the great contribution

PR #1038, opened 2022-02-08
change pg_hba.conf when external connections should not be allowed

mcalmer:
## What does this PR change?

When external connections to the postgresql DB should no be allowed, do not configure pg_hba.conf to do so.


rjmateus (approved):
lgtm

PR #1040, opened 2022-02-08
Enable autorefresh in all the repository definitions added with salt

srbarrios:
## What does this PR change?

Enable autorefresh in all the repository definitions added with salt.

This PR try to prevent an issue during deployments and even during our test suite.
Once we do a `zypper refresh`, we obtain a metadata file from the repositories. If by bad luck the repository gets updated after this `zypper refresh`, our metadata will be old and invalid, so the packages we try to download will not be available. This autorefresh in the repository definition will force zypper to always run `zypper install`, having before a `zypper refresh`.

WARN:
This change could very well cause issues, even if we hope to prevent one issue.
Probably the best approach is to after deployment, disabled this parameter in all the repositories again, but that could be kind of tough :confused:

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -25,6 +25,7 @@ custom_avahi_repo:
     {% elif grains['osrelease'] == '15.3' %}
     - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.7/SLE_15_SP3/
     {% endif %}
+    - refresh: True
     - priority: 95
     - gpgcheck: 0
 {% endif %}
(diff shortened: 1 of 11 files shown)


juliogonzalez (approved):
Should be OK if you changed all the repos :-)

But wait until @mcalmer reviews, just in case I am missing something that could break because of this change.

mcalmer (approved):
Let' give it a try

Bischoff (commented):
Could we make sure it happens only on head? Or does this PR ensure it is the case? I did not check the `{% if`s as jinja is spaghetti-code

My concern is that we start not testing like customers. So better keep it in head only. Or just introduce a mirror in NUE.

srbarrios:
I made this change for all the repositories injected using a salt state. As we experienced this issue in all the product versions.

jordimassaguerpla (approved):

jordimassaguerpla:
Good idea! :+1:

PR #1043, opened 2022-02-09
Remove ifs that are not necessary anymore

Bischoff:
## What does this PR change?

older salt had problems with `unless`

with recent salt, those `{% if`s are not needed anymore

### DIFF
--- a/salt/minion/testsuite.sls
+++ b/salt/minion/testsuite.sls
@@ -36,15 +36,12 @@ update_ca_truststore:
       - file: suse_certificate
     - require:
       - pkg: suse_minion_cucumber_requisites
-{%- if grains['saltversioninfo'][0] >= 3002 %} # Workaround for bsc#1188641
     - unless:
       - fun: service.status
         args:
           - ca-certificates.path
-{%- endif %}
 
 {% endif %}
-
 {% endif %}
 
 {% endif %}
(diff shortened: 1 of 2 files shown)


salt/minion/testsuite.sls
```diff
@@ -36,15 +36,12 @@ update_ca_truststore:
       - file: suse_certificate
     - require:
       - pkg: suse_minion_cucumber_requisites
-{%- if grains['saltversioninfo'][0] >= 3002 %} # Workaround for bsc#1188641
```
srbarrios, line 39: Are we sure this is not necessary anymore in any of our product_version deployments?
I mean, if we deploy 4.1, 4.2, Head, it will work for any of them?

cbbayburt: IIRC, the oldest Salt we support was 3000 at the time. With the fix to bsc#1188641, `unless` is supported all the way to 3000, so I think the answer is yes, but someone from Ion better confirm this.

cbbayburt (approved):

cbbayburt:
@Bischoff could you please apply the same to `certs/suse.sls` while you're at it?

salt/minion/testsuite.sls
Bischoff: "in worst case we revert"...

Bischoff:
> @Bischoff could you please apply the same to `certs/suse.sls` while you're at it?

sure, will do in a separate PR

PR #1044, opened 2022-02-11
add minion role to build host machine

rjmateus:

## What does this PR change?

The new module build host module was a spin-off from the minion module. 
I think we should also add the minion role to allow, for example, auto-register the machine to the suse manager server.
Cucumber testsuite module disable the auto registration, so I'm expecting the same behaviour we already have on the CI.


Bischoff:
I considered doing this, but I did not, because I was uncertain of the consequences. Approved but still not sure about what will happen.

Bischoff (approved):

srbarrios:
@rjmateus two years old PR, should we close or merge?

PR #1058, opened 2022-03-03
New parameter to allow not installing any kind of salt package

srbarrios:
## What does this PR change?

In order to test regular salt minion, we expect no salt package installed in the host where we will bootstrap the salt minion.

Related to card: https://github.com/SUSE/spacewalk/issues/16687


srbarrios:
Closing. Does not make sense.

PR #1059, opened 2022-03-04
Add server_mounted_mirror to the testsuite server

jordimassaguerpla:
This parameter is useful to "skip" SCC and instead get the info and
packages from a mirror.

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>

## What does this PR change?

**add description**


jordimassaguerpla:
@srbarrios I need this to use minima-mirror as a mirror for SCC. With this, we can mount with nfs minima-mirror.mgr.prv.suse.net/srv/mirror ... and it works very good! This way it is faster and, more importantly, it is stable. I am doing this because I was having timeouts with scc in the deploy phase.

srbarrios (approved):

srbarrios:
Looks good to me, but better if @rjmateus approve too.

PR #1062, opened 2022-03-08
Fix products version for Uyuni to force Salt bundle usage

meaksh:
## What does this PR change?

This PR fixes a problem detected on Uyuni Master testsuite caused by a typo on the list of products version where to force the usage of Salt Bundle


ktsamis (approved):
Thank you for the quick fix :100:

PR #1064, opened 2022-03-12
images and other projects were moved from moio to uyuni-project

mcalmer:
## What does this PR change?

Use new project names as things were moved from `moio` to `uyuni-project`


mcalmer:
@moio @Bischoff @ktsamis @nodeg - JFYI

PR #1066, opened 2022-03-16
initial install for containerized proxy

rjmateus:

## What does this PR change?

install needed packages for containerized proxy

### DIFF
--- a/modules/proxy/variables.tf
+++ b/modules/proxy/variables.tf
@@ -148,3 +148,8 @@ variable "accept_all_ssl_protocols" {
   description = "Turn to true to force Apache to accept a greater range of protocol versions"
   default     = false
 }
+
+variable "proxy_containerized" {
+  description = "Proxy should be prepared to be a containerized proxy"
+  default     = false
+}
--- a/salt/proxy/init.sls
+++ b/salt/proxy/init.sls
@@ -6,6 +6,12 @@ include:
   - minion
 {% endif %}
 
+{% set install_proxy_container_packages = false %}
+{% if grains.get('proxy_containerized') | default(false, true) or grains.get('testsuite') | default(false, true)%}
+{% if 'head' in grains.get('product_version') or 'uyuni-master' in grains.get('product_version') %}
+    {% set install_proxy_container_packages = true %}
+{% endif %}
+{% endif %}
 
 proxy-packages:
   pkg.installed:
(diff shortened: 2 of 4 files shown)


salt/repos/proxy.sls
```diff
@@ -291,6 +291,39 @@ testing_overlay_devel_repo:
...
+    {% set sle_version_path = '15' %}
+    {% elif grains['osrelease'] == '15.1' %}
+    {% set sle_version_path = '15-SP1' %}
+    {% elif grains['osrelease'] == '15.2' %}
+    {% set sle_version_path = '15-SP2' %}
+    {% elif grains['osrelease'] == '15.3' %}
+    {% set sle_version_path = '15-SP3' %}
+    {% elif grains['osrelease'] == '15.4' %}
+    {% set sle_version_path = '15-SP4' %}
+    {% endif %}
```
srbarrios, line 307: What about openSUSE for Uyuni?

salt/proxy/init.sls
```diff
@@ -20,6 +20,11 @@ proxy-packages:
       - firewalld
 {% else %}
       - SuSEfirewall2
+{% endif %}
+{% if grains.get('proxy_containerized') | default('false', true) %}
+      - podman
+      - unzip
+      - uyuni-proxy-systemd-services
```
srbarrios, line 27: I have a doubt about `uyuni-proxy-systemd-services`, once it is installed, the service is enabled and started automatically?
In the tests, we expect that the service is not started automatically, as first we need to switch ssh default port of the VM.

srbarrios:
As initial PR, looks good.
But I guess we will need to also support the possibility to have the services enabled and started, not only have the packages installed.
And for that, I think better we handle with two parameters.
- So we have the possibility in our tests to have all necessary packages installed, but the service stopped.
- We have the possibility in demos or developments to have the containerized proxy service enabled and started, and for that we need to also change the default ssh port of the VM, as I do in the tests:
  ```
  sed -i.bak -Ee 's/^#?Port\s+[0-9]*$/Port #{port}/g;' /etc/ssh/sshd_config
  systemctl restart sshd
end
  ```

modules/proxy/variables.tf
```diff
@@ -148,3 +148,8 @@ variable "accept_all_ssl_protocols" {
   description = "Turn to true to force Apache to accept a greater range of protocol versions"
   default     = false
 }
+
+variable "proxy_containerized" {
+  description = "Proxy should be prepared to be a containerized proxy"
+  default     = false
+}
```
srbarrios, line 155: New line I guess

salt/repos/proxy.sls
rjmateus: I think podmand is in the main channel. If not we need to include it. I will test it, but for now you are unblocked to test head

salt/proxy/init.sls
rjmateus: I tested and after install it's not started

rjmateus:
I need to think about the auto configuration, since we need to get the server up and running, to get the configuration zip file. I would let it automate for latter, when we have a clear picture for it

PR #1072, opened 2022-03-31
modify vendor files also when additional_repos_only is set

jordimassaguerpla:
## What does this PR change?

Otherwise, we cannot add additional_repos via terraform to test, for
example, salt packages under development.


rjmateus (approved):
LGTM

PR #1073, opened 2022-04-04
download CA cert and key from a hub when setting up a peripheral server

mcalmer:
## What does this PR change?

A Hub environment require 1 Root CA for the hub and all peripheral servers to make the reporting feature work.
Download CA cert and key from "server" and re-use it when generating the SSL server certificates

### DIFF
--- a/modules/server/variables.tf
+++ b/modules/server/variables.tf
@@ -53,7 +53,7 @@ variable "auto_register" {
 }
 
 variable "download_private_ssl_key" {
-  description = "copy SSL certificates from the server upon deployment"
+  description = "copy SSL certificates from the hub server upon deployment"
   default     = true
 }
 
--- a/modules/server/variables.tf
+++ b/modules/server/variables.tf
@@ -52,6 +52,11 @@ variable "auto_register" {
   default     = true
 }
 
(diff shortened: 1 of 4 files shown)


salt/server/init.sls
```diff
@@ -27,6 +27,39 @@ server_packages:
       - sls: repos
       - sls: server.firewall
 
+{% if grains.get('download_private_ssl_key') and grains.get('server') %}
```
rjmateus, line 30: I think here we should also check if the role `minion` is also on the grains, since that will affect if the server will be registered to a hub.

modules/server/variables.tf
```diff
@@ -52,6 +52,11 @@ variable "auto_register" {
   default     = true
 }
 
+variable "download_private_ssl_key" {
+  description = "copy SSL certificates from the server upon deployment"
```
rjmateus, line 56:
```suggestion
  description = "copy SSL certificates from the hub server upon deployment"
```

rjmateus (commented):
just one question

salt/server/init.sls
mcalmer: ok, changed

PR #1074, opened 2022-04-04
Devel parameter: extend login timeout

ncounter:
## tl;dr
At _sumaforming time_ set the webUI `login_timeout` as desired.

## Verbose reason

At development time you don't really want to re-login again during the day while you are going back and forth fir a bit more than just 1 hour. This _optional_ parameters overrides the default one by editing the `/etc/rhn/rhn.conf` file in the same way the normal user should in the real life to change that value.

### Review

Note: review asked to some random developer


rjmateus (approved):
LGTM

PR #1078, opened 2022-04-07
Rollback simplification of 4.1/4.2-released

juliogonzalez:
## What does this PR change?

Nigthlies can be unified, but not `released`, as Uyuni "Stable" is also `uyuni-released`


salt/repos/default.sls
```diff
@@ -636,7 +643,7 @@ tools_update_repo_raised_priority:
             Package: *
             Pin: release l=systemsmanagement:Uyuni:Master:Ubuntu{{ short_release }}-Uyuni-Client-Tools
             Pin-Priority: 800
-{% elif 'released' in grains.get('product_version') | default('', true) %}
+{% elif 'uyuni-released' in grains.get('product_version') | default('', true) %}
```
juliogonzalez, line 646: This will make more evident we are talking about Uyuni here

srbarrios (approved):

srbarrios:
Looks good, but I'm not good catching issues in SLS files :)

juliogonzalez:
> Looks good, but I'm not good catching issues in SLS files :)

Nobody is, I am afraid. But at least you check the logic :-)

PR #1082, opened 2022-04-07
Add possitibility to specify an ami image for AWS

maximenoel8:
## What does this PR change?

Testing build images is currently done manually. 
This PR is part of an automation effort to test build imagefor AWS.
This PR adds the possibility to specify an image ami id for server and proxy in sumaform.

There are three main features :  
 - adding the possibility to declare an ami id in server and proxy host
 - adding the possibility to deploy without declaring repositories
 - adding a new ENV variable for controller to detect when we are using build image ( mandatory for running the test suite)

### DIFF
--- a/modules/controller/variables.tf
+++ b/modules/controller/variables.tf
@@ -513,7 +513,7 @@ variable "no_mirror" {
   default     = false
 }
 
-variable "server_build_image" {
+variable "is_using_build_image" {
 description = "Specify to controller that server image is a build image"
 default     = false
 }
--- a/modules/controller/variables.tf
+++ b/modules/controller/variables.tf
@@ -512,3 +512,8 @@ variable "no_mirror" {
   description = "Ignore mirror even if base configuration has one set"
   default     = false
 }
+
+variable "server_build_image" {
+description = "Specify to controller that server image is a build image"
+default     = false
+}
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -20,6 +20,7 @@ module "host" {
   connect_to_base_network       = var.connect_to_base_network
   connect_to_additional_network = var.connect_to_additional_network
   image                         = var.image
+  ami_image                     = var.ami_image
   provision                     = var.provision
   provider_settings             = var.provider_settings
   additional_disk_size          = var.additional_disk_size
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -20,7 +20,6 @@ module "host" {
   connect_to_base_network       = var.connect_to_base_network
   connect_to_additional_network = var.connect_to_additional_network
   image                         = var.image
-  ami_image                     = var.ami_image
   provision                     = var.provision
   provider_settings             = var.provider_settings
   additional_disk_size          = var.additional_disk_size
--- a/modules/proxy/main.tf
+++ b/modules/proxy/main.tf
@@ -58,7 +58,6 @@ module "proxy" {
   }
 
   image                    = var.image == "default" || var.product_version == "head" ? var.images[var.product_version] : var.image
-  ami_image                = var.ami_image
   provider_settings        = var.provider_settings
   additional_disk_size     = var.repository_disk_size
   volume_provider_settings = var.volume_provider_settings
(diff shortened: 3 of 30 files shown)


salt/server/testsuite.sls
```diff
@@ -74,8 +72,6 @@ testsuite_salt_packages:
 {% if 'head' in grains.get('product_version') or 'uyuni-master' in grains.get('product_version') or 'nightly' in grains.get('product_version') %}
     - fromrepo: testing_overlay_devel_repo
 {% endif %}
-    - require:
-      - sls: repos
```
srbarrios, line 78: About all the "require" sections removed.
I think we should not remove it. One thing is to include repos in the init.sls, the other to let the salt state knows that it has a dependency with repos.sls

modules/proxy/variables.tf
```diff
@@ -152,4 +152,9 @@ variable "accept_all_ssl_protocols" {
...
   description = "Proxy should be prepared to be a containerized proxy"
   default     = false
-}
\ No newline at end of file
+}
+
+variable "ami_image" {
+  description = "An ami image to overwrite the default value"
+  default     = null
+}
```
srbarrios, line 160: To be consisten with the definition in the null/host module, maybe we should have here:
```
variable "ami_image" {
  description = "An ami image to overwrite the default value"
  type        = string
  default     = null
}
```

modules/controller/variables.tf
```diff
@@ -512,3 +512,8 @@ variable "no_mirror" {
   description = "Ignore mirror even if base configuration has one set"
   default     = false
 }
+
+variable "server_build_image" {
```
srbarrios, line 516: Personally I would choose a variable name that let us think this is a boolean.
Something like... "server_use_build_image"

salt/controller/bashrc
```diff
@@ -78,6 +78,7 @@ export AUTH_REGISTRY_CREDENTIALS="{{ grains.get('auth_registry_username') }}|{{
 {% if grains.get('server_http_proxy') | default(false, true) %}export SERVER_HTTP_PROXY="{{ grains.get('server_http_proxy') }}" {% else %}# no server HTTP proxy defined {% endif %}
 {% if grains.get('pxeboot_image') | default(false, true) %}export PXEBOOT_IMAGE="{{ grains.get('pxeboot_image') }}" {% else %}# no PXE boot image defined {% endif %}
 export PROVIDER="{{ grains.get('provider') }}"
+export SERVER_BUILD_IMAGE="{{ grains.get('server_build_image') }}"
```
Bischoff, line 81: Is it always defined ?

Bischoff (commented):
just a question

Bischoff (changes requested):
Isn't this PR missing changes to README_ADVANCED.md?

modules/host/main.tf
```diff
@@ -20,6 +20,7 @@ module "host" {
   connect_to_base_network       = var.connect_to_base_network
   connect_to_additional_network = var.connect_to_additional_network
   image                         = var.image
+  ami_image                     = var.ami_image
```
rjmateus, line 23: I don't think we need this, as explained above

modules/proxy/main.tf
```diff
@@ -57,6 +58,7 @@ module "proxy" {
   }
 
   image                    = var.image == "default" || var.product_version == "head" ? var.images[var.product_version] : var.image
+  ami_image                = var.ami_image
```
rjmateus, line 61: we don't need this variable

modules/server/main.tf
```diff
@@ -82,6 +83,7 @@ module "server" {
 
 
   image                    = var.image == "default" || var.product_version == "head" ? var.images[var.product_version] : var.image
+  ami_image                = var.ami_image
```
rjmateus, line 86: we don't need this variable

modules/server/variables.tf
```diff
@@ -269,3 +269,8 @@ variable "login_timeout" {
   description = "How long the webUI login session cookie is valid"
   default     = null
 }
+
+variable "ami_image" {
```
rjmateus, line 273: we don't need this variable

rjmateus (changes requested):
some changes that I think should be done

salt/server/testsuite.sls
```diff
@@ -76,6 +81,7 @@ testsuite_salt_packages:
 {% endif %}
     - require:
       - sls: repos
+{% endif %} {# 'build_image' not in grains.get('product_version') | default('', true) #}
```
srbarrios, line 84: I don't get why we need this comment.

PR #1086, opened 2022-04-20
Why would we install the bundle on a trad client?

Bischoff:
## What does this PR change?

Small doc fix: I don't think it makes sense to install the bundle on a traditional client.


meaksh:
@Bischoff we need the Salt bundle preinstalled in the client as this is required for the migration from traditional client to minion that we run in our testsuite scenarios in order for them to work using the Salt bundle.

Bischoff:
ok, closing then

PR #1090, opened 2022-05-06
lock package to allow deployment of prometheus in sles15sp3 and newer

rjmateus:

## What does this PR change?

Resolution for: https://github.com/SUSE/spacewalk/issues/17512

In the nightly build installations, we add a devel repo for client tools for each minion. This channel has a higher priority than all others, to make sure we install packages from this channel.
When we install the package `golang-github-prometheus-prometheus` it has as a dependency the package `firewalld-prometheus-config`.
On sles15SP3 this package is obsoleted by the package `firewalld`, provided in the OS base channel.
Since the devel repo has a higher priority, it will ignore this relationship and installation fail with a package conflict.

This PR solves this by locking the package `firewalld-prometheus-config` on sles15SP3 and higher, which solves the conflict problem.

I have tested it locally and the lock mechanism solves the problem.


nodeg:
Thank you!

PR #1091, opened 2022-05-09
add iptables package to sshminion to test jumphost functionality

mcalmer:
## What does this PR change?

The testsuite got a feature to test Proxy as jumphost. This try to forbid direct access from server to sshminion using iptables
firewall rule. For this we need to have iptables installed.


Etheryte:
@mcalmer Is this still relevant?

mcalmer:
I think not

PR #1100, opened 2022-05-24
define http method for tool repository

rjmateus:
Signed-off-by: Ricardo Mateus <rmateus@suse.com>

## What does this PR change?

follow-up PR of: https://github.com/uyuni-project/sumaform/pull/1099

Define http in the download URL.

PR #1101, opened 2022-05-24
Proxy repo recursive

rjmateus:
## What does this PR change?

follow-up of https://github.com/uyuni-project/sumaform/pull/1100

Correct syntax
Remove requires, which was forcing a circular dependency because of the repository name.


ktsamis:
Thanks for this, I'm restarting the uyuni and head testsuites

PR #1109, opened 2022-05-30
Disable JMX ports per default

renner:
## What does this PR change?

This patch would disable remote monitoring via Java Management Extensions (JMX) per default for development servers, while allowing to easily enable it from the config file.


aaannz:
Is this really needed? First line of those files is check for java_debuging grain. So IMO we should just by default either set that grain to false or not set it at all.

mcalmer (approved):

mcalmer:
@aaannz The goal is to disable JMX always, but have an easy way to know which options are needed to enable it again. sumaform always set debug on, but even then we do not need jmx

PR #1110, opened 2022-06-03
Change inheritance of pool when no variable is set

ktsamis:
## What does this PR change?

Change inheritance of pool when no variable is set. From using the default pool to using the pool that is defined in the base module.


Bischoff (approved):
yeah, was an inheritance problem

rjmateus (approved):

PR #1112, opened 2022-06-03
change tools repo for RHEL based os to use EL8

rjmateus:

## What does this PR change?

Centos tools repository are out of support and we should use EL8 tool for all rhel based systems from now on.


rjmateus:
I have made changes according to your comment.
I also changed the cloud-init repo

PR #1117, opened 2022-06-20
set trace to log level for tests

jordimassaguerpla:
## What does this PR change?

This way we can debug and do not need to rerun the pipelines/tests.


jordimassaguerpla:
I think some tests fail because of this. Sorry for the noise.

PR #1119, opened 2022-06-21
remove fqdn from /etc/hosts

rjmateus:

## What does this PR change?

remove fqdn from /etc/hosts added by cloud-init which was causing problems when running debian on Build Validation testsuite.

### DIFF
--- a/salt/default/network.sls
+++ b/salt/default/network.sls
@@ -75,3 +75,19 @@ mdns_iptables:
     - protocol: udp
     - save: True
 {% endif %}
+
+{% if grains['os'] == 'Debian' %}
+comment_hosts_fqdn:
+  file.replace:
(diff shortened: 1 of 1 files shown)


salt/default/network.sls
```diff
@@ -75,3 +75,20 @@ mdns_iptables:
...
+    - count: 1
+
+cloud_init_disable_manage_hosts:
+  file.replace:
+    - name: /etc/cloud/cloud.cfg
+    - pattern: "^manage_etc_hosts: true$"
+    - repl: "manage_etc_hosts: false"
+    - append_if_not_found: true
+
+{% endif %}
```
Bischoff, line 94: add linefeed at the end of last line

rjmateus: done

PR #1120, opened 2022-06-27
Add harvester provider support

aaannz:
## What does this PR change?

Inspired by libvirt provider adapted to harvester resources.

~~Until https://github.com/harvester/harvester/issues/2255 is fixed does not work in one go and after initial provisioning terraform needs to reapply to provision salt part.~~ Fixed and released as v0.5.0

### DIFF
--- a/backend_modules/harvester/base/main.tf
+++ b/backend_modules/harvester/base/main.tf
@@ -37,7 +37,6 @@ locals {
     slemicro51-ign  = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/microos/images_51/SUSE-MicroOS.x86_64-sumaform.qcow2"
     slemicro51o-ign = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/install/SLE-Micro-5.1-GM/SUSE-MicroOS.x86_64-5.1.0-Default-GM.raw.xz"
   }
-  pool               = lookup(var.provider_settings, "pool", "default")
   network_name       = lookup(var.provider_settings, "network_name", "default")
   bridge             = lookup(var.provider_settings, "bridge", null)
   additional_network = lookup(var.provider_settings, "additional_network", null)
@@ -47,11 +46,11 @@ locals {
 resource "harvester_image" "images" {
   for_each = local.images_used
 
-  name   = "${var.name_prefix}${each.value}"
-  display_name   = "${var.name_prefix}${each.value}"
-  namespace = local.namespace
-  source_type = "download"
-  url = local.image_urls[each.value]
+  name         = "${var.name_prefix}${each.value}"
+  display_name = "${var.name_prefix}${each.value}"
+  namespace    = local.namespace
+  source_type  = "download"
+  url          = local.image_urls[each.value]
 }
 
 resource "harvester_network" "additional_network" {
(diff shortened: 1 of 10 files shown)


aaannz:
Ready to review, thing to consider is that this duplicates list of images where both libvirt and harvester uses the same cloud-init images (except SLE Micro). So having this list unified should be something to look at.

backend_modules/harvester/base/versions.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_version = ">= 1.1.5"
```
rjmateus, line 2: Do we really need this terraform version? Currently, sumaform is tested to work with terraform version "1.0.10"

backend_modules/harvester/host/versions.tf
```diff
@@ -0,0 +1,13 @@
+terraform {
+  required_version = ">= 1.1.5"
```
rjmateus, line 2: Same comment about terraform version as before.

backend_modules/harvester/base/main.tf
```diff
@@ -0,0 +1,78 @@
...
+    opensuse152-ci-pr        = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse152-ci-pr.x86_64.qcow2"
+    opensuse153-ci-pr        = "http://minima-mirror.mgr.prv.suse.net/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse153-ci-pr.x86_64.qcow2"
+    opensuse153-ci-pr-client = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/libvirt/images/opensuse153-ci-pr-client.x86_64.qcow2"
+    slemicro51-ign  = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.opensuse.org"}/repositories/systemsmanagement:/sumaform:/images:/microos/images_51/SUSE-MicroOS.x86_64-sumaform.qcow2"
+    slemicro51o-ign = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/install/SLE-Micro-5.1-GM/SUSE-MicroOS.x86_64-5.1.0-Default-GM.raw.xz"
+  }
+  network_name       = lookup(var.provider_settings, "network_name", "default")
+  bridge             = lookup(var.provider_settings, "bridge", null)
```
rjmateus, line 41: this variable doesn't look to be used.

rjmateus (commented):
just some small comments, overall look really good.
thanks for the great contribution

rjmateus:
> Ready to review, thing to consider is that this duplicates list of images where both libvirt and harvester uses the same cloud-init images (except SLE Micro). So having this list unified should be something to look at.

That is something we can think of, on how to share this list.
Maybe a share module which outputs the list of images.

backend_modules/harvester/base/main.tf
aaannz: Right! I will remove it.

backend_modules/harvester/base/versions.tf
rjmateus: good for me, and it should work fine in there. 
Also we should update terraform version again :)

aaannz:
I no longer have access to harvester instance and cannot continue on this. Let's close it then.

PR #1124, opened 2022-07-14
Temporary assign beta client tools for 4.3

juliogonzalez:
## What does this PR change?

Temporary assign beta client tools for 4.3, until `Manager-4.3` is fully branched and the testsuite can stop requiring beta client tools.

### DIFF
--- a/salt/server/init.sls
+++ b/salt/server/init.sls
@@ -138,7 +138,9 @@ extend_login_timeout:
         - cmd: server_setup
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


ktsamis (approved):

Bischoff (changes requested):
Could we add a `# WORKAROUND` comment, so we minimize the risk to forget removing it again ?

juliogonzalez:
@Bischoff: done

PR #1126, opened 2022-07-19
Migration test support aws

maximenoel8:
## What does this PR change?

Update sumaform deployment on AWS for SUMA 4.3 : 
- remove some SCC modules for minion for an easier migration test
- fix `install_salt_bundle` declaration in host/main.tf
- create SCC to support build_host module
- add sles15sp4 to image declaration


salt/scc/minion.sls
```diff
@@ -8,16 +8,4 @@ basesystem_activation:
    cmd.run:
      - name: SUSEConnect -p sle-module-basesystem/{{ grains['osrelease'] }}/x86_64
 
-containers_activation:
```
rjmateus, line 11: This was being applied to all minions, with this change will only be applied to build_host, is this what you want?

rjmateus (commented):
Just one question

salt/scc/minion.sls
maximenoel8: Yes, I don't want this modules activated for minion and ssh minion.
I added that change in comparison with libvirt deployment.
The migration test was not working because my minion had more modules activated than the minion on libvirt so I decided to limit the modules. It doesn't look to impact other part of the test.

srbarrios: I wonder why we even activate those in our minions. Not sure what's the test case making use of them. Any idea @rjmateus ?

modules/build_host/variables.tf
```diff
@@ -113,3 +113,8 @@ variable "provider_settings" {
   description = "Map of provider-specific settings, see the modules/libvirt/README.md"
   default     = {}
 }
+
+variable "sles_registration_code" {
+  description = "SUMA SCC registration code to enable the SLES server"
+  default     = null
+}
```
srbarrios, line 120: It might need an extra new line.

salt/scc/build_host.sls
```diff
@@ -0,0 +1,23 @@
...
+   cmd.run:
+     - name: SUSEConnect -p sle-module-basesystem/{{ grains['osrelease'] }}/x86_64
+
+containers_activation:
+   cmd.run:
+     - name: SUSEConnect -p sle-module-containers/{{ grains['osrelease'] }}/x86_64
+
+desktop_activation:
+   cmd.run:
+     - name: SUSEConnect -p sle-module-desktop-applications/{{ grains['osrelease'] }}/x86_64
```
srbarrios, line 17: Do we really use desktop module? 🤔

salt/scc/init.sls
```diff
@@ -0,0 +1,2 @@
+include:
+  - scc.minion
```
srbarrios, line 2: We might need a new line here

salt/scc/minion.sls
rjmateus: I think it was added by @maximenoel8  to speed up high state apply, so it could connect to official repositories to download updates to minion.

maximenoel8: They are not activated on libvirt minion.

PR #1127, opened 2022-07-20
Remove moving target now that 15sp4 is out

ktsamis:
## What does this PR change?

Remove moving target now that 15sp4 is out, it's causing issues in BV


rjmateus (approved):
LGTM.

PR #1130, opened 2022-07-29
Add SLE Micro 5.2

nodeg:
## What does this PR change?

This will

- add SLE Micro 5.2 support


modules/cucumber_testsuite/main.tf
```diff
@@ -115,7 +115,7 @@ module "suse-client" {
   quantity = contains(local.hosts, "suse-client") ? 1 : 0
   base_configuration = module.base.configuration
   product_version    = var.product_version
-  image              = lookup(local.images, "suse-client", "sles15sp2o")
+  image              = lookup(local.images, "suse-client", "sles15sp4o")
```
srbarrios, line 118: Better don't mix PRs.
This PR should only have changes to support SLE Micro.
Move other stuff in another PR please.

srbarrios (approved):

srbarrios:
We might need to also include one VM for SSH Salt Minion, let's wait for Abid's reply.

modules/cucumber_testsuite/main.tf
nodeg: I can do that. I used explicitly 2 commits to separate the changes.

srbarrios (approved):

PR #1131, opened 2022-07-29
Update cucumber_testsuite module images

nodeg:
## What does this PR change?

This will update the images in the cucumber_testsuite module


srbarrios (approved):

vandabarata (approved):

srbarrios:
LGTM. As for every environment in our CT, we are overriding these default values, so it will not affect.

nodeg:
Yes. No the local deployments are on par to the stuff in the CI.

PR #1135, opened 2022-08-08
replace Devel:Galaxy gpg key - new RSA 2048 Bit

mcalmer:
## What does this PR change?

We changed the internal GPG key. Add the new to sumaform


mcalmer:
@moio I have no idea why this error happenes, but it seems something internal.

moio (approved):

PR #1136, opened 2022-08-08
[AWS] Salt upgrade during cloud init

maximenoel8:
## What does this PR change?
For sles15sp3o, upgrade salt during cloud-init.

## Reason 

Salt upgrade during highstate application is crashing the deployment. This PR will upgrade salt during cloud-init using external 
repository. 

## Why this changes are not needed for sles15sp1o / sles15sp2o / sles15sp4o

Deployment doesn't crash with this versions

### DIFF
--- a/backend_modules/aws/host/user_data.yaml
+++ b/backend_modules/aws/host/user_data.yaml
@@ -61,15 +61,11 @@ packages: ["salt-minion"]
 %{ endif }
 
 %{ if image == "sles15sp3o" }
-zypper:
-  repos:
-    - id: salt_qubes_build
-      name: salt_qubes_build
-      baseurl: https://download.opensuse.org/repositories/home:/wicked:/qubes-build/SLE_15_SP3/
-      enabled: 1
-      autorefresh: 1
-
-packages: ["salt-minion"]
+runcmd:
+  - zypper addrepo -G -e http://download.opensuse.org/repositories/systemsmanagement:/saltstack:/products/SLE_15_SP3/ salt_test
+  - zypper ref
+  - zypper in --allow-vendor-change --no-confirm salt-minion
+  - zypper removerepo --all
 
 %{ endif }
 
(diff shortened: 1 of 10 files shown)


backend_modules/aws/host/user_data.yaml
```diff
@@ -61,15 +61,11 @@ packages: ["salt-minion"]
...
-      name: salt_qubes_build
-      baseurl: https://download.opensuse.org/repositories/home:/wicked:/qubes-build/SLE_15_SP3/
-      enabled: 1
-      autorefresh: 1
-
-packages: ["salt-minion"]
+runcmd:
+  - zypper addrepo -G -e http://download.opensuse.org/repositories/systemsmanagement:/saltstack:/products/SLE_15_SP3/ salt_test
+  - zypper ref
+  - zypper in --allow-vendor-change --no-confirm salt-minion
```
mcalmer, line 67: Is the GPG key trusted at this point already?

backend_modules/aws/host/user_data.yaml
```diff
@@ -61,15 +61,11 @@ packages: ["salt-minion"]
...
-zypper:
-  repos:
-    - id: salt_qubes_build
-      name: salt_qubes_build
-      baseurl: https://download.opensuse.org/repositories/home:/wicked:/qubes-build/SLE_15_SP3/
-      enabled: 1
-      autorefresh: 1
-
-packages: ["salt-minion"]
+runcmd:
```
srbarrios, line 64: I wonder we we can keep using zypper option in cloudinit, there is no way to pass those extra parameters from that option?
Using runcmd looks faster yes, but.. if cloudinit has a specific option for zypper I think better to use it.

maximenoel8: Packages doesn't work in this case because of installation conflict. That's why I added the option `--allow-vendor-change`. I tried to work with zypper repo and package but after a time I switch to that method because I have more control.

backend_modules/aws/host/user_data.yaml
mcalmer: Oh, this option was new to me :-) But this is ok

PR #1138, opened 2022-08-08
[AWS] Additional packages with scc

maximenoel8:
## What does this PR change?

Currently additional package installation will be executed before SUSEConnect registration. It will fail most of package installation if the deployment use scc method ( no internal repository / setup repositories using SUSEConnect -p).

A scc salt module refactoring was necessary to be compatible with the different modules ( server, proxy, sles minion, ...)

### DIFF
--- a/salt/scc/clean.sls
+++ b/salt/scc/clean.sls
@@ -1,0 +1,7 @@
+clean_sles_release_package:
+   cmd.run:
+     - name: rpm -e --nodeps sles-release; exit 0
+
+clean_suseconnect_registration :
+   cmd.run:
+     - name: SUSEConnect --cleanup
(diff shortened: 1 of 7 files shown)


mcalmer (approved):

salt/scc/clean.sls
```diff
@@ -0,0 +1,7 @@
+clean_sles_release_package:
+   cmd.run:
+     - name: rpm -e --nodeps sles-release; exit 0
```
srbarrios, line 3: Wouldn't be better to use https://docs.saltproject.io/en/latest/ref/states/all/salt.states.pkg.html#salt.states.pkg.removed ?

mcalmer: The --nodeps is important. Not sure if salt support it. I think for this you need lowlevel calls

PR #1145, opened 2022-08-16
Add Rocky 8 to image list

nodeg:
## What does this PR change?

Add Rocky 8 to default and cucumber module image list.


Bischoff (approved):
fine; but since rocky8 is supposed to replace centos8, would it make sense to remove centos8o ?

nodeg:
We can do that. I am not sure how we should proceed with deprecated products but I guess this is also something for hexagon. For now I would merge this and the removal of CentOS 6 and 8 can be done with another PR.

PR #1149, opened 2022-08-22
Fix ordering: install Prometheus components before starting Tomcat

moio:
## What does this PR change?

When setting up a monitored server, Tomcat will not start, Apache sending a 503 error, and several states will fail. The reason is Tomcat needs `/etc/prometheus-jmx_exporter/tomcat/java_agent.yml` to start, but that file has not yet been created as the corresponding states run later.

`journalctl -u tomcat` indeed shows:
```
Aug 22 03:58:04 suma-ref41-srv server[5940]: Caused by: java.io.FileNotFoundException: /etc/prometheus-jmx_exporter/tomcat/java_agent.yml (No such file or directory)
```

This patch fixes ordering, thereby allowing Tomcat to start correctly.

Noticed in the internal reference host: https://ci.suse.de/view/Manager/view/Manager-4.1/job/manager-4.1-infra-reference-PRV/788/console
Tested: https://ci.suse.de/view/Manager/view/Manager-4.1/job/manager-4.1-infra-reference-PRV/789/console

### DIFF
--- a/salt/server/init.sls
+++ b/salt/server/init.sls
@@ -8,11 +8,11 @@ include:
   {% if grains.get('db_configuration')['local'] %}
   - server.postgres
   {% endif %}
+  - server.prometheus
   - server.tomcat
   - server.taskomatic
   - server.spacewalk-search
   - server.rhn
-  - server.prometheus
   - server.initial_content
   - server.iss
   - server.testsuite


moio:
cc @srbarrios for awareness

rjmateus (approved):
LGTM

witekest:
I don't understand why this happens. The _systemd_ environment option `CATALINA_OPTS` is set in [jmx_tomcat_config](https://github.com/uyuni-project/sumaform/blob/master/salt/server/prometheus.sls#L63). The missing file `/etc/prometheus-jmx_exporter/tomcat/java_agent.yml` is provided just one step before.

Perhaps the steps are executed concurrently and additional `require` is needed?
Or there is some old Tomcat service configuration not cleaned up?

moio:
@witekest it's indeed weird - I did not notice initially.

You actually have two module calls under `jmx_tomcat_config` - one `module.run` and one `file.managed` - that is OK.

What is surprising is that Salt takes the liberty to split the two and execute them at different points with states in between. It could be argued that perhaps that's still OK. 

But it does so not necessarily respecting the order! In fact from the output above I can see:


```
      ID: jmx_tomcat_config
Function: file.managed
    Name: /usr/lib/systemd/system/tomcat.service.d/jmx.conf
  Result: True
 Comment: File /usr/lib/systemd/system/tomcat.service.d/jmx.conf updated
 Started: 03:57:49.690472
Duration: 29.018 ms
 Changes:
          ----------
          diff:
              New file
          mode:
              0644
```

Then:

```
          ID: tomcat_service
    Function: service.running
        Name: tomcat
      Result: True
     Comment: Service restarted
     Started: 03:57:50.967927
    Duration: 3920.368 ms
     Changes:
              ----------
              tomcat:
                  True
```

(this is expected due to the [mod_watch](https://docs.saltproject.io/en/latest/ref/states/all/salt.states.service.html#salt.states.service.mod_watch) clause [here](https://github.com/uyuni-project/sumaform/blob/master/salt/server/tomcat.sls#L34-L39))

Then several other states and...

```
          ID: jmx_exporter_tomcat_yaml_config
    Function: file.managed
        Name: /etc/prometheus-jmx_exporter/tomcat/java_agent.yml
      Result: True
     Comment: File /etc/prometheus-jmx_exporter/tomcat/java_agent.yml updated
     Started: 03:58:04.976642
    Duration: 19.615 ms
     Changes:
              ----------
              diff:
                  New file
              mode:
                  0644

          ID: jmx_tomcat_config
    Function: module.run
        Name: service.systemctl_reload
      Result: True
     Comment: Module function service.systemctl_reload executed
     Started: 03:58:04.996907
    Duration: 541.954 ms
     Changes:
              ----------
              ret:
                  True
```

And that leads to the problem.

I can now see my patch here worked around what seems like a core Salt ordering issue. In my eyes executing two calls of the same ID with other states in between [is at odds with the docs](https://docs.saltproject.io/en/getstarted/config/requisites.html), given there are no constraints on `jmx_exporter_tomcat_yaml_config`.

IOW, unless somebody can come up with a convincing explanation, I am inclined to think this could be a Salt bug.

PR #1151, opened 2022-08-23
Test suite: Replace CentOS 7 with Rocky 8

nodeg:
## What does this PR change?

This will replace CentOS 7 with Rocky 8 in the test suite module.

See https://github.com/SUSE/spacewalk/issues/12595


nodeg:
Rebase done.

srbarrios (approved):
If we change the default in the module, remember that we must change the default value by centos7o in all the main.tfs that are older than SUMA 4.3 version!

Don't merge this before that requirement or things will break :)

nodeg:
Do we? I thought we specify the image version for the SUMA versions in the terraform configuration files for each version?

srbarrios:
> Do we? I thought we specify the image version for the SUMA versions in the terraform configuration files for each version?

Right, it seems is the case. If we took the default value in some of them it will fail, but apparently they all use a specific value. 👍🏼

Bischoff (approved):

PR #1153, opened 2022-08-24
Add support for ubuntu 2204

rjmateus:

## What does this PR change?

The title says it all.

Implementation of: https://github.com/SUSE/spacewalk/issues/18280


Bischoff (commented):
about this: I am wondering if Debian 9 and 10 are not in the "ens3" case too

(would need a manual verification)

rjmateus:
> about this: I am wondering if Debian 9 and 10 are not in the "ens3" case too
> 

I was wundering the same,but looks like they are being deployed, so should be working

Bischoff:
> > about this: I am wondering if Debian 9 and 10 are not in the "ens3" case too
> 
> I was wundering the same,but looks like they are being deployed, so should be working

My bad, it's Debian 11 that has ens3

And that one does not use netplan :smile_cat:

Bischoff (approved):

PR #1160, opened 2022-09-07
Cucumber test suite module: Update used SUSE images

nodeg:
## What does this PR change?

This will update the SUSE minion and SSH minion from SLES15 SP3 to SP4 since we use SP4 in the test suite for them. The migration from SP3 to SP4 was moved to build validation so they do not need to be SP3 any longer.

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -138,7 +138,7 @@ module "suse-minion" {
   quantity = contains(local.hosts, "suse-minion") ? 1 : 0
   base_configuration = module.base.configuration
   product_version    = var.product_version
-  image              = lookup(local.images, "suse-minion", "sles15sp3o")
+  image              = lookup(local.images, "suse-minion", "sles15sp4o")
   name               = lookup(local.names, "suse-minion", "min-sles15")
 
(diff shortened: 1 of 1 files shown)


elariekerboull (approved):

Bischoff (approved):
Nice. Please notice these are default values that can be overridden from the tf file.

modules/cucumber_testsuite/main.tf
```diff
@@ -138,7 +138,7 @@ module "suse-minion" {
   quantity = contains(local.hosts, "suse-minion") ? 1 : 0
   base_configuration = module.base.configuration
   product_version    = var.product_version
-  image              = lookup(local.images, "suse-minion", "sles15sp3o")
+  image              = lookup(local.images, "suse-minion", "sles15sp4o")
```
srbarrios, line 141: Ideally it should depends on product_version set, as we have for every test suite code branch. But yeah, let's keep the latest as default.

srbarrios (approved):

nodeg:
Yes, both of you are correct. We override those with our .tf files that are used in Jenkins. But just to be up to date I wanted to change this.

PR #1162, opened 2022-09-08
aws: bastion host with static IP

rjmateus:

## What does this PR change?

On AWS bastion host now have a public static  IP (eip) assigned to it.
Before this, the bastion had a public IP but it was not static. This is a better approach to deploying the test infrastructure needed for testsuite.


rjmateus:
@mcalmer for awareness

mcalmer:
JFYI - I used it for my tests. Seems to work

PR #1173, opened 2022-09-30
Aws without os152

mcalmer:
## What does this PR change?

Seems opensuse152o was removed from AWS. Try to remove the image and use something different

### DIFF
--- a/backend_modules/aws/README.md
+++ b/backend_modules/aws/README.md
@@ -51,7 +51,7 @@ Available provider settings for the base module:
 | server_registration_code | string | `null`          | SUMA SCC server registration code to use SCC repositories and disable internal repositories                             |
 | proxy_registration_code  | string | `null`          | SUMA SCC proxy registration code to use SCC repositories and disable internal repositories                              |
 | sles_registration_code   | string | `null`          | SLE registration code to use SCC repositories and disable internal repositories ( use for minion, sshminion and client) |
-| bastion_image            | string | `opensuse152o`  | Image name to be used when deploying the bastion host                                                                   |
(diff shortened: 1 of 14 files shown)


mcalmer:
I used this to run our uyuni testsuite now in AWS.
`opensuse152o` does not exist anymore so I switched over to `opensuse154o`.
We need to see how it works. But I had to change the controller image directly.
Just overwriting in the TF file did not work and still created an error.

So when this gets merged, also libvirt installations are affected and will use 15.4 .

mcalmer:
Have a look at 
https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-master-dev-acceptance-tests-AWS/
starting at run 101 .
Looks pretty good so far.

rjmateus (approved):
LGTM. 
Thanks Michael

srbarrios:
Wouldnt be easier to use sles15sp2o ?
I mean, the transition should be smoother.

rjmateus:
controller also needs to run uyuni testsuite, so we should keep it with OpenSuse. Otherwise we can have problem when opening uyuni testsuite

srbarrios:
> controller also needs to run uyuni testsuite, so we should keep it with OpenSuse. Otherwise we can have problem when opening uyuni testsuite

Make sense.
I would prefer that we (QE squad) have control over controller image, as changing the controller version also means to change some dependencies versions for our test framework.

In any case, it start to be common to have these huge PRs not including only changes related to what it's described in the PR, but also many other changes, like changing Ubuntu version here.

Please please, let's open a new PR for every different scope we are working on @mcalmer .

Issue #1171, opened 2022-09-29
Image for SLES12 SP4 JeOS GM unavailable on download.suse.de


vandabarata:
Right now, it seems all the images for SLES12SP4 have been removed from `download.suse.de`, including the one in the title which is used by the cucumber testsuite module (https://download.suse.de/install/SLE-12-SP4-JeOS-GM/SLES12-SP4-JeOS.x86_64-12.4-OpenStack-Cloud-GM.qcow2). I'm unsure of the reason and have e-mailed the auto-build team on the 27th of September, with no reply as of yet. So, to ensure the sumaform deployments keep working, we need to use mirrors where the needed images are still hosted, or to filter for specific images for our deployments and not include this one.

juliogonzalez:
@vandabarata autobuild is short on people, so it seems they are overloaded.

Send them a reminder.

However I seem to remember someone else saw this issue before, and autobuild told the images were moved somewhere else? Maybe you want to ask at #team-susemanager

PR #1175, opened 2022-09-30
QE: Fix download link for SLES12SP4o image

vandabarata:
## What does this PR change?
The SLES12 SP4 images were removed from `download.suse.de`. So I've updated the link to `schnell.suse.de`, where they are now being hosted. New link being used is https://schnell.suse.de/SLE12/SLE-12-SP4-JeOS-GM/SLES12-SP4-JeOS.x86_64-12.4-OpenStack-Cloud-GM.qcow2
Fixes https://github.com/uyuni-project/sumaform/issues/1171

### DIFF
--- a/backend_modules/libvirt/base/main.tf
+++ b/backend_modules/libvirt/base/main.tf
@@ -25,7 +25,7 @@ locals {
     sles11sp4       = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles11sp4.x86_64.qcow2"
     sles12sp3       = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp3.x86_64.qcow2"
     sles12sp4       = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp4.x86_64.qcow2"
-    sles12sp4o      = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/install/SLE-12-SP4-JeOS-GM/SLES12-SP4-JeOS.x86_64-12.4-OpenStack-Cloud-GM.qcow2"
+    sles12sp4o      = "${var.use_mirror_images ? "http://${var.mirror}" : "http://schnell.suse.de"}/SLE12/SLE-12-SP4-JeOS-GM/SLES12-SP4-JeOS.x86_64-12.4-OpenStack-Cloud-GM.qcow2"
     sles12sp5o      = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/install/SLE-12-SP5-JeOS-GM/SLES12-SP5-JeOS.x86_64-12.5-OpenStack-Cloud-GM.qcow2"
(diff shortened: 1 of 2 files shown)


backend_modules/libvirt/base/main.tf
```diff
@@ -25,7 +25,7 @@ locals {
     sles11sp4       = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles11sp4.x86_64.qcow2"
     sles12sp3       = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp3.x86_64.qcow2"
     sles12sp4       = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/ibs/Devel:/Galaxy:/Terraform:/Images/images/sles12sp4.x86_64.qcow2"
-    sles12sp4o      = "${var.use_mirror_images ? "http://${var.mirror}" : "http://download.suse.de"}/install/SLE-12-SP4-JeOS-GM/SLES12-SP4-JeOS.x86_64-12.4-OpenStack-Cloud-GM.qcow2"
+    sles12sp4o      = "${var.use_mirror_images ? "http://${var.mirror}/install" : "http://schnell.suse.de/SLE12"}/SLE-12-SP4-JeOS-GM/SLES12-SP4-JeOS.x86_64-12.4-OpenStack-Cloud-GM.qcow2"
```
rjmateus, line 28: why the URL from mirror now uses "/install"?

vandabarata: unless I misinterpreted how this works, which is also a possibility :sweat_smile:

rjmateus: got it. in the mirror you want to keep using the image already synced

rjmateus (approved):

backend_modules/libvirt/base/main.tf
srbarrios, line 28: I think we need to do something like that:

```suggestion
    sles12sp4o      = "${var.use_mirror_images ? "http://${var.mirror}" : "http://schnell.suse.de"}/SLE12/install/SLE-12-SP4-JeOS-GM/SLES12-SP4-JeOS.x86_64-12.4-OpenStack-Cloud-GM.qcow2"
```

srbarrios: Basically, enabling the mirror option we want to use a mirror domain, and disabling it we use the origin domain. But better not mixing domains and URIs inside the domain.

srbarrios: Also, remember to update the URL in this other file:
https://github.com/uyuni-project/sumaform/blob/master/salt/mirror/etc/mirror-images.conf

And then also in:
https://gitlab.suse.de/galaxy/infrastructure/-/blob/master/srv/salt/minima/mirror-images.conf

And:
https://gitlab.suse.de/galaxy/infrastructure/-/blob/master/srv/salt/minima-bv/mirror-images.conf

Once done, you will need to apply states in the mirrors.

srbarrios: Better don't keep the image already sync. and use the new one, there is no reason to keep our mirror outdated.

srbarrios: We don't want to keep the mirror outdate just for the sake of re-use what was mirrored.
Let's update everything.

srbarrios: What we do in our mirror is to mirror the exactly path used in the origin.
So, if you have ORIGIN/my_path/my_file we should provide it through MIRROR/my_path/my_file

vandabarata:
infra PR: https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/601

PR #1176, opened 2022-10-03
Fix typo in README.md

zzaimeche:
Fix typo in symlink command 

## What does this PR change?

`ln-sfn` is now `ln -sfn`


rjmateus (approved):
thanks :+1:

zzaimeche:
Not sure what's up with the checks here; I made the change in the browser through the github web ui instead of the commandline so maybe that confused it somehow? Maybe there's some hook that runs with one workflow and not the other...

rjmateus:
@zzaimeche did you rabase on top of master and see if the checks keep failing?

zzaimeche:
I think github doesn't like it because I made the change from the webui, which created a branch directly against sumaform origin rather than a branch on my own fork. I can close this PR and open an identical one from my fork and see if that works. :)

PR #1177, opened 2022-10-04
Use newer image in example libvirt and ssh config

zzaimeche:
This commit updates the image name to "sles15sp4o" in the example config for libvirt and ssh. Previously, this was set to "sles12sp3", an image which is no longer available, so `terraform apply` would fail.


zzaimeche:
(I'm not sure if the server product_version should also be updated. That worked for me so I left it alone.)

agraul (approved):
LGTM

rjmateus (approved):

PR #1178, opened 2022-10-05
Adding missing support for Ubuntu 22.04

srbarrios:
## What does this PR change?

It seems we still missing some code pieces to support Ubuntu 22.04, this PR also fix some typo introduced in an older PR.


juliogonzalez (approved):
From what I can tell, it looks good. Thanks for fixing the repos for the other Ubuntu and Debian versions.

PR #1180, opened 2022-10-11
Use openSUSE 15.4 for ARM

Bischoff:
## What does this PR change?

switch from 15.3 to 15.4 for openSUSE ARM minion

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -716,7 +716,7 @@ packages: ["avahi", "nss-mdns", "qemu-guest-agent", "salt-minion"]
   
 %{ endif }
 
-%{ if image == "opensuse153armo" }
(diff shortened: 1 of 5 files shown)


backend_modules/libvirt/host/user_data.yaml
```diff
@@ -725,4 +725,4 @@ packages: ["salt-minion"]
 
 runcmd:
   - zypper removerepo --all
-%{ endif }
\ No newline at end of file
+%{ endif }
```
nodeg, line 728: Missing new line at the end.

nodeg (approved):

backend_modules/libvirt/host/user_data.yaml
Bischoff: Yes, I added it :-P . It was missing.

PR #1182, opened 2022-10-15
Instance type changes

mcalmer:
## What does this PR change?

Use t3(a) instance types for bastion host and mirror. They are cheaper an more powerful.
But this has changes to the device names which are important for the mirror.
Try to handle all of this in terraform and salt.

### DIFF
--- a/backend_modules/aws/base/main.tf
+++ b/backend_modules/aws/base/main.tf
@@ -95,7 +95,7 @@ module "bastion" {
   name                          = "bastion"
   connect_to_additional_network = true
   provider_settings = {
-    instance_type   = "t2.micro"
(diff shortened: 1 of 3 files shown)


mcalmer:
@rjmateus Not sure if this is it the right way to handle it. I wonder why AWS Console show different device names than we finally get? Maybe there are other places where we need to do changes?

rjmateus:
@mcalmer I'm not sure about that. I would say we should test it and see if it solves the issue.

rjmateus (approved):

mcalmer:
I think @maximenoel8 PR does the same but also more at other places. So I close this one in favor of https://github.com/uyuni-project/sumaform/pull/1183

PR #1183, opened 2022-10-17
Add support for AWS partition name

maximenoel8:
## What does this PR change?

When deploying on AWS, if we use an instance type > t2.XXX, the partition name will be nvme1nX . Sumaform doesn't support this naming for mirror and server with additional disk.

This PR adds nvme1nX naming support

traditionaly partitions are just numbered after the device name:

sda1, sda2, xvdf1, xvdf2, etc.

On NVMe drives this seems to have changed as the device name ends with a number. So they use p1, p2, etc. for partition 1, partition 2, etc.

nvme0n1p1,
nvme0n1p2,
nvme1n1p1

### DIFF
--- a/salt/mirror/configuration.sls
+++ b/salt/mirror/configuration.sls
@@ -145,9 +145,15 @@ scc-data_script:
 # partitioning
 
 {% set fstype = grains.get('data_disk_fstype') | default('ext4', true) %}
+{% if grains['data_disk_device'] == "nvme1n1" %}
+{% set partition_name = '/dev/' + grains['data_disk_device'] + 'p1' %}
+{% else %}
+{% set partition_name = '/dev/' + grains['data_disk_device'] + '1' %}
+{% endif %}
+
 mirror_partition:
   cmd.run:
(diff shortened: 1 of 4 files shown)


mcalmer:
Very similar to https://github.com/uyuni-project/sumaform/pull/1182

salt/mirror/configuration.sls
```diff
@@ -145,9 +145,15 @@ scc-data_script:
 # partitioning
 
 {% set fstype = grains.get('data_disk_fstype') | default('ext4', true) %}
+{% if grains['data_disk_device'] == "nvme1n1" %}
+{% set partition_name = '/dev/' + grains['data_disk_device'] + 'p1' %}
+{% else %}
+{% set partition_name = '/dev/' + grains['data_disk_device'] + '1' %}
```
srbarrios, line 151: I don't get the reason of this different appended partition name "p1" or "1".
Could you explain it a bit in the description of the PR please?

mcalmer: done

srbarrios: thank you

mcalmer:
@maximenoel8 I closed my PR. Would you please change the default type of the bastion host to "t3" or "t3a" ?
This seems to be the only change which is missing compared to my PR.

maximenoel8:
I tested this changes with testsuite deployment ( mirror + additional disk for server)

PR #1184, opened 2022-10-17
Removing unnecessary dependencies for pod proxy

srbarrios:
## What does this PR change?

Removing unnecessary dependencies for pod proxy


cbosdo (approved):
👍🏻

PR #1190, opened 2022-10-24
update uyuni gpg key with enhanced validity period

mcalmer:
## What does this PR change?

The uyuni key was expired and the validity period was enhanced.
Use the changed key for sumaform deployment.


srbarrios (approved):
Thank you!

PR #1192, opened 2022-10-25
Add custom repositories needed to run testsuite on AWS deploying SUMA on SLES

maximenoel8:
## What does this PR change?

Running the testsuite when deploying SUMA on SLES need some internal repositories. 
This repositories are not available on cloud.
This repository list is design to list all repositories that need to be mirror in cloud.
Internal and external repositories are mirror to use the default mirror settings in sumaform.


srbarrios (approved):

salt/mirror/utils/minimum_repositories_testsuite.yaml
```diff
@@ -0,0 +1,53 @@
...
+  - url:  http://download.suse.de/ibs/SUSE/Updates/SLE-Manager-Tools/15/x86_64/update/
+    archs: [x86_64]
+
+  - url:  http://download.suse.de/ibs/SUSE/Products/SLE-Manager-Tools/15/x86_64/product/
+    archs: [x86_64]
+
+  - url:  http://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/SLE15-Uyuni-Client-Tools/SLE_15/
+    archs: [x86_64]
+
+  - url:  http://minima-mirror.mgr.prv.suse.net/repo/$RCE/RES7-SUSE-Manager-Tools/x86_64/
```
mcalmer, line 37: Why is this taken from another mirror?

I think the source is https://download.suse.de/download/ibs/SUSE/Updates/RES/7-CLIENT-TOOLS/x86_64/update/
but need to be written into a different destination.

mcalmer (commented):
I am not sure why this is not in `salt/mirror/etc/` where the main minima.yaml is.
And if it would be better to get all the repos which are in SCC also from it?

Access to the internal resources are only needed for devel packages.

salt/mirror/utils/minimum_repositories_testsuite.yaml
```diff
@@ -0,0 +1,53 @@
+storage:
+  type: file
+  path: /srv/mirror
+
+http:
+
+  - url:  http://download.suse.de/ibs/SUSE/Updates/Ubuntu/20.04-CLIENT-TOOLS/x86_64/update/
```
rjmateus, line 7: download.suse.de are not available on the cloud

rjmateus (commented):
LGTM, just one comment

salt/mirror/utils/minimum_repositories_testsuite.yaml
maximenoel8: Yes indeed, this is a list for internal mirror.

salt/mirror/utils/minimum_repositories_testsuite.yaml
maximenoel8: Good catch, I forget about that workaround. I will fix it. Thx

maximenoel8: I don't copy the SCC repositories because it takes too much time. I only use specifics repositories.

PR #1193, opened 2022-10-25
Remove old amis, update ubuntu 2204 ami

maximenoel8:
## What does this PR change?

Remove old amis
Update ubuntu owner

### DIFF
--- a/backend_modules/aws/base/ami.tf
+++ b/backend_modules/aws/base/ami.tf
@@ -236,7 +236,7 @@ data "aws_ami" "centos7" {
 data "aws_ami" "ubuntu2204" {
   most_recent = true
   name_regex  = "^ubuntu/images/hvm-ssd/ubuntu-jammy-22.04"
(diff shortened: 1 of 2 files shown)


mcalmer (approved):

backend_modules/aws/base/ami.tf
```diff
@@ -256,7 +236,7 @@ data "aws_ami" "centos7" {
 data "aws_ami" "ubuntu2204" {
   most_recent = true
   name_regex  = "^ubuntu/images/hvm-ssd/ubuntu-jammy-22.04"
-  owners      = ["679593333241"]
+  owners      = ["099720109477"]
```
rjmateus, line 239: why did you change the owner for the ubuntu image? If I got it right, this owner is for an image that is not official.
We should keep the official one, otherwise we can be using corrupt images and pay more than needed.

rjmateus (changes requested):

PR #1209, opened 2022-11-17
Align Rocky 9 tools repo with Rocky 8

Bischoff:
## What does this PR change?

Was there a good reason for using a different path as for Rocky 8 for the tools ?

This PR aligns the path for both distros.

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -643,7 +643,7 @@ packages: ["avahi", "nss-mdns", "salt-minion"]
 yum_repos:
   # repo for salt
   tools_pool_repo:
-    baseurl: http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Master:/EL9-Uyuni-Client-Tools/EL_9
+    baseurl: http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Stable:/EL9-Uyuni-Client-Tools/EL_9
     enabled: true
     gpgcheck: false
     name: tools_pool_repo


nodeg (approved):

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -643,7 +643,7 @@ packages: ["avahi", "nss-mdns", "salt-minion"]
 yum_repos:
   # repo for salt
   tools_pool_repo:
-    baseurl: http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Master:/EL9-Uyuni-Client-Tools/EL_9
+    baseurl: http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Stable:/EL9-Uyuni-Client-Tools/EL_9
```
srbarrios, line 646: Minima mirror seems not ready yet. Please populate it before merge this.

http://minima-mirror.mgr.suse.de/repositories/systemsmanagement:/Uyuni:/Stable:/EL9-Uyuni-Client-Tools/EL_9

Bischoff: it is ready

http://minima-mirror-bv.mgr.prv.suse.net/repositories/systemsmanagement:/Uyuni:/Stable:/EL9-Uyuni-Client-Tools/EL_9/

We're speaking about BV for now.

srbarrios: Ok, about adding it only to BV for now.
I think we need to revisit the use of HTTP section, instead of SCC section in the mirror. I think that for BV, we must be using only SCC products + our custom channels injecting the MU repositories. And not using other repositories coming from IBS/OBS.

srbarrios (approved):

backend_modules/libvirt/host/user_data.yaml
Bischoff: Yes, if the client tools for Uyuni exist in SCC (I sincerely don't know whether it's the case or not).

srbarrios: In such case, we need to move this https://gitlab.suse.de/galaxy/infrastructure/-/blob/master/srv/salt/minima-bv/minima-rhlike.yaml#L26-35 to SCC section, and see if that change the URI

juliogonzalez: > Yes, if the client tools for Uyuni exist in SCC (I sincerely don't know whether it's the case or not).

No, they don't. Uyuni does not exist in SCC at all.

srbarrios: > > Yes, if the client tools for Uyuni exist in SCC (I sincerely don't know whether it's the case or not).
> 
> No, they don't. Uyuni does not exist in SCC at all.

We are at BV, these are repos in use in 4.3, we they don't?

srbarrios: > No, those are beta. Use the same names without `-Beta` in them. Beta is only used by beta customers for milestones.

They don't exist yet.

ktsamis (approved):

backend_modules/libvirt/host/user_data.yaml
Bischoff: For now, i will intentionally merge this wrong path. @srbarrios created a card so we can fix this mess quietly. I think other products than rocky 9 are concerned.

Bischoff:
As commented above, merged this **WRONG** path that uses Uyuni when we know we should be using SUSE Manager.

PR #1211, opened 2022-11-19
PR testsuite should test like master

mcalmer:
## What does this PR change?

fix testsuite error on enabling repos for prometheus test.
The test enable also the "tools update" repo which does not exist when the product_version grain is set to "uyuni-pr"


rjmateus (approved):
LGTM

PR #1212, opened 2022-11-21
Add EL 9 manager tools

Bischoff:
## What does this PR change?

RES9 is being renamed to EL9 due to the phasing out of RES in favour of SLL.

This PR introduces a variable `rhlike_client_tools_prefix` that maps to either `EL` or `RES` according to the release.

This PR also renames existing variable `centos_client_tool_prefix` to the same name.


juliogonzalez (approved):
Repos look good, but I did only a light review of the logic.

Logic looks good, but for other reviewers:

- For Uyuni, we have EL8 and EL9 (we had to rename CentOS8 to EL8 when CentOS8 went EoL)
- For SUSE manager we have RES8 and EL9. We can't rename codestreams, and RES8 is still "alive".

nodeg (approved):

rjmateus (approved):
LGTM

PR #1216, opened 2022-11-22
Revert "PR testsuite should test like master"

jordimassaguerpla:
Reverts uyuni-project/sumaform#1211

Hi ! This is not correct. This PR is adding a repo from opensuse mirror infrastructure:

`baseurl: http://{{ grains.get("mirror") | default("download.opensuse.org", true) }}/repositories/systemsmanagement:/Uyuni:/Master:/SLE15-Uyuni-Client-Tools/SLE_15/`

we are in purpose not using these repos in the `uyuni-pr`

We should use the repos from the worker instead:

See: https://github.com/SUSE/susemanager-ci/blob/master/jenkins_pipelines/environments/common/pipeline-pull-request.groovy#L250

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -339,7 +339,7 @@ lock_firewalld_prometheus_config_cmd:
      - name: zypper addlock firewalld-prometheus-config
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


mcalmer:
The problem is, that the testsuite expect a repo with a special name. Just provide an alternative url under the same name and we are fine. But we need a 'tools_update_repo'.

PR #1219, opened 2022-12-01
Adding support for SLES15.4 in Avahi

srbarrios:
## What does this PR change?

Otherwise we are relying on the repositories injected during the Cloud Init phase.
And I have a open PR to precisely get rid of these left-overs in Cloud init. (https://github.com/uyuni-project/sumaform/pull/1218)

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -3,7 +3,7 @@ include:
 
 {% if grains['use_avahi'] and grains.get('osmajorrelease', None) != None %}
 
-# TODO: remove the following state when fix to bsc#1163683 is applied to all the SLES <= SLES15SP3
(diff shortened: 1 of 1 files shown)


Bischoff (approved):

salt/default/avahi.sls
```diff
@@ -4,7 +4,7 @@ include:
 {% if grains['use_avahi'] and grains.get('osmajorrelease', None) != None %}
 
 # TODO: remove the following state when fix to bsc#1163683 is applied to all the SLES <= SLES15SP3
```
juliogonzalez, line 6:
```suggestion
# TODO: remove the following state when fix to bsc#1163683 is applied to all the SLES <= SLES15SP4
```

juliogonzalez (changes requested):

PR #1220, opened 2022-12-06
Aws set private ip and support multiple environments

mcalmer:
## What does this PR change?

This PR is implementing 2 things:

1. possibility to create new private and additional private networks with instances for a given public network and VPC. This allow multiple separated environments in the same Cloud VPC using the same mirror and bastion host
2. assign fixed private IP addresses to the instances. Together with "Route 53" we can have dedicated names for our instances.

### DIFF
--- a/backend_modules/aws/base/main.tf
+++ b/backend_modules/aws/base/main.tf
@@ -28,16 +28,18 @@ locals {
 module "network" {
   source = "../network"
 
-  availability_zone  = local.availability_zone
-  region             = local.region
-  ssh_allowed_ips    = local.ssh_allowed_ips
-  name_prefix        = local.name_prefix
-  create_network     = local.create_network
-  create_db_network  = local.create_db_network
-  private_network    = local.private_network
-  additional_network = local.additional_network
-  public_subnet_id   = local.public_subnet_id
-  vpc_id             = local.vpc_id
+  availability_zone         = local.availability_zone
+  region                    = local.region
+  ssh_allowed_ips           = local.ssh_allowed_ips
+  name_prefix               = local.name_prefix
+  create_network            = local.create_network
+  create_private_network    = local.create_private_network
+  create_additional_network = local.create_additional_network
+  create_db_network         = local.create_db_network
+  private_network           = local.private_network
+  additional_network        = local.additional_network
+  public_subnet_id          = local.public_subnet_id
+  vpc_id                    = local.vpc_id
 }
 
 locals {
@@ -103,7 +105,7 @@ module "bastion" {
   base_configuration            = local.configuration_output
   image                         = lookup(var.provider_settings, "bastion_image", "opensuse154o")
   name                          = "bastion"
-  connect_to_additional_network = true
+  //connect_to_additional_network = true
   provider_settings = {
     instance_type   = "t3a.micro"
     public_instance = true
--- a/backend_modules/aws/base/main.tf
+++ b/backend_modules/aws/base/main.tf
@@ -105,7 +105,6 @@ module "bastion" {
   base_configuration            = local.configuration_output
   image                         = lookup(var.provider_settings, "bastion_image", "opensuse154o")
   name                          = "bastion"
-  //connect_to_additional_network = true
   provider_settings = {
     instance_type   = "t3a.micro"
     public_instance = true
--- a/backend_modules/aws/base/main.tf
+++ b/backend_modules/aws/base/main.tf
@@ -8,6 +8,8 @@ locals {
   key_file = lookup(var.provider_settings, "key_file", null)
 
   create_network                       = lookup(var.provider_settings, "create_network", true)
+  create_private_network               = lookup(var.provider_settings, "create_private_network", true)
+  create_additional_network            = lookup(var.provider_settings, "create_additional_network", true)
   create_db_network                    = lookup(var.provider_settings, "create_db_network", false)
   public_subnet_id                     = lookup(var.provider_settings, "public_subnet_id", null)
   private_subnet_id                    = lookup(var.provider_settings, "private_subnet_id", null)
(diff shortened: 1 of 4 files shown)


mcalmer:
Use this branch for run
https://ci.nue.suse.com/view/Manager/view/Uyuni/job/uyuni-master-dev-acceptance-tests-AWS/463/

without changed config.
Test the old config which should still work also with the new code

backend_modules/aws/base/main.tf
```diff
@@ -30,7 +34,10 @@ module "network" {
   name_prefix        = local.name_prefix
   create_network     = local.create_network
   create_db_network  = local.create_db_network
+  private_network    = local.private_network
```
rjmateus, line 37: I think you also need to pass in here the variables  `create_private_network` and `create_additional_network`, otherwise, it will use the default values for that module.

backend_modules/aws/network/main.tf
```diff
@@ -26,9 +26,19 @@ resource "aws_internet_gateway" "main" {
   }
 }
 
+data "aws_vpc" "selected" {
+  count = var.create_network ? 0 : 1
+
+  id = var.vpc_id
+  filter {
+    name   = "tag:Name"
+    values = ["mgrinfra-vpc"]
```
rjmateus, line 35: I'm not sure if this value should be hard-coded.

backend_modules/aws/network/main.tf
```diff
@@ -91,13 +105,13 @@ resource "aws_main_route_table_association" "vpc_internet" {
...
 resource "aws_route_table" "private" {
-  count = var.create_network ? 1 : 0
+  count = var.create_private_network ? 1 : 0
 
   vpc_id = local.vpc_id
 
   route {
     cidr_block     = "0.0.0.0/0"
-    nat_gateway_id = aws_nat_gateway.nat[0].id
+    nat_gateway_id = data.aws_nat_gateway.default.id
```
rjmateus, line 114: I think this should be made conditional on the creation of the network. also look at line 53 of this file, where the nat is created.

rjmateus (commented):
first round. I will continue tomorrow

backend_modules/aws/network/main.tf
rjmateus: maybe you can pass the VPC ID directly when you want to create a private network

mcalmer: I think I have fixed it

backend_modules/aws/network/main.tf
mcalmer: I think also this is fixed now

backend_modules/aws/base/main.tf
```diff
@@ -91,7 +105,7 @@ module "bastion" {
   base_configuration            = local.configuration_output
   image                         = lookup(var.provider_settings, "bastion_image", "opensuse154o")
   name                          = "bastion"
-  connect_to_additional_network = true
+  //connect_to_additional_network = true
```
rjmateus, line 108: there is a reason to comment this line?

backend_modules/aws/network/main.tf
```diff
@@ -53,10 +60,15 @@ resource "aws_nat_gateway" "nat" {
   }
 }
 
+data "aws_nat_gateway" "default" {
```
rjmateus, line 63: Why do you need to create an extra subnet? We should already have a nat gateway for the public subnet:
https://github.com/uyuni-project/sumaform/blob/master/backend_modules/aws/network/main.tf#L43-L54

backend_modules/aws/network/main.tf
```diff
@@ -86,18 +98,18 @@ resource "aws_route_table" "additional-public" {
...
-  count = var.create_network ? 1 : 0
+  count = var.create_private_network ? 1 : 0
 
-  vpc_id = local.vpc_id
+  vpc_id = local.l_vpc_id
 
   route {
     cidr_block     = "0.0.0.0/0"
-    nat_gateway_id = aws_nat_gateway.nat[0].id
+    nat_gateway_id = var.create_network ? aws_nat_gateway.nat[0].id : data.aws_nat_gateway.default[0].id
```
rjmateus, line 112: Maybe we should make the nat_gateway_id a parameter to be able to reuse it.

rjmateus (commented):
Just few comments left.
Did you try the deployment with azure? I would say lets focus on aws, if it's needed in azure in the future, then we can implement it

backend_modules/azure/host/main.tf
```diff
@@ -80,6 +83,7 @@ resource "azurerm_network_interface" "suma-additional-nic" {
 resource "azurerm_linux_virtual_machine" "instance" {
   count                            = var.quantity
   name                             = "${local.resource_name_prefix}${var.quantity > 1 ? "-${count.index + 1}" : ""}"
+  private_ip                       = local.private_ip
```
rjmateus, line 86: I think this will not work, since in azure the names are different. it should be "private_ip_address"
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_virtual_machine#private_ip_address

rjmateus (commented):
one more item for azure. But I think if we cannot test it, or no need to use it now, then lets not change if

backend_modules/aws/base/main.tf
mcalmer: I think I will remove it.
When we just create bastion + mirror we only need the private network. Maybe as "infrastructure network".
An additional is useless as it is empty.

When we setup a test environment we now specify private and additional for every environment.

backend_modules/aws/network/main.tf
mcalmer: This is not an additional. It is a "data" node, not a "resource". With this I let terraform get the existing configuration I need to setup the environment.

backend_modules/azure/host/main.tf
mcalmer: No, I didn't test it. I think i will remove the commit for azure

backend_modules/aws/network/main.tf
rjmateus: sure, my bad. I need a new pair of glasses :)

PR #1221, opened 2022-12-09
do not hide configure-proxy issue

mbussolotto:
## What does this PR change?

Follow up of https://github.com/SUSE/spacewalk/issues/18094#issuecomment-1341160388

Currently it could happen that something goes wrong running configure-proxy.sh, but error is not visible (an example here https://github.com/SUSE/spacewalk/issues/18094#issuecomment-1341130039 ).  We should make this error visible

### DIFF
--- a/salt/proxy/init.sls
+++ b/salt/proxy/init.sls
@@ -206,7 +206,7 @@ ssl-building-ca-configuration:
 
 configure-proxy:
   cmd.run:
-    - name: configure-proxy.sh --non-interactive --rhn-user={{ grains.get('server_username') | default('admin', true) }} --rhn-password={{ grains.get('server_password') | default('admin', true) }} --answer-file=/root/config-answers.txt ; true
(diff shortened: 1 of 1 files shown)


mbussolotto:
I'll do that in `manager-TEST-Orion-acceptance-tests`, since I think PR testsuite allows just to test uyuni PR and not sumaform PR

cbosdo (commented):
I have no real clue about that piece of code, but the `; true` seems like it allows the bootstrap script generation states to be done even if the configure failed. I'm not sure what is the good approach here

mbussolotto:
test: https://ci.suse.de/view/Manager/view/Manager-Test/job/manager-TEST-Orion-acceptance-tests/1076/

rjmateus (approved):
LGTM. we should handle any problem(s) that can happen during the proxy installation with sumaform

mcalmer (approved):
give it a try

Bischoff (approved):

Bischoff:
Let's try it, we can always revert.

PR #1222, opened 2022-12-09
Use original repo for uyuni repos

mcalmer:
## What does this PR change?

We have often problems with Uyuni repositories on opensuse.org as the metadata do not match the packages on the mirror we get redirected to.

This is a try to use downloadcontent.opensuse.org for repositories/systemsmanagement mostly for Uyuni repositories.

Let's see if we get better results with this setup.


juliogonzalez (approved):
Interesting, I was not aware of this repository.

rjmateus (approved):
Let's give it a try. I was not aware of this endpoint

mcalmer:
I did a run with this branch and it was looking good. So I merged now and we will see if we gain stability.

PR #1225, opened 2022-12-15
Rhel9 support on aws

maximenoel8:
## What does this PR change?

- Update user_data to support rhel9 deployment on AWS
- Add rhel 9 support to controller


backend_modules/aws/host/user_data.yaml
```diff
@@ -118,8 +119,21 @@ yum_repos:
...
+    enabled: true
+    gpgcheck: false
+    name: epel
+    priority: 98
 
 packages: ["salt-minion"]
+runcmd:
+  # WORKAROUND: rhel9 change ssh security for ssh key
+  - echo "PermitRootLogin yes" >> /etc/ssh/sshd_config
+  - update-crypto-policies --set LEGACY
```
mcalmer, line 134: I am sceptical to use LEGACY here. Can you say why we needs this?

backend_modules/aws/host/user_data.yaml
```diff
@@ -118,8 +119,21 @@ yum_repos:
     gpgcheck: false
     name: temp_tools_pool_repo
     priority: 98
+  epel:
+    baseurl: http://download-cc-rdu01.fedoraproject.org/pub/epel/9/Everything/x86_64/
+    failovermethod: priority
+    enabled: true
+    gpgcheck: false
+    name: epel
+    priority: 98
```
srbarrios, line 128: Cloud-init stage has as a only goal prepare the instance, but the repositories we will use after the cloud-init stage must go to the salt state part. If we are not installing anything in this image during cloud-init stage, then... I would say we don't need this repository here.

maximenoel8: salt-minion is only available in this repository.

backend_modules/aws/host/user_data.yaml
maximenoel8: With the depreciation of SHA1 on RHEL9, using aws ssh key with terraform is blocked by RHEL9. Without this command, I fail to connect the minion and run salt.

mcalmer: Should we better create a new key? @juliogonzalez ?

juliogonzalez: What key are we talking about? The key pair created at AWS to connect to the VMs? Is it SHA1 now?

If so, yes, just create an key and use it.

maximenoel8: I switch to use an ed25519 ssh key and it works

PR #1227, opened 2022-12-21
use latest Rocky Linux images

mcalmer:
## What does this PR change?

The specific SP versions are removed as soon as a newer is available. Use always the latest by referencing the major version folder.


rjmateus (approved):
LGTM

PR #1230, opened 2022-12-27
Fix "download_ubuntu_repo.sh" after changes in d.o.o repos

meaksh:
## What does this PR change?

The latest upgrade of HTML pages from download.opensuse.org is causing that our script for downloading ubuntu repos is now producing a non-zero retcode that is breaking our automation. The issue is caused because latest upgrades in the HTML from d.o.o is making our script to raise some exception while wget is recursively downloading.

This PR simply fixes the script by passing `--adjust-extension` to wget command in order to prevent the mess.


rjmateus (approved):
lgtm

PR #1233, opened 2023-01-02
overwrite host FQDN via provider settings

mcalmer:
## What does this PR change?

In AWS we use Route 53 to provide meaningful hostnames for the systems.
We also make use of some tricks with CNAMES. 
To be able to use these hostnames, this PR add the possibility to overwrite the FQDN with a given value in the provider settings.

See also https://github.com/SUSE/spacewalk/issues/19888

### DIFF
--- a/backend_modules/aws/host/main.tf
+++ b/backend_modules/aws/host/main.tf
@@ -161,6 +161,11 @@ resource "aws_volume_attachment" "data_disk_attachment" {
 }
 /** END: Set up an extra data disk */
 
+locals {
+  hname = (local.overwrite_fqdn != null ? split(".", local.overwrite_fqdn)[0] :
+	 (length(aws_instance.instance) > 0 ? replace(aws_instance.instance[0].private_dns, ".${local.region == "us-east-1" ? "ec2.internal" : "${local.region}.compute.internal"}", "") : ""))
+}
+
 /** START: provisioning */
 resource "null_resource" "host_salt_configuration" {
   depends_on = [aws_instance.instance, aws_volume_attachment.data_disk_attachment]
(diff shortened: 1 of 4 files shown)


srbarrios (approved):
We tested on the last runs, so it should be good to merge. Unless we want to use another strategy, but LGTM.

Bischoff (commented):
Speaking of another strategy, wasn't it possible to simply tune with ` name_prefix =` and `domain =` in base module ? @mcalmer

mcalmer:
This does not have an effect on the hostnames AWS return for us.

Bischoff:
> This does not have an effect on the hostnames AWS return for us.

ok, and i suppose we could not adapt to those hostnames?

(I could see the problem better with some example of those names)

mcalmer:
> > This does not have an effect on the hostnames AWS return for us.
> 
> ok, and i suppose we could not adapt to those hostnames?

I don't know. Maybe we can with some Cloud Tricks I do not know. Maybe Route 53 + some special configuration make it possible.
 
> (I could see the problem better with some example of those names)

AWS set "ip-172-16-3-6.eu-central-1.compute.internal" and we have also "uyuni-master-srv.uyuni.cloud"

```
$> nslookup 172.16.3.6
6.3.16.172.in-addr.arpa name = ip-172-16-3-6.eu-central-1.compute.internal.

Authoritative answers can be found from:

$> nslookup uyuni-master-srv.uyuni.cloud
Server:         172.16.0.2
Address:        172.16.0.2#53

Non-authoritative answer:
Name:   uyuni-master-srv.uyuni.cloud
Address: 172.16.3.6
```

Bischoff:
> AWS set "ip-172-16-3-6.eu-central-1.compute.internal" and we have also "uyuni-master-srv.uyuni.cloud"
> 
> ```
> $> nslookup 172.16.3.6
> 6.3.16.172.in-addr.arpa name = ip-172-16-3-6.eu-central-1.compute.internal.
> 
> Authoritative answers can be found from:
> 
> $> nslookup uyuni-master-srv.uyuni.cloud
> Server:         172.16.0.2
> Address:        172.16.0.2#53
> 
> Non-authoritative answer:
> Name:   uyuni-master-srv.uyuni.cloud
> Address: 172.16.3.6
> ```

OK, indeed what I suggested will not work out. If you set
```
name_prefix = "uyuni-master-"
domain = "uyuni.cloud"
```
(as you probably already do), you will be fine for direct resolution, but reverse resolution will always return a domain in `eu-central-1.compute.internal`, if you use the DNS server provided by AWS.

Let me have a second look at this PR now I understand better the problem.

mcalmer:
Update: I have now fully setup Route 53 DNS server in AWS with reverse lookup. This PR change now also the hostnames in the Cloud to the provided overwrite FQDN. With this the testsuite works again as good as in our datacenter.

PR #1235, opened 2023-01-10
Deprecate use_os_unreleased_updates flag

Bischoff:
## What does this PR change?

This PR removes the `use_os_unreleased_updates` flag.

The plan is to start testing SLE updates with the build validation test suite. The MI numbers are passed through the usual json file.

The technical advantage over using `Maintenance:/Test:` aggregation of maintenance updates is that it is easier to spot the MI causing a regression, as we apply only one.

The organizational advantage is that we have not the bandwidth to follow another CI test suite that would retest periodically all features with the unreleased updates.

### DIFF
--- a/.ci-travis/salt-client-validation
+++ b/.ci-travis/salt-client-validation
@@ -20,7 +20,6 @@ roles: [client]
 mirror: null
 auto_register: false
 testsuite: true
-use_os_unreleased_updates: 0
 ipv6: {'accept_ra': true, 'enable': true}
 EOF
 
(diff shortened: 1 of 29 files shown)


srbarrios:
I think is also interesting to share that we are developing a new test suite that will cover the validation of our SUMA product using SLE Unreleased maintenance updates.
We will do it through the injection in SUMA server of the IBS/OBS repositories containing these MIs, in addition to custom channels/repositories that will be included as part of the Activation Key for our Proxy and our SLE Minion.

rjmateus (approved):
lgtm

PR #1237, opened 2023-01-11
Permit root login in Oracle and Alma 9

ktsamis:
## What does this PR change?

Permit root login in Oracle and Alma 9 because it's not set by cloud init and we need this before adding the ssh key from the controller.


ktsamis:
Not sure if it's needed in oracle and alma 8 but I guess either nobody used it or it's not needed so not adding it now.

rjmateus (approved):
lgtm

Bischoff (approved):

PR #1240, opened 2023-01-13
Disable IPv6 for apt-get until we have IPv6 in Provo

Bischoff:
## What does this PR change?

Ubuntu 22.04 does not always fall back correctly to IPv6 when IPv6 route to the outside is missing, which is the case for now in Provo.

Remove these lines (and many other similar workarounds we did for IPv6 in the past) when IPv6 appears in Provo.

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -480,6 +480,8 @@ apt:
 runcmd:
   # WORKAROUND: cloud-init in Ubuntu 22.04 does not take care of the following
   - echo "PermitRootLogin yes" >> /etc/ssh/sshd_config
+  # WORKAROUND: disable IPv6 until we have it in Provo
+  - echo 'Acquire::ForceIPv4 "true";' > /etc/apt/apt.conf.d/99force-ipv4
   - systemctl restart sshd
   - systemctl start qemu-guest-agent
 


nodeg (approved):
LGTM.
The solution comes from https://unix.stackexchange.com/questions/9940/convince-apt-get-not-to-use-ipv6-method

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -480,6 +480,8 @@ apt:
 runcmd:
   # WORKAROUND: cloud-init in Ubuntu 22.04 does not take care of the following
   - echo "PermitRootLogin yes" >> /etc/ssh/sshd_config
+  # WORKAROUND: disable IPv6 until we have it in Provo
+  - echo 'Acquire::ForceIPv4 "true";' > /etc/apt/apt.conf.d/99force-ipv4
```
srbarrios, line 484: Why only in Ubuntu22.04?

srbarrios: Ah all clear, I missed the comment.

srbarrios (approved):

backend_modules/libvirt/host/user_data.yaml
Bischoff: Because we saw the issue there - we don't know if it also affects 20.04 and earlier.

PR #1244, opened 2023-01-19
Make eip feature for bastion configurable

maximenoel8:
## What does this PR change?

Deploying a new bastion with instance_with_eip set to true is failing the deployment.
When I deploy my environment with a new bastion, terraform will try to connect to bastion with a wrong IP.

I created a new base variable to configure the bastion setup.


mcalmer (approved):
should work as the default is still true

PR #1246, opened 2023-01-23
SUSE AMIs in AWS changed the owner to the marketplace account

mcalmer:
## What does this PR change?

Only current images changed. The very old once are still only listed under the old owner ID:


rjmateus (approved):
LGTM

srbarrios (approved):
LGTM, according to the discussion in Slack.
But I'm not sure if we need to change others, somewhere, for SUMA Server and Proxy AMIs

https://pint.suse.com/?resource=images&csp=amazon&state=active&search=manager-server-4-3

mcalmer:
AMIs for SUSE Manager Server/Proxy are not yet defined. So we would need to add them first.

maximenoel8:
We don't use SUMA AMI in sumaform. I test its every 3-4 months manually.

PR #1247, opened 2023-01-25
Add SLE Micro 5.3

nodeg:
## What does this PR change?

This adds SLE Micro 5.3 as supported client. ~~We still need to build the image in the OBS, tough.~~ @aaannz added it here: https://build.opensuse.org/package/show/systemsmanagement:sumaform:images:microos/SLE-MicroOS53


vandabarata (approved):
lgtm

srbarrios (approved):

backend_modules/null/base/variables.tf
```diff
@@ -65,7 +65,7 @@ variable "provider_settings" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
-  default     = [ "almalinux8o", "amazonlinux2o", "centos7o", "opensuse153o", "opensuse154o", "rocky8o", "rocky9o", "sles12sp4o", "sles12sp5o", "sles15sp3o", "sles15sp4o", "ubuntu1804o", "ubuntu2004o", "ubuntu2204o" ]
+  default     = [ "almalinux8o", "amazonlinux2o", "centos7o", "opensuse153o", "opensuse154o", "rocky8o", "rocky9o", "sles12sp4o", "sles12sp5o", "sles15sp3o", "sles15sp4o", "slemicro52-ign", "slemicro53-ign", "ubuntu1804o", "ubuntu2004o", "ubuntu2204o" ]
```
ktsamis, line 68: So I don't think we need to download this by default. I think we need to specify it when we need it. This would download the image on every base module that we don't specify and we use default. I deliberately didn't add alma and oracle to this but if we add all the images to this let's add all of them

nodeg: Makes sense. Thank you! I was also wondering about `opensuse153o`, `sles12sp4o` and `ubuntu1804o`. I think we do not need those by default, too.

nodeg: Nevertheless I removed SLE Micro from there.

srbarrios (approved):

ktsamis (approved):

PR #1253, opened 2023-02-03
change for SLE Micro 5.3

mbussolotto:
## What does this PR change?

add some adjustment for SLE Micro

Fixes: https://github.com/SUSE/spacewalk/issues/20512


mbussolotto:
I tested it and I was able to deploy correctly SLE Micro 5.3, so it fixes `/var/lib/rpm/.rpm.lock` . Not sure if it's the best way to do that, but works. Let me know what you think about it

salt/minion/testsuite.sls
```diff
@@ -1,12 +1,13 @@
 {% if grains.get('testsuite') | default(false, true) %}
-
+{% if not grains['osfullname'] == 'SLE Micro' %}  #  Read-only file system: '/var/lib/rpm/.rpm.lock
```
aaannz, line 2: Maybe you can add a note, something like

`To avoid need to reboot after deployment requires are managed in kiwi profile at https://build.opensuse.org/package/show/systemsmanagement:sumaform:images:microos in sumaform profile`

Otherwise we would need to adapt states to use `transactional_update.pkg` module and reboot after provisioning.

mbussolotto: sorry, I can add it but I'm not sure if I get what you mean (still lack of knowledge on this argument :) ). I'll try to rephrase to check if I understand: the reason why we do not need to check if those packages are already installed it's because the image that we are using ( https://build.opensuse.org/package/show/systemsmanagement:sumaform:images:microos ) has already all the packages required?

aaannz: Yes, exactly. Kiwi files for sle micro images have sumaform profile in them with list of extra packages. So if there is later some need to add new package to support testsuite, rather then having them install in salt state and then reboot, they can be added to this sumaform profile in kiwi file.

For example see file https://build.opensuse.org/package/view_file/systemsmanagement:sumaform:images:microos/SLE-MicroOS53/SLE-Micro.kiwi?expand=1 at the end
```xml
    <packages type="image" profiles="sumaform">
        <!-- common requirements -->
        <package name="qemu-guest-agent"/>
        <package name="salt-minion"/>
        <!-- evil minions -->
        <!-- disabled by default, uncomment when needed
        <package name="evil-minions"/> -->
        <!-- test suite requirements -->
        <package name="aaa_base-extras"/>
        <package name="ca-certificates"/>
        <package name="wget"/>
        <!-- avahi support -->
        <package name="avahi"/>
        <package name="avahi-lang" />
        <package name="libavahi-core7"/>
        <!-- test suite test packages -->
        <package name="andromeda-dummy" />
        <package name="milkyway-dummy" />
        <package name="virgo-dummy" />
    </packages>
```

And I made a mistake in previous comment. The link to images is https://build.opensuse.org/project/show/systemsmanagement:sumaform:images:microos (notice `project` vs `package` in URL )

mbussolotto: thanks for the explanation :) I added comments

modules/cucumber_testsuite/main.tf
```diff
@@ -178,6 +178,29 @@ module "suse-sshminion" {
...
+  server_configuration = local.minimal_configuration
+  sles_registration_code = lookup(local.sles_registration_code, "slemicro-minion", null)
+
+  use_os_released_updates = true
+  ssh_key_path            = "./salt/controller/id_rsa.pub"
+  install_salt_bundle     = lookup(local.install_salt_bundle, "slemicro-minion", false)
+
+  additional_repos  = lookup(local.additional_repos, "slemicro-minion", {})
+  additional_repos_only  = lookup(local.additional_repos_only, "slemicro-minion", {})
+  additional_packages = lookup(local.additional_packages, "slemicro-minion", ["avahi", "avahi-lang", "libavahi-common3", "libavahi-core7"])
```
aaannz, line 199: I'm not sure about this here.

This list is passed to the `salt/default/pkgs.sls` and sure it is now doing nothing, but we are using `pkg.latest` there. So there is tiny window when image is not yet rebuild after avahi update where deployment can fail.

OTOH this list can keep our images in check if someone accidentally removes avahi.

aaannz (approved):
Look good! I have just one point to discuss, but whatever you decide I'm ok with it. :)

PR #1256, opened 2023-02-20
Use by default chrony except in specific cases

ktsamis:
## What does this PR change?

Use by default chrony except in specific cases, SLES 11.4, Debian 10 and Leap. Those use ntp.

Without this we don't set ntp or chrony in ubuntu, debian, ol, almalinux, rocky and in BV where we do use them we see failures because the clock drifts.

### DIFF
--- a/salt/default/time.sls
+++ b/salt/default/time.sls
@@ -23,9 +23,7 @@ timezone_setting:
 
 {% if grains['use_ntp'] %}
 
-{% if grains['osfullname'] == 'SLES' %}
-
-{% if grains['osrelease'] == '11.4' %}
+{% if ((grains['osfullname'] == 'SLES') and (grains['osrelease'] == '11.4')) or ((grains['os_family'] == 'Debian') and (grains['osrelease'] == '10')) or (grains['osfullname'] == 'Leap') %}
 
 ntp_pkg:
   pkg.installed:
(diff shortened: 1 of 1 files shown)


salt/default/time.sls
```diff
@@ -23,9 +23,7 @@ timezone_setting:
 
 {% if grains['use_ntp'] %}
 
-{% if grains['osfullname'] == 'SLES' %}
-
-{% if grains['osrelease'] == '11.4' %}
+{% if ((grains['osfullname'] == 'SLES') and (grains['osrelease'] == '11.4')) or ((grains['os_family'] == 'Debian') and (grains['osrelease'] == '10')) or (grains['osfullname'] == 'Leap') %}
```
Bischoff, line 26: suggestion of presentation (admitting it works:
```
{% if ((grains['osfullname'] == 'SLES') and (grains['osrelease'] == '11.4'))
      or ((grains['os_family'] == 'Debian') and (grains['osrelease'] == '10'))
      or (grains['osfullname'] == 'Leap') %}
%}
```

Bischoff (approved):
see suggestion

salt/default/time.sls
ktsamis: It works and yes it's more readable, done like this now

PR #1257, opened 2023-02-21
Fix time state to use ntpd in Leap

ktsamis:
## What does this PR change?

Fix time state to use ntpd in Leap instead of ntp. Partially reverts https://github.com/uyuni-project/sumaform/pull/1256 and fixes the issue for the controller in HEAD.


Bischoff (approved):
makes sense

nodeg (approved):

PR #1263, opened 2023-03-02
Adjust virthost for static nested VM hostname

nodeg:
## What does this PR change?

After adding a static hostname for the nested VM in the infrastructure repository, I need to make adjustments here, too.

- `min-nested` is the default name for nested minions. Furthermore we only need one default hostname.
- `MIN_NESTED` is the default environment variable that is exposed on the controller.


ktsamis (approved):
They look fine to me but I'm not sure if we miss any corner case. Looks good to merge to run a test PR run

nodeg:
Most of the work was  already done in https://github.com/uyuni-project/sumaform/pull/1197/files. This is just a small adjustment.

PR #1264, opened 2023-03-06
Make sumaform to use "venv-salt-call" for deployments if available and wait for "cloud-init" to finish

meaksh:
## What does this PR change?

This PR makes sumaform to use `venv-salt-call` if available instead of  `salt-call` for running the provisioning of the new instances. As we are moving to Salt Bundle, there are instances now where classic `salt-call` is not really available.

Addtionally, this PR introduces a mechanism to check that `cloud-init` is not still running before we start with the provisioning, as it was detected that sometimes `salt-call` execution starts with `cloud-init` still installing packages on the instance.

### DIFF
--- a/salt/first_deployment_highstate.sh
+++ b/salt/first_deployment_highstate.sh
@@ -5,6 +5,27 @@
 
 FILE_ROOT="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 
+# Force direct call module executors on MicroOS images
+MODULE_EXEC=""
+if grep -q "cpe:/o:.*suse:.*micro" /etc/os-release; then
+MODULE_EXEC="--module-executors=[direct_call]"
+fi
+
+# Wait for cloud-init to finish
+NEXT_TRY=0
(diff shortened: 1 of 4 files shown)


Bischoff (approved):
thanks

srbarrios (approved):

salt/first_deployment_highstate.sh
```diff
@@ -5,6 +5,18 @@
 
 FILE_ROOT="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 
+if [ -x /usr/bin/venv-salt-call ]; then
+    echo "Salt Bundle detected! We use it for running sumaform deployment"
+    echo "Copying /etc/salt/grains to /etc/venv-salt-minion"
```
cbbayburt, line 31: nitpick:
```suggestion
    echo "Copying /etc/salt/grains to /etc/venv-salt-minion/grains"
```

cbbayburt (approved):
LGTM :+1: 

just one small nitpick below

PR #1265, opened 2023-03-06
Do not use "/etc/salt/grains" as the place to deploy sumaform grains

meaksh:
## What does this PR change?

This PR is a continuation of https://github.com/uyuni-project/sumaform/pull/1264, in order to fix some issues:

- Grains are deployed after initial `wait_for_salt.sh` script runs, so grains are not properly captured in case of Salt Bundle.
- Sumaform deploys grains to `/etc/salt/grains` and this directory might not exist if there is only Salt Bundle.

In order to fix these issues, this PR does the following changes:

- sumaform grains are now set to `/tmp/grains` instead of directly to `/etc/salt/grains` (which might not exist)
- the deployment of `/tmp/grains` is now done before initial `wait_for_salt.sh` execution.
- the initial stage, `wait_for_salt.sh`, waits for cloud-init and then check the expected salt installed, then it copies grains to `/etc/salt/grains` or `/etc/venv-salt-minion/grains` accordingly.

So, deployment would run like this:

1) VM creation.
2) cloud-init starts execution.
3) Grains are deployed by sumaform to `/tmp/grains`
4) Sumaform executes `wait_for_salt.sh` (waits until cloudinit finished and set grains to corresponding `/etc/` path, and corresponding "salt-call" or "venv-salt-call" is available)
5) Sumaform executes `first_deployment_highstate.sh` (apply Salt states using `salt-call` or `venv-salt-call` accordingly)
6) Rest of execution.

### DIFF
--- a/backend_modules/aws/host/main.tf
+++ b/backend_modules/aws/host/main.tf
@@ -213,12 +213,6 @@ resource "null_resource" "host_salt_configuration" {
     destination = "/tmp"
   }
 
-  provisioner "remote-exec" {
-    inline = [
-      "bash /tmp/salt/wait_for_salt.sh",
-    ]
-  }
-
   provisioner "file" {
 
     content = yamlencode(merge(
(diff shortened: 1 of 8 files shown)


Bischoff (commented):
I get the idea and I approve it, but I'm unable to judge the correctness of the implementation. Some over review, please.

srbarrios (approved):

maximenoel8:
Deployment is still failing. From what I can understand this PR should fix it. I did some local testing with writing grain in /tmp and it works

cbbayburt (approved):

PR #1273, opened 2023-03-27
Fix for nested Leap 15.4 VM

nodeg:
## What does this PR change?

This will add the official Leap 15.4 repositories and install e.g. the `venv-salt-minion` package in order to not have to sync the complete Leap 15.4 repository in the test suite CI servers to be able to bootstrap this nested VM. Furthermore I added `nss-mdns` and  `qemu-guest-agent` like we do in sumaform.

### DIFF
--- a/salt/virthost/init.sls
+++ b/salt/virthost/init.sls
@@ -90,6 +90,7 @@ ifcfg-br0:
 
 ### adjustments for cloud init and the test suite ---
 # https://cloudinit.readthedocs.io/en/latest/topics/examples.html
+# We use openSUSE Leap 15.4 and SLES 15 SP4 as nested VMs
 rezise-{{ os_type }}-disk-image-template:
   cmd.run:
     - name: qemu-img resize /var/testsuite-data/{{ os_type }}-disk-image-template.qcow2 3G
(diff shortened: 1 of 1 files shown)


agraul:
Is the virthost state only used in the test suite or would this change apply to my sumaform deployment for development?

salt/virthost/init.sls
```diff
@@ -187,9 +188,20 @@ cloudinit-user-data-{{ os_type }}:
 {% if os_type == 'sles' %}
         # add SLES 15 SP4 base repository
         - zypper --non-interactive ar "http://download.suse.de/ibs/SUSE/Products/SLE-Module-Basesystem/15-SP4/x86_64/product/" SLE-Module-Basesystem15-SP4-Pool
+{% elif os_type == 'leap' %}
+        # add Leap 15.4 repositories
+        - zypper --non-interactive ar "http://download.opensuse.org/distribution/leap/15.4/repo/oss/" os_pool_repo
+        - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/oss/" os_update_repo
+        - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/sle/" sle_update_repo
+        - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/backports/" backports_update_repo
+        - zypper --non-interactive ar -p 98 "http://downloadcontent.opensuse.org/repositories/systemsmanagement:/Uyuni:/Master:/openSUSE_Leap_15-Uyuni-Client-Tools/openSUSE_Leap_15.0/" tools_pool_repo
```
vandabarata, line 197: honest question - don't these repos come automatically when having a Leap system? other than the last one I guess but in my own system, I seem to have the other ones and don't remember having had to add them manually

nodeg: Since this is a cloud-init image, the repositories were not automatically there. I did not see them with `zypper lr`.

ktsamis (approved):
In general I would like this to be written instead of a command to be executed as state files but I see the whole state is written like this so it's fine for now.

My only concern is that we use virthost anywhere else and we do repo manipulation via the testsuite so we will have duplicates but I doubt it.

vandabarata (approved):

PR #1277, opened 2023-03-30
Use HTTPS instead of HTTP in mirror files

nodeg:
## What does this PR change?

This will switch the URLs for the mirror files from HTTP to HTTPS. See See https://github.com/SUSE/spacewalk/issues/18738


Bischoff (commented):
Having the correct certificates can be a pain...

nodeg:
We discussed with Oscar and Eric that we will not do it. See https://suse.slack.com/archives/C02D12TNYLS/p1680508259998449

PR #1283, opened 2023-04-11
QE: Enable testing_overlay repo for prs CI

elariekerboull:
## What does this PR change?

This PR enables the use of `testing_overlay` repository for the PR CI.

### DIFF
--- a/salt/server/testsuite.sls
+++ b/salt/server/testsuite.sls
@@ -88,7 +88,7 @@ testsuite_salt_packages:
   pkg.installed:
     - pkgs:
       - salt-ssh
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
Hi! I am not sure this would be enough. Usually repos are mirrored into the jenkins worker, see for example:

https://github.com/SUSE/susemanager-ci/blob/master/jenkins_pipelines/environments/common/pipeline-pull-request.groovy#L189

We do this to avoid going through the openSUSE mirror.

However, I do not see any risk on trying it out ... so let me approve it and give it a try :)

jordimassaguerpla (approved):

PR #1287, opened 2023-04-14
Fix Uyuni build validation issues

nodeg:
## What does this PR change?

This will fix issues found during deploying our new Uyuni BV pipeline.

- add SLE Micro 5.4 to the minima configuration
- fix SLE Micro 5.x issues
- `salt-minion` vs. `venv-salt-minion` related issues
- Cobbler adjustments for the server

Before merging, I will squash and sort the commits according to the change.

Partially fixes https://github.com/SUSE/spacewalk/issues/16082

### DIFF
--- a/salt/default/time.sls
+++ b/salt/default/time.sls
@@ -1,3 +1,6 @@
+{% if not grains['osfullname'] == 'SLE Micro' %}
+# Dependencies already satisfied by the images
+# https://build.opensuse.org/project/show/systemsmanagement:sumaform:images:microos
 timezone_package:
   pkg.installed:
 {% if grains['os_family'] == 'Suse' %}
@@ -5,14 +8,17 @@ timezone_package:
 {% else %}
     - name: tzdata
 {% endif %}
+{% endif %}
 
 timezone_symlink:
   file.symlink:
     - name: /etc/localtime
     - target: /usr/share/zoneinfo/{{ grains['timezone'] }}
     - force: true
+{% if not grains['osfullname'] == 'SLE Micro' %}
     - require:
       - pkg: timezone_package
+{% endif %}
 
 timezone_setting:
   timezone.system:
@@ -59,9 +65,12 @@ ntp_enable_service:
 
 {% else %}
 
+{% if not grains['osfullname'] == 'SLE Micro' %}
+# Dependencies already satisfied by SLE Micro itself
 chrony_pkg:
   pkg.installed:
     - name: chrony
+{% endif %}
 
 chrony_conf_file:
   file.managed:
--- a/salt/minion/init.sls
+++ b/salt/minion/init.sls
@@ -11,7 +11,11 @@ include:
 # https://build.opensuse.org/project/show/systemsmanagement:sumaform:images:microos
 minion_package:
   pkg.installed:
+{% if grains['install_salt_bundle'] %}
+    - name: venv-salt-minion
+{% else %}
     - name: salt-minion
+{% endif %}
     - require:
       - sls: default
 {% endif %}
@@ -36,7 +40,11 @@ reload_systemd_modules:
 
 minion_id:
   file.managed:
+{% if grains['install_salt_bundle'] %}
+    - name: /etc/venv-salt-minion/minion_id
+{% else %}
     - name: /etc/salt/minion_id
+{% endif %}
     - contents: {{ grains['hostname'] }}.{{ grains['domain'] }}
 
 {% if grains.get('auto_connect_to_master') %}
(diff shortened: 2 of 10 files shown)


salt/minion/init.sls
```diff
@@ -62,3 +62,11 @@ minion_service:
...
       - file: master_configuration
 {% endif %}
+
+{% if grains['osfullname'] == 'SLE Micro' %}
+reboot:
+  module.run:
+    - name: system.reboot
+    - at_time: +2
+    - order: last
+{% endif %}
```
nodeg, line 72: I copied that over from the `virthost` minion.

aaannz: Yeah, 2 mins seems like too much for testsuite. I don't know if terraform monitors client liveness at this stage, but I think something like 5 to 10s would be better.

salt/default/time.sls
```diff
@@ -1,10 +1,14 @@
+{% if not grains['osfullname'] == 'SLE Micro' %}
```
aaannz, line 1: I don't have all the instances running, but is this same for all Micro systems?

Looking at release packages, for [SLE Micro 5.4](https://build.opensuse.org/package/view_file/SUSE:SLE-15-SP4:Update:Products:Micro54/000release-packages/SLE-Micro-release.spec?expand=1) and [SLE Micro 5.3](https://build.opensuse.org/package/view_file/SUSE:SLE-15-SP4:Update:Products:Micro53/000release-packages/SLE-Micro-release.spec?expand=1) we have `SLE Micro`. For [SLE Micro 5.2](https://build.opensuse.org/package/view_file/SUSE:SLE-15-SP3:Update:Products:MicroOS52/000release-packages/SUSE-MicroOS-release.spec?expand=1) and [SLE Micro 5.1](https://build.opensuse.org/package/view_file/SUSE:SLE-15-SP3:Update:Products:MicroOS51/000release-packages/SUSE-MicroOS-release.spec?expand=1) we have `SUSE MicroOS`

salt/minion/init.sls
```diff
@@ -62,3 +62,11 @@ minion_service:
     - watch:
       - file: master_configuration
 {% endif %}
+
+{% if grains['osfullname'] == 'SLE Micro' %}
```
aaannz, line 66: Same like about.

salt/repos/testsuite.sls
```diff
@@ -8,7 +8,11 @@ uyuni_key_for_fake_packages:
     - name: /tmp/uyuni.key
     - source: salt://default/gpg_keys/uyuni.key
   cmd.wait:
+{% if grains['osfullname'] == 'SLE Micro' %}
```
aaannz, line 11: Third time the charm. But maybe we do not have to check particularly for Micro OS name, but rather for `transactional` grain like we do in [bootstrap state](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls#L24)

salt/default/time.sls
nodeg: I can confirm the name for SLE Micro 5.2:

```bash
suma-bv-43-minssh-slemicro52:~ # cat /etc/os-release
NAME="SLE Micro"
VERSION="5.2"
VERSION_ID="5.2"
PRETTY_NAME="SUSE Linux Enterprise Micro 5.2"
ID="suse-microos"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:suse-microos:5.2"
```
I am not 100 % sure about 5.1 but I would assume it is the same as 5.2 since the renaming started only with 5.3.

salt/repos/testsuite.sls
nodeg: That would also be an option. Since @mbussolotto used the check from above, too I just copy pasted it over to my changes.

salt/minion/init.sls
nodeg: Ok, makes sense. I wonder why we stick with 2mins for the virhost, though since the value is also used in the test suite. Furthermore how can I change the value to e.g. 10s when the value is always minutes? I am not sure if a floating value like `+0.1` works.

salt/default/time.sls
aaannz: Then it's all fine by me and my commend can be ignored.

Btw. this prompted me to look more into the spec file and indeed, on lines 163 to 173 os-release with name `SUSE-MicroOS` is written, but then on line 179 it's seded and renamed to `SLE Micro` (🤦🏾) There was probably some reason to do this like this, but why?

salt/minion/init.sls
aaannz: `system.reboot` takes only minutes, you are right, but it's not mandatory to use just this module. This module just calls `shutdown -r` command. So instead we can use `cmd.run` state and call something like `sleep 10; shutdown -r`. Preferably in background so salt would fork and execute it so it does not interfere with running salt minion.

We use this construct in [saltboot to delay stopping salt minion](https://github.com/uyuni-project/retail/blob/master/saltboot-formula/_states/saltboot.py#L1905).

aaannz: Ah then `shutdown -r now` should be used.

nodeg: Yes, that would be a solution for that.

aaannz (approved):
LGTM

nodeg:
Rebase and squashing done.

PR #1288, opened 2023-04-20
Differentiate between Uyuni Master and Stable

nodeg:
## What does this PR change?

- Switch all EL9 clients to the Uyuni Stable client tools since they are now released
- Fix client tools for the Uyuni Proxy

### DIFF
--- a/salt/virthost/init.sls
+++ b/salt/virthost/init.sls
@@ -177,7 +177,7 @@ cloudinit-user-data-{{ os_type }}:
         - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/oss/" os_update_repo
         - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/sle/" sle_update_repo
         - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/backports/" backports_update_repo
-        - zypper --non-interactive ar -p 98 "http://downloadcontent.opensuse.org/repositories/systemsmanagement:/Uyuni:/Master:/openSUSE_Leap_15-Uyuni-Client-Tools/openSUSE_Leap_15.0/" tools_pool_repo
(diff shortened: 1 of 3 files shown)


vandabarata (approved):

salt/virthost/init.sls
```diff
@@ -177,7 +177,7 @@ cloudinit-user-data-{{ os_type }}:
         - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/oss/" os_update_repo
         - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/sle/" sle_update_repo
         - zypper --non-interactive ar "http://download.opensuse.org/update/leap/15.4/backports/" backports_update_repo
-        - zypper --non-interactive ar -p 98 "http://downloadcontent.opensuse.org/repositories/systemsmanagement:/Uyuni:/Master:/openSUSE_Leap_15-Uyuni-Client-Tools/openSUSE_Leap_15.0/" tools_pool_repo
+        - zypper --non-interactive ar -p 98 "http://downloadcontent.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/openSUSE_Leap_15-Uyuni-Client-Tools/openSUSE_Leap_15.0/" tools_pool_repo
```
juliogonzalez, line 180: Not saying this is wrong, but I can't tell if this is right or not. I don't know in what context this is used.

How would this work if we are testing code for master?

nodeg: This was implemented by me and is for the Salt Bundle migration tests. I was not aware that we should use Stable so I just used Master. With my knowledge know, I think Stable fits better here. It is basically only an openSUSE Leap 15.4 nested VM on the `virthost` that gets migrated from OS Salt to the Salt bundle.

juliogonzalez: ACK, then PR LGTM.

PR #1289, opened 2023-04-21
Add Leap 15.5 OS image

cbbayburt:
## What does this PR change?

**add description**


cbbayburt:
I'll update again when SLES15SP5 GA is available.

backend_modules/libvirt/base/main.tf
```diff
@@ -18,6 +18,8 @@ locals {
     opensuse153o    = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/distribution/leap/15.3/appliances/openSUSE-Leap-15.3-JeOS.x86_64-OpenStack-Cloud.qcow2"
     opensuse154o    = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/distribution/leap/15.4/appliances/openSUSE-Leap-15.4-JeOS.x86_64-OpenStack-Cloud.qcow2"
     opensuse154armo = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/distribution/leap/15.4/appliances/openSUSE-Leap-15.4-ARM-JeOS-efi.aarch64.qcow2"
+    opensuse155o    = "${var.use_mirror_images ? "http://${var.mirror}" : "https://download.opensuse.org"}/distribution/leap/15.5/appliances/openSUSE-Leap-15.5-Minimal-VM.x86_64-Cloud.qcow2"
```
nodeg, line 21: I would stick to the sorting and move the x64 Leap 15.5 image above the `opensuse154armo ` one so all x86 and arm stick together.

nodeg (changes requested):

rjmateus:
To be able to deploy minions with sles 15.5 you will also need to update the salt part, especially the repos files

cbbayburt:
> To be able to deploy minions with sles 15.5 you will also need to update the salt part, especially the repos files

Good point, I'm having some weird behavior after bootstrapping, so there might be even more to fix.

salt/repos/default.sls
```diff
@@ -407,13 +407,31 @@ os_update_repo:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Basesystem/15-SP4/x86_64/update/
     - refresh: True
 
+os_ltss_repo:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/SUSE/Updates/SLE-Product-SLES/15-SP4-LTSS/x86_64/update/
+    - refresh: True
```
nodeg, line 413: Should we already enable this? Is SP4 already considered LTSS?

salt/repos/virthost.sls
```diff
@@ -1,57 +1,31 @@
...
 {% elif grains['osrelease'] == '15.4' %}
+{% set sle_version_path = '15-SP4' %}
+{% elif grains['osrelease'] == '15.5' %}
+{% set sle_version_path = '15-SP5' %}
+{% endif %}
+
 module_server_applications_pool_repo:
   pkgrepo.managed:
-    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Server-Applications/15-SP4/x86_64/product/
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-Server-Applications/{{ sle_version_path }}/x86_64/product/
```
nodeg, line 21: Nice refactor. This removes much of the duplicated code :+1:

nodeg (approved):
Small comment. Otherwise it LGTM. Thank you Can!

salt/repos/default.sls
cbbayburt: Not yet, but the (empty) repository is online so I just added it. It's basically a *noop* for now, but we won't have to remember this again in the future.

nodeg: I see, ok. Thanks for the clarification.

PR #1294, opened 2023-05-01
Clean RH like minion repository

maximenoel8:
## What does this PR change?

`venv-salt` is not available in SUSE/Products. From my test SUSE/Updates is enough for our testing. 
Using SUSE/Product for Rocky 8 was failing my test because venv-salt in not available in Products repo.
I tried to simplify this block. Tell me if I'm missing something.


salt/repos/default.sls
```diff
@@ -487,29 +487,13 @@ os_cb_updates_repo:
...
-    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/{{ rhlike_client_tools_prefix }}/{{ release }}-CLIENT-TOOLS-BETA/x86_64/product/
-    {% else %}
-    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/{{ rhlike_client_tools_prefix }}/{{ release }}-CLIENT-TOOLS/x86_64/product/
-    {% endif %}
-    {% elif grains.get('mirror') %}
-    {% if 'beta' in grains.get('product_version') | default('', true) %}
-    - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/RES{{ release }}-SUSE-Manager-Tools-Beta/x86_64/
-    {% else %}
-    - baseurl: http://{{ grains.get("mirror") }}/repo/$RCE/RES{{ release }}-SUSE-Manager-Tools/x86_64/
-    {% endif %}
```
meaksh, line 501: IIUC I think we should not remove this block, as we would have problems then with RES7 when using a mirror. Maybe we want to refactor this instead.

From what I can see, the problem is on RES7 only, as even if https://download.suse.de/ibs/SUSE/Updates/RES/7-CLIENT-TOOLS/x86_64/update/x86_64/ exists, the "minima" mirrors we use in our CIs (which synchronize these channels from SCC), are still getting this other "$RCE" path synced:

```
minima-mirror:/srv/mirror # find .| grep RES | grep repomd.xml$
./repo/$RCE/RES7-SUSE-Manager-Tools/x86_64/repodata/repomd.xml
./repo/$RCE/RES7-SUSE-Manager-Tools-Beta/x86_64/repodata/repomd.xml
./SUSE/Products/RES/8-CLIENT-TOOLS/x86_64/product/repodata/repomd.xml
./SUSE/Updates/RES-CB/8/x86_64/update/repodata/repomd.xml
./SUSE/Updates/RES-AS/8/x86_64/update/repodata/repomd.xml
./SUSE/Updates/RES/8-CLIENT-TOOLS/x86_64/update/repodata/repomd.xml
./SUSE/Updates/RES/8/x86_64/update/repodata/repomd.xml
```

I have no idea why we are getting that one $RCE repo from SCC instead of the other ones, but with the current changes proposed on this PR, the RES7 instances would have errors as configured repositories won't be existing when using "minima mirrors".

meaksh (changes requested):

salt/repos/default.sls
srbarrios: If the issue is only in the mirror, and it's about a different path.
I propose to create a symbolic link that points to that /repo/$RCE folder and that would solve the problem (hacky but effective, I guess)

srbarrios (approved):

PR #1304, opened 2023-05-09
Upgrade to Terraform 1.4.6

cbbayburt:
Upgrade to [Terraform 1.4.6 (latest)](https://github.com/hashicorp/terraform/blob/v1.4/CHANGELOG.md#146-april-26-2023)

See [OBS build](https://build.opensuse.org/package/show/home:cbbayburt:branches:systemsmanagement:sumaform/terraform)


backend_modules/aws/db_host/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "1.0.10"
+  required_version = "1.4.6"
```
aaannz, line 2: I'm all for this, but can we at least allow minor versions updates (`~> 1.4.6`) if not anything in the future by using `>= 1.4.6`.

rjmateus (approved):
LGTM

I think we could also include the change of libvirt provider in this PR, and do all in one go.
more information in: 
https://github.com/uyuni-project/sumaform/pull/1023

backend_modules/aws/db_host/versions.tf
cbbayburt: AFAICS, there haven't been any breaking changes since the first stable release, so I'm also fine trying this again. Plus, as long as we keep building terraform in OBS, we're still in charge of what actual version the testsuite uses.

aaannz: Well there was one, where we had to change `cpu = { ... }` to `cpu { ... }` in our sumaform files.

rjmateus: but that was because of a libvirt plugin update, not a terraform update

aaannz: ah, thanks for clarifying.

nodeg:
You should probably update the PR. We are already at version 1.5.2 😉 https://github.com/hashicorp/terraform/blob/v1.5/CHANGELOG.md

cbbayburt:
Oh haha hard to catch up with them 😂

Last time I worked on this we had problems with Salt in the test suite so I had to leave it for the time. Now there are more pressing issues about sumaform so this is on hold for now.

nodeg:
Yes, this is a never ending story. On my personal machine I use the version Tumbleweed ships which is currently also 1.5.2, and I have no issues so far. But our SUMA setup is quite different. If you need help with testing this PR in the future, just ping me.

Bischoff:
If I am not wrong, the latest version with an open source license is 1.5.7. Maybe it should be our target for this PR.

Then we could take our time to switch to openTofu 1.6.0 or upper.

PR #1307, opened 2023-05-15
Replace the registry we must use for development versions of the Proxy container services

srbarrios:
## What does this PR change?

Related card: https://github.com/SUSE/spacewalk/issues/21238

Context:

We caught an issue in Uyuni CI, few hours before a submission. We found it thanks to a removal of images before building new ones in the Docker/PodMan Registry that we use in production for released RPMs.
So that let us know that we are using the wrong Namespace in the containerized proxy service configuration for our CI test environments (discovered in Uyuni, to be check in 4.3/Head), we are using the namespace for production with the old released images, not the new ones.

Task:
When deploying with Sumaform, change the configuration of the Containerized proxy systemd service, and use the correct NAMESPACE for the in-develop images.


ktsamis (approved):

nodeg (approved):

aaannz (approved):

salt/proxy/init.sls
```diff
@@ -84,6 +84,27 @@ proxy-container-packages:
...
+        - require:
+          - uyuni-proxy-systemd-services
+  {% elif '4.3' in grains.get('product_version') %}
+    proxy_substitute_suma_registry:
+      file.replace:
+        - name: /etc/sysconfig/uyuni-proxy-systemd-services
+        - pattern: registry.suse.com/suse/manager/4.3
+        - repl: registry.suse.de/devel/galaxy/manager/4.3/containers/suse/manager/4.3
+        - require:
+          - uyuni-proxy-systemd-services
```
aaannz, line 104: I already approved, but would this work with YAML? These states have wrong indentation wrt to other states.

srbarrios: Fixed. Thanks to catch it!

PR #1309, opened 2023-05-17
Do not install both venv-salt-minion and salt-minion

Bischoff:
## What does this PR change?

Clean up  cloud-init so that we do not install both the bundle and normal `salt-minion` packages.

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -28,7 +28,7 @@ yum_repos:
     name: epel
 
 %{ if install_salt_bundle }
-packages: ["venv-salt-minion", "salt-minion", "avahi", "nss-mdns", "qemu-guest-agent"]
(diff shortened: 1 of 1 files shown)


nodeg (approved):
LGTM. Thank you for taking care of it Eric!

meaksh (approved):

nodeg (approved):
LGTM.

meaksh (approved):

srbarrios (approved):

Bischoff:
The tests show, while the situation is cleaner right after sumaform has run, we still have the problem when the test suite runs:
```
[root@suma-42-min-centos7 ~]# grep master /etc/salt/minion.d/susemanager.conf /etc/venv-salt-minion/minion.d/susemanager.conf
/etc/salt/minion.d/susemanager.conf:master: suma-42-pxy.mgr.suse.de
/etc/venv-salt-minion/minion.d/susemanager.conf:master: suma-42-pxy.mgr.suse.de
```

Affects only 4.2 and only the RH-like minion.

PR #1312, opened 2023-05-24
Make "cloud-init" to setup a repo that provides "venv-salt-minion" package

meaksh:
## What does this PR change?

This PR makes `cloud-init` to configure corresponding Client Tools channels in SLE/openSUSE images, in order to have a configured repository which is able to provide `venv-salt-minion` package to run the deployment.

This should fix the current issue we see on our testsuites deployments: https://github.com/SUSE/spacewalk/issues/21545

I've tested deployments of SLE12SP4 and SLE15SP4/SP3 instances with these changes and they were successful.

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -123,6 +123,15 @@ packages: ["salt-minion", "avahi", "nss-mdns", "qemu-guest-agent"]
 %{ endif }
 
 %{ if image == "opensuse152o" }
+zypper:
+  # repo for salt
+  tools_pool_repo:
+    baseurl: http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Stable:/openSUSE_Leap_15-Uyuni-Client-Tools/openSUSE_Leap_15.0/
+    enabled: true
+    gpgcheck: false
+    name: tools_pool_repo
+    priority: 98
+
 %{ if install_salt_bundle }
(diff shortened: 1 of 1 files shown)


backend_modules/libvirt/host/user_data.yaml
```diff
@@ -152,6 +170,12 @@ zypper:
       baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-SERVER/12-SP4/x86_64/product/
       enabled: 1
       autorefresh: 1
+    - id: tools_pool_repo
+      baseurl: http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Stable:/SLE12-Uyuni-Client-Tools/SLE_12/
```
cbbayburt, line 174: I think it would be more accurate to use SUSE Manager Tools repos with SLES images (also with the others below).
```suggestion
      baseurl: http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/Devel:/Galaxy:/Manager:/4.3:/SLE12-SUSE-Manager-Tools/SLE_12/
```

cbbayburt (changes requested):

cbbayburt (approved):

agraul (approved):
As long as we have something after the "update packages" state, which updates Salt, I think the errors we have seen can still happen with clients that include Salt in the image. Not when we tell sumaform to install the bundle, that part probably works now (as long as there are no network hickups when cloud-init installs the bundle).

We can merge this now, but that shouldn't keep us from fixing the other parts that contribute to the problem.

PR #1322, opened 2023-06-07
Add openSUSE 15.5

Bischoff:
## What does this PR change?

Add openSUSE 15.5 for Build Validation (only ARM architecture)

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -868,3 +868,22 @@ packages: ["salt-minion"]
 runcmd:
   - zypper removerepo --all
 %{ endif }
+
+%{ if image == "opensuse155armo" }
+zypper:
+  repos:
+    - id: tools_pool_repo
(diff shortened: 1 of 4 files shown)


nodeg (approved):
LGTM.

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -868,3 +868,22 @@ packages: ["salt-minion"]
 runcmd:
   - zypper removerepo --all
 %{ endif }
+
+%{ if image == "opensuse155armo" }
+zypper:
+  repos:
+    - id: tools_pool_repo
+      baseurl: http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Stable:/openSUSE_Leap_15-Uyuni-Client-Tools/openSUSE_Leap_15.0/
```
ktsamis, line 876: shouldn't that be for openSUSE_Leap_15.5? I see only the openSUSE_Leap_15.0 is available in the mirror but it might be that it's not built/published for 15.5 or the mirror is not up to date. Can we confirm this with releng?

Bischoff: If I am not wrong, the tools are shared between all SPs

ktsamis: ok then we should be good to go

ktsamis (approved):

PR #1324, opened 2023-06-08
Add missing bv repo

maximenoel8:
## What does this PR change?

- Update user-data to use venv-salt ( salt-minion not available in some of the repositories )
- Update last repositories for ubuntu
- Add sle12 support to user-data
- Add SCC support to sle12 ( + add SUMA tools repo to minimum list for sle12 )


backend_modules/aws/host/user_data.yaml
```diff
@@ -98,6 +98,31 @@ zypper:
 
 packages: ["salt-minion"]
 %{ endif }
+
+%{ if image == "sles12sp4"}
+zypper:
+  repos:
+    - id: tools_pool_repo
+      baseurl: http://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/SLE12-Uyuni-Client-Tools/SLE_12/
```
meaksh, line 106: Shouldn't we need add this new repo also to `salt/mirror/etc/minimum_repositories_testsuite.yaml` file?

maximenoel8: This repositories are public so we don't need to sync them for AWS.

maximenoel8:
We currently mostly not using `mirror_url` in user_data ( only for obsolete version from what I can see ).  Is it something we should change @mcalmer ?

backend_modules/aws/host/user_data.yaml
srbarrios: Even if they are public, we need to create a zypper repository containing it in the host.

mcalmer:
> We are currently mostly not using `mirror_url` in user_data ( only for obsolete version from what I can see ). Is it something we should change @mcalmer ?

I think yes. In case we define to use the mirror, and cloud-init does not, we may get different versions of the packages then we expect.

PR #1328, opened 2023-06-12
Disable c3p0 connection timeout for now

nadvornik:
## What does this PR change?

c3p0 connection timeout is breaking the testsuite, it needs more fixes in java.
Let's disable it for now.


srbarrios (approved):

Bischoff (commented):
Could we please add a `# WORKAROUND: ....` comment?

PR #1331, opened 2023-06-14
Only remove only salt-client if this is a server

jordimassaguerpla:
## What does this PR change?
Only remove only salt-client if this is a server.

Otherwise, remove all salt packages.

This reverts commit 59104d3fa4f80ddcdda2182580f214b89367debe and adds a "is server" case.

### DIFF
--- a/salt/post_provisioning_cleanup.sh
+++ b/salt/post_provisioning_cleanup.sh
@@ -14,6 +14,17 @@ fi
 
 # Nothing to do in case "install_salt_bundle" grain is not true
 INSTALL_SALT_BUNDLE=$(${SALT_CALL} --local --log-level=quiet --output=txt grains.get install_salt_bundle)
+SERVER=false
+ROLES=$(${SALT_CALL} --local --log-level=quiet --output=txt grains.get roles)
+if [[ "${ROLES}" == "local: ['server']" ]];then
+    echo "This is a server"
+    SERVER=true
+    PKGS="Salt minion"
+else
+    echo "This is not a server"
+    PKGS="Salt"
+fi
+
 
 if [[ "$INSTALL_SALT_BUNDLE" != "local: True" ]]; then
     exit 0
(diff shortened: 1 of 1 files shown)


salt/post_provisioning_cleanup.sh
```diff
@@ -31,13 +42,25 @@ elif [ -x /usr/bin/apt ]; then
...
-yum -y remove salt salt-minion python3-salt python2-salt > /dev/null 2>&1 ||:
+    if [[ "$SERVER" == "true" ]];then
+        yum -y remove salt-minion > /dev/null 2>&1 ||:
+    else
+        yum -y remove salt salt-minion python3-salt python2-salt > /dev/null 2>&1 ||:
+    fi
 elif [[ "$INSTALLER" == "apt" ]]; then
-apt-get --yes purge salt-common > /dev/null 2>&1 ||:
+    if [[ "$SERVER" == "true" ]];then
+        apt-get --yes purge salt-client > /dev/null 2>&1 ||:
```
meaksh, line 60: I think it should be `salt-minion` here, right?

```suggestion
        apt-get --yes purge salt-minion > /dev/null 2>&1 ||:
```

salt/post_provisioning_cleanup.sh
```diff
@@ -14,6 +14,17 @@ fi
 
 # Nothing to do in case "install_salt_bundle" grain is not true
 INSTALL_SALT_BUNDLE=$(${SALT_CALL} --local --log-level=quiet --output=txt grains.get install_salt_bundle)
+SERVER=false
+ROLES=$(${SALT_CALL} --local --log-level=quiet --output=txt grains.get roles)
+if [[ "${ROLES}" == "local: ['server']" ]];then
```
meaksh, line 19: The `roles` variable is a actually a list, so in principle it could accept multiple values and therefore this comparision would fail.

Maybe something like:

```suggestion
if [[ "${ROLES}" == *"'server'"* ]];then
```
What do you think?

PR #1333, opened 2023-06-16
Adjust c3p0 connection timeout

nadvornik:
## What does this PR change?

This should be the final tuning for 3p0 connection timeout.

Manager 4.3 and uyuni is fixed to work with this setting. 
The workaround apparently did not work because of the default in sls file. Anyway, it is no longer needed.

The default timeout is set to 3 hours, because processing of RES7 repos on 4.2 takes up to 2 hours.


nadvornik:
fixed in https://github.com/uyuni-project/sumaform/pull/1334

PR #1334, opened 2023-06-20
fix default filter and increase default timeout to 4 hours

mcalmer:
## What does this PR change?

The default filter was used wrong. When the second parameter is true, it will use the default for
undefined, empty string and false .

Setting the grain to "false" result in "true" due to this filter.

Additionally 15 minutes timeout is too short for metadata generation on slow hardware and big repos like RES7.
Increased the default to 4 hours.


nadvornik:
LGTM

PR #1335, opened 2023-06-20
Prepare SLE micro properly

Bischoff:
## What does this PR change?

Most of SLE Micro setup was disabled because of some problems with package updates. This resulted for example in no hostname for SLE Micro 5.1 and 5.2. We can't disable all initialization at once.

The idea with SLE Micro is to prepare all packages in the image and having no package to update in sumaform - merely to check everything is there. However, `pkg.install` does not do the job because it is impossible to acquire a rpm lock  on a read-only filesystem.

Fixes https://github.com/SUSE/spacewalk/issues/21792

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -43,11 +43,10 @@ dbus_enable_service:
 # TODO: replace 'pkg.latest' with 'pkg.installed' when fix to bsc#1163683 is applied to all the SLES versions we use
 avahi_pkg:
 {% if grains['os_family'] == 'Suse' and grains['osfullname'] == 'SLE Micro'  %}
-# WORKAROUND for sle micro we should not ask for the latest by just check that is installed. We are building our image for sumaform.
-  pkg.installed:
+  cmd.run:
+    - name: transactional-update -c package up avahi avahi-lang libavahi-common3 libavahi-core7
 {% else %}
   pkg.latest:
-{% endif %}
     - pkgs:
       {% if grains['os_family'] == 'Debian' %}
       - avahi-daemon
@@ -68,6 +67,7 @@ avahi_pkg:
       - libavahi-core7
       {% endif %}
       {% endif %}
+{% endif %}
 
 # WORKAROUND: watch does not really work with Salt 2016.11
 avahi_dead_before_config:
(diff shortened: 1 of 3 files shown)


salt/default/avahi.sls
```diff
@@ -43,11 +43,10 @@ dbus_enable_service:
 # TODO: replace 'pkg.latest' with 'pkg.installed' when fix to bsc#1163683 is applied to all the SLES versions we use
 avahi_pkg:
 {% if grains['os_family'] == 'Suse' and grains['osfullname'] == 'SLE Micro'  %}
-# WORKAROUND for sle micro we should not ask for the latest by just check that is installed. We are building our image for sumaform.
-  pkg.installed:
+  cmd.run:
+    - name: transactional-update package up avahi avahi-lang libavahi-common3 libavahi-core7
```
nodeg, line 47: I wonder why we need this 2 times. Once here and the other time in `minimal.sls`?

Bischoff: oops, i probably forgot to change package names! nice catch !

F*** copy & paste errors....

salt/default/minimal.sls
```diff
@@ -1,7 +1,3 @@
-# WORKAROUND
-# This file should already be excluded from SLE Micro with the 
-# first few lines in salt/default/init.sls
-{% if not grains['osfullname'] == 'SLE Micro' %}
```
nodeg, line 4: You probably missed to remove the `endif` for that.

Bischoff: No, it's due to the way github shows diffs. The `endif` just stayed in place but now refers to another `if`. Please check carefully.

salt/default/minimal.sls
```diff
@@ -15,6 +11,14 @@ include:
...
 
 minimal_package_update:
+{% if grains['os_family'] == 'Suse' and grains['osfullname'] == 'SLE Micro'  %}
+  cmd.run:
+{% if grains['install_salt_bundle'] %}
+    - name: transactional-update package up zypper libzypp venv-salt-minion
+{% else %}
+    - name: transactional-update package up zypper libzypp salt-minion
+{% endif %}
+{% else %}
```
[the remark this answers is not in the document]
Bischoff: welcome to Jinja spaghetti coding.

The `endif` on line 20 matches the `if` on line 16.

The `else` on line 21 matches the `if` on line 14.

nodeg (approved):
LGTM, and you already mentioned you tested the code.

Bischoff:
> LGTM, and you already mentioned you tested the code.

i still need to check the bug with avahi-lang you reported does not reappear :-) .

PR #1338, opened 2023-06-22
Remove priority from user data AWS

maximenoel8:
## What does this PR change?

Remove priority for AWS user_data
This cloud init repositories have only venv-salt-minion version 3004 and so bock upgrade.

### DIFF
--- a/backend_modules/aws/host/user_data.yaml
+++ b/backend_modules/aws/host/user_data.yaml
@@ -159,7 +159,6 @@ yum_repos:
     enabled: true
     gpgcheck: false
     name: temp_tools_pool_repo
-    priority: 98
 
 packages: ["venv-salt-minion"]
 runcmd:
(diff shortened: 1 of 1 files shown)


Bischoff (approved):
yes we did it in `backend_modules/libvirt`, but overlooked `backend_modules/aws`

srbarrios (approved):
Make sense, but I would rather remove it from others too.
I don't know why we took the decision to have this priority:98 instead of just the default.
If we need to prioritize, we do, but if we want at the very end of priorities, well the default would be enough.

vandabarata (approved):

nodeg (approved):

Bischoff:
> Make sense, but I would rather remove it from others too. I don't know why we took the decision to have this priority:98 instead of just the default. If we need to prioritize, we do, but if we want at the very end of priorities, well the default would be enough.

I don't know why we wanted priorities at some point.

I just know that Michael said we only need them for `tools_additional_repo` and all the others priorities could be removed. See his PR 9e6e6aa4efb7bce29da012ef851791ffd3224203 .

PR #1345, opened 2023-07-05
Add EL9 GPG key

cbbayburt:
Adds EL9 GPG key to `repos.default` state


salt/repos/default.sls
```diff
@@ -535,6 +535,14 @@ tools_pool_repo:
     - require:
       - cmd: galaxy_key
 
+suse_el9_key:
```
ktsamis, line 538: you need to add a requires for this here https://github.com/uyuni-project/sumaform/pull/1345/files#diff-4bec81dc3a34f144c789099dd92e91b46249516d5907edd2899af940ef5acbc8R644

ktsamis (changes requested):

ktsamis (approved):

PR #1346, opened 2023-07-06
Enable EL9 key for all versions

cbbayburt:
https://github.com/uyuni-project/sumaform/pull/1345 added the keys only for `Manager`. This fixes it for other versions as well.

Fixes https://github.com/SUSE/spacewalk/issues/21955

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -448,6 +448,16 @@ install_recommends:
 
 {% set release = grains.get('osmajorrelease', None)|int() %}
 
+{# Soon this key will be used for all non-suse repos. When it happens, replace galaxy_key with this #}
+suse_el9_key:
+  file.managed:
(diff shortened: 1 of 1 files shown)


salt/repos/default.sls
```diff
@@ -448,6 +448,16 @@ install_recommends:
 
 {% set release = grains.get('osmajorrelease', None)|int() %}
 
+{# Soon this key will be used for all non-suse repos. When it happens, replace galaxy_key with this #}
```
nodeg, line 451: Nice. I did not know that.

nodeg (approved):
I tested the changes in https://ci.suse.de/view/Manager/view/Manager-qe/job/uyuni-master-qe-build-validation-NUE/43 and can confirm the issue is gone now. Rocky 8 is still failing because of the new GPG key, but this is expected.

PR #1349, opened 2023-07-11
fix uyuni-tools build

mbussolotto:
## What does this PR change?

Since `uyuni-tools` was refactored, we need to adapt `sumaform` to correctly build it


cbosdo (commented):
Avoid running that script from sumaform if possible: it pulls the golang BCI init image

mbussolotto:
ok, just creating `bin` folder seems works fine

cbosdo (approved):

PR #1353, opened 2023-07-20
Fix avahi installation

cbbayburt:
- Enable avahi from mirror if set up
- Make sure default repos are set up before installing avahi


srbarrios (approved):
It lgtm

nodeg (changes requested):

PR #1355, opened 2023-07-20
Disable auto_generate_bootstrap_repo when running the testsuite

maximenoel8:
## What does this PR change?

We want to be in control on what the server is doing during our testsuite to avoid any instability.
This PR disable the auto_generate_bootstrap_repo when testsuite is true on server side.
I reuse the testsuite variable to avoid adding new variable.
It tested the change and it works as expected.

### DIFF
--- a/salt/server/rhn.sls
+++ b/salt/server/rhn.sls
@@ -116,6 +116,17 @@ rhn_conf_c3p0_connection_debug_log:
 
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


mcalmer (approved):

srbarrios (approved):

nodeg (approved):
LGTM. Thank you Maxime!

PR #1356, opened 2023-07-24
Azure deployment image selection errors

rjmateus:
## What does this PR change?

Azure exit with errors when selecting the images available to deploy. Needs investigation on how to get those with the new department account.

We need to check:
- terraform azure provider version
- azure-cli tool supported version
- Image filtering (offer, publisher, etc)
- Azure account image subcription


maximenoel8:
I haven't used sumaform with Azure with our new okta account.
I tried to deploy opensuse-leap-15-4 manually. The search is still working. But we need to have the security type to standard otherwise it fails with Trusted Launch Visual Image.

PR #1357, opened 2023-07-30
Create a grains variable to disable auto generate bootstrap repo

maximenoel8:
## What does this PR change?

Create a specific variable to disable auto generate boostrap repo

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -69,6 +69,7 @@ module "server" {
   create_sample_bootstrap_script = false
   publish_private_ssl_key        = false
   disable_download_tokens        = false
+  disable_auto_bootstrap         = false
   forward_registration           = false
   monitored                      = true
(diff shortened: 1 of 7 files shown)


mcalmer:
But this allow only to globaly enable or disable generation of the bs-repo.
Is this what you want? Generate it in 4.3, but disable it in Head?

maximenoel8:
We only need to disable bootstrap during BV.

srbarrios (approved):
LGTM. If it is followed by a PR to set this as true in our BV environments.
And we assure a bs-repo regeneration after the custom channels for each OS are repo-sync too.

maximenoel8:
https://github.com/SUSE/susemanager-ci/pull/914

mcalmer:
> We only need to disable auto bootstrap during BV because we are calling manually mgr-create-bootstrap

Ahh, only for BV testsuite. This was the missing piece :-)

mcalmer (approved):

PR #1369, opened 2023-08-18
Add custom grain for CPE

HoussemNasri:
## What does this PR change?
Hi! I'm a GSoC contributor for Uyuni working on the OVAL-based CVE auditing project.

The primary goal of this PR is to be able to access minions CPEs from the backend in order to be able to conduct more accurate vulnerability scanning. With the CPE we can accurately identify the minion operating system and with that, we can decide what vulnerabilities affect the minion and what vulnerabilities do not.

Additional custom grain modules can be added to the `_grains` folder and propagated automatically to the host machines in subsequent terraform state applications. 

## Notes
- One problem that bugs me is that the custom grain modules are getting propagated to clients and servers even though we only need it on the server.
- Also, I'm unsure how to use [file_roots](https://docs.saltproject.io/en/latest/ref/configuration/master.html#std-conf_master-file_roots) instead of harcoding the path.

### DIFF
--- a/_grains/cpe_grain.py
+++ b/_grains/cpe_grain.py
@@ -1,0 +1,78 @@
+import re
+
+def get_cpe_grain(grains):
+    ret = {"cpe": ""}
+
+    os_release = _parse_os_release("/etc/os-release")
+
+    cpe = os_release["CPE_NAME"]
+    if cpe:
+        ret["cpe"] = cpe
+    else:
+        derived_cpe = _derive_cpe(grains)
+        if derived_cpe:
+            ret["cpe"] = derived_cpe
+
+    return ret
+
+# Copy-pasted from https://github.com/saltstack/salt/blame/master/salt/grains/core.py
+def _parse_os_release(*os_release_files):
+    """
+    Parse os-release and return a parameter dictionary
+
+    This function will behave identical to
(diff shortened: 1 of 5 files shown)


mcalmer:
I think sumaform is the wrong place. Sumaform is only used for testing. On production server this would not be available.
But i think you want this for all uyuni managed clients, right?

For this you better add this to the main uyuni repository under susemanager-utils/susemanager-sls/src/grains/ .
These grains will be deployed to the client during registration to uyuni server.

HoussemNasri:
Ah cool, that's actually easier than I thought.

Just one thing I need to ask you. In the backend, I could set the CPE of the server once it's registered in`RegisterMinionEventMessageAction`, but how about already registered servers? First, I need to call `saltutil.sync_grains` on all clients so they update their grains, which I can do by calling the `saltApi.syncGrains`, and then getting the grains through `saltApi` too, but what is the best time to do these operations? In other words, what is the best migration strategy to assign servers a CPE?

mcalmer:
Make it part of package list refresh action.
This is part of the highstate and if often executed often after a package change on the client.

CPE is a software thing, so it belongs to it.

HoussemNasri:
Thanks, I'll close this then.

PR #1372, opened 2023-08-22
Fix Leap 15.4 CI images not using a mirror

nodeg:
This will

- fix the alignment of the image variables
- fix not using a mirror in case of the openSUSE Leap 15.5 CI images


Bischoff (approved):
lgtm, but beware that other backend modules than libvirt could also have there variables

nodeg:
I checked that and the `libvirt` backend is the only one having those 2 variables.

PR #1383, opened 2023-09-04
Add kiwi packages to buildhost

nodeg:
## What does this PR change?

Since we kill the openSUSE Leap reposync in the test suite, we have to install the kiwi package in order to be able to successfully bootstrap the buildhost. We also see failures regarding this in our 4.3 CI.

See https://github.com/SUSE/spacewalk/issues/22440

### DIFF
--- a/salt/build_host/init.sls
+++ b/salt/build_host/init.sls
@@ -7,6 +7,13 @@ certificate_authority_certificate:
     - source: salt://build_host/certs/ca.cert.pem
     - makedirs: True
 
+{% if '4.2' not in grains.get('product_version') %}
+cucumber_requisites:
+  pkg.installed:
+    - pkgs:
+      - python3-kiwi
+{% endif %}
+
 {% if '11' in grains['osrelease'] %}
 
 update_ca_truststore_registry_build_host:


vandabarata (approved):

Bischoff (approved):

aaannz (approved):
LGTM

salt/build_host/init.sls
```diff
@@ -7,6 +7,13 @@ certificate_authority_certificate:
     - source: salt://build_host/certs/ca.cert.pem
     - makedirs: True
 
+{% if '4.2' not in grains.get('product_version') %}
+cucumber_requisites:
```
srbarrios, line 11: Having packages matching the word "kiwi" in the repositories of the BuildHost Minion.
```
{%- set available_packages = salt['pkg.search']('kiwi').keys() %}
```
But this doesn't mean we need to install the package in advance, as the installation is made few lines below in the salt states of `kiwi-image-server.sls` service:
```
mgr_install_kiwi:
  pkg.installed:
    - pkgs:
{%- if use_kiwi_ng %}
      - python3-kiwi
{%- else %}
      - kiwi
{%- endif %}
{%- for km in kiwi_modules %}
{%-   if km in available_packages %}
      - {{ km }}
{%-   endif %}
{%- endfor %}
```

The way I would have fixed this is by injecting the missing repository in Build Host VM, that must contain packages matching `kiwi` word.

And another remark, I would not use `cucumber_requisites` as salt name, because the fact is that wiki is a dependency for the Build Host onboarding, so I would choose `kiwi_imager_server_dependency`.

aaannz: As far as I understood, the problem is that there is no repo to inject, unless you want to go outside Uyuni provided repos. Because reposync is not allowed to finish for Leap, repositories provided by Uyuni are empty.

A agree with the naming and maybe note that this is a hack because of the way tests are preformed. Under normal operation, these deps are provided by synced channels.

srbarrios: @aaannz Before we decided to synchronize channels from scc and wait for them to be synchronized, we were injected through sumaform repositories pointing directly to the content, I don't know the exact URL now, but I think it should exist one.

srbarrios: Actually, you can check that there is no repository mirrored for Leap here:
http://minima-mirror.mgr.suse.de/SUSE/Updates/SLE-Module-Development-Tools/

Where we suppose to get Development channel for Leap.

PR #1396, opened 2023-09-21
fix bootstrap path in container

mbussolotto:
## What does this PR change?

fix bootstrap path in container


mbussolotto:
not required

PR #1404, opened 2023-10-03
Introduce a new libvirt module for testing Salt: "salt_testenv"

meaksh:
## What does this PR change?

This PR creates a new module for libvirt backend in sumaform that takes care of deploying machines and preparing a Salt testing environment, with all necessary test dependencies, in order to run the Saltstack "unit/integration/functional" tests.

Tracks: https://github.com/SUSE/spacewalk/issues/21165

## Workflow
1. Salt Bundle is used to run the sumaform deployment for "salt_testenv" nodes. As some OSes don't have classic package available anymore.
2. Depending on the selected OS image, the classic Salt package and/or Salt Bundle will be deployed for the testing:
    - Classic Salt package (python3-salt + python3-salt-testsuite)
      * SLE15 family (all SPs)
      * openSUSE Leap
    - Salt Bundle package (venv-salt-minion + venv-salt-minion-testsuite)
      * SLE15 family (all SPs)
      * SLE12 family
      * openSUSE Leap
      * AlmaLinux 8/9
      * CentOS 7
      * Debian 9/10/11
      * Ubuntu 20.04 / 22.04
3. The different Salt flavors can be selected for testing (i.a "products", "products:testing" or "products:next"), by using the new `salt_obs_flavor` variable.
4. After deployment is done, `pytest` should be able to run the deployed tests.

## Example of use

You can simply add `salt_testenv` nodes in your `main.tf` like this:
```
module "test-salt-sles15sp1" {
  source             = "./modules/libvirt/salt_testenv"

  name               = "test-salt-sles15sp1"
  image              = "sles15sp1o"
  salt_obs_flavor    = "products:next"
}

module "test-salt-alma8" {
  source             = "./modules/libvirt/salt_testenv"

  name               = "test-salt-alma8"
  image              = "almalinux8o"
  salt_obs_flavor    = "products:next"
}
```

**IMPORTANT:** At the time of creating this PR, since some of the necessary "testsuite" subprojects for the Salt Bundle are not yet fully ready at OBS, you might need to use `salt_obs_flavor = "products:next"` when deploying nodes that are going to use the Salt Bundle in order to prevent repository issues during deployment. As in the example above.

### Classic Salt:
```console
# cd /usr/lib/python3.6/site-packages/salt-testsuite
# pytest -vvvv tests/unit tests/pytests/unit/ ...
```
### Salt Bundle:
```console
# cd /usr/lib/venv-salt-minion/lib/python3.10/site-packages/salt-testsuite/
# ../../../../bin/pytest -vvvv tests/unit tests/pytests/unit/ ...
```


salt/salt_testenv/test_repos.sls
```diff
@@ -0,0 +1,108 @@
...
+    {% if grains['osrelease_info'][0] == 8 %}
+        {% set repo_path = 'AlmaLinux_8' %}
+    {% elif grains['osrelease_info'][0] == 9 %}
+        {% set repo_path = 'AlmaLinux_9' %}
+    {% endif %}
+{% elif grains['osfullname'] == 'CentOS Linux' %}
+    {% if grains['osrelease_info'][0] == 7 %}
+        {% set repo_path = 'CentOS_7' %}
+    {% endif %}
+{% elif grains['osfullname'] == 'SLES' %}
```
vzhestkov, line 73: Not sure if I understand the calculation for SLE15 repo right, but sounds like `_SPx` is possible in the calculation made before, but for the salt bundle repos we don't have separate repos for SPs. maybe it makes sense to set SLE_15 repo here explicitly as for SLE 12?

meaksh: Thanks for having a look here!

So, in this initial iteration of the implementation, we do not test Salt Bundle in SLE15 (or any SPs). Hence, for openSUSE/SLE15 only classic package.

That is why we are not taking care at all of the Salt Bundle for SLE15 in this calculation here.

Hth!

meaksh: UPDATE: as discussed in the standup, I've just adapted the logic to install both classic and bundle packages for SLE15 / openSUSE distros.

salt/salt_testenv/test_repos.sls
```diff
@@ -0,0 +1,113 @@
...
+    - refresh: True
+    - gpgcheck: 0
+    - gpgkey: http://{{ grains.get("mirror") | default("download.opensuse.org", true) }}/repositories/systemsmanagement:saltstack:{{ grains["salt_obs_flavor"] }}/{{ repo_path }}/repodata/repomd.xml.key
+
+install_salt_testsuite:
+  pkg.installed:
+    - name: python3-salt-testsuite
+    - require:
+      - pkgrepo: salt_testsuite_dependencies_repo
+      - pkgrepo: salt_testing_repo
```
agraul, line 43: The state file is called `test_repos`, but this state installs the package itself. Can this be refactored? Maybe move this (or even the whole file?) into the init.sls? Or rename the file?

salt/salt_testenv/test_repos.sls
```diff
@@ -0,0 +1,113 @@
...
+    - gpgkey: http://{{ grains.get("mirror") | default("download.opensuse.org", true) }}/repositories/systemsmanagement:saltstack:{{ salt_flavor_path }}:testsuite/{{ repo_path }}/repodata/repomd.xml.key
+    - gpgcheck: 0
+{% endif %}
+    - refresh: True
+
+install_salt_bundle_testsuite:
+  pkg.installed:
+    - name: venv-salt-minion-testsuite
+    - require:
+      - pkgrepo: salt_bundle_testsuite_repo
```
agraul, line 113: Same question about refactoring applies here.

agraul (changes requested):
Great work, looks pretty good! I especially like the `salt_obs_flavor` variable to easily control what's used.

I just left a comment about the organization and re-raised Victor's question about the Bunlde and SLE/Leap. Once that's clarified we're good to merge IMO.

salt/salt_testenv/test_repos.sls
meaksh: For the Salt Bundle, and SLE/Leap, we use one repo-for-bundle-testsuite per OS major version, so all SPs goes to the same repo: `SLE_15`.

As we have the repositories configured at: https://build.opensuse.org/project/show/systemsmanagement:saltstack:bundle:next:testsuite

agraul (approved):

agraul:
Looking at the commit history, what's your plan @meaksh?

PR #769, opened 2020-09-09
Update libvirt readme

Etheryte:
Currently, there's no information about what `cc_username` and `cc_password` stand for. Without knowing the context, it's difficult to search for and figure out what they mean. This PR adds a small reference to contextualize the terms.

## What does this PR change?

This PR updates the libvirt configuration readme with background context.

PR #1405, opened 2023-10-04
fix(README): Document prerequisites and common variables for deployment

m-czernek:
## What does this PR change?

In this PR, I am documenting:

* Libvirt prerequisites: this seemed reasonable to me since the section also contains description of installing the `terraform-provider-libvirt` package
* Variables: a newcomer to Uyuni probably won't know what `CC` stands for. Additionally, the `images` section might be helpful

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,22 @@
 
 **Libvirt provider version**: 0.6.3
 
+### Libvirt Provider Prerequisites
+
+If you use the `Libvirt` provider, install and enable `libvirt` before you attempt to run the terraform deployment.
+The `virt-manager` package is recommended because it configures default resources that the terraform deployment uses, e.g. the `default` virtual network.
+
+```bash
+# Download and install libvirt and virt-manager, for example:
+sudo zypper install libvirt virt-manager
+
+# Start libvirt
+sudo systemctl start libvirtd
+
+# Optionally, enable libvirt so that it starts at boot time
+sudo systemctl enable libvirtd
+```
+
 openSUSE and SUSE Linux Enterprise Server:
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,23 @@
...
 
-openSUSE and SUSE Linux Enterprise Server:
+### Prerequisites
+
+If you use the `Libvirt` provider, install and enable `libvirt` before you attempt to run the terraform deployment.
+The `virt-manager` package is recommended because it configures default resources that the terraform deployment uses, e.g. the `default` virtual network.
+
+```bash
+# Download and install libvirt and virt-manager, for example:
+sudo zypper install libvirt virt-manager
```
meaksh, line 21: It seems this line is SUSE/openSUSE specific, but we are in the general prerequisite section (which could apply also to Debian/Ubuntu).

Can we maybe make this more generic or maybe include this part in the "openSUSE / SUSE" section? What do you think?

meaksh (commented):
Thanks a lot for these improvements! :tada:

I just made some suggestions that maybe you could consider. Thanks in advance!

agraul:
There is also https://github.com/uyuni-project/sumaform/pull/769 which covers some of the same changes. Maybe we should also take a look at that one.

README.md
m-czernek: Changed to:

a) move to a better (IMHO) location
b) include Debian/Ubuntu code example

m-czernek:
@agraul I think this PR supersedes changes in #769 , doesn't it? IMO the `cc_*` variables are necessary for any backend, so it probably doesn't make sense to document them in the libvirt module. Perhaps we can delete the current `cc` docs in https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/README.md? 

```
complete the cc_password variable in the base module
```

PR #1410, opened 2023-10-13
We forgot 4.3-released case

Bischoff:
## What does this PR change?

While reviewing sumaform's code with Julio, we noticed those two places where `4.3-released` case has apparently been forgotten. This PR fixes that.

**Note:** to avoid this problem in the future, it could be a good idea to refactor, testing `contains(..., '-released')` to cover all `4.x-released` cases, while taking care of the fact that `uyuni-released` also exists.

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -720,6 +720,8 @@ tools_update_repo:
 {% set tools_repo_url = 'http://' + grains.get("mirror") | default("download.suse.de/ibs", true) + '/SUSE/Updates/Ubuntu/' + release + '-CLIENT-TOOLS/x86_64/update/' %}
 {% elif '4.2-released' in grains.get('product_version') | default('', true) %}
 {% set tools_repo_url = 'http://' + grains.get("mirror") | default("download.suse.de/ibs", true) + '/SUSE/Updates/Ubuntu/' + release + '-CLIENT-TOOLS/x86_64/update/' %}
+{% elif '4.3-released' in grains.get('product_version') | default('', true) %}
+{% set tools_repo_url = 'http://' + grains.get("mirror") | default("download.suse.de/ibs", true) + '/SUSE/Updates/Ubuntu/' + release + '-CLIENT-TOOLS/x86_64/update/' %}
 {% elif 'uyuni-master' in grains.get('product_version') | default('', true) %}
 {% set tools_repo_url = 'http://' + grains.get("mirror") | default("downloadcontent.opensuse.org", true) + '/repositories/systemsmanagement:/Uyuni:/Master:/Ubuntu' + short_release + '-Uyuni-Client-Tools/xUbuntu_' + release %}
 {% else %}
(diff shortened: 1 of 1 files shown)


nodeg (approved):

juliogonzalez (approved):
LGTM. I agree with Eric's idea about the refactor, and I agree that should be a separate PR so it can be tested and verified carefully. Maybe there was a good reason we did it this way (I can't remember)

vandabarata (approved):

salt/repos/default.sls
```diff
@@ -752,6 +754,11 @@ tools_update_repo_raised_priority:
             Package: *
             Pin: release l=SUSE:Updates:Ubuntu:{{ release }}-CLIENT-TOOLS:x86_64:update
             Pin-Priority: 800
+{% elif '4.3-released' in grains.get('product_version') | default('', true) %}
+    - contents: |
+            Package: *
+            Pin: release l=SUSE:Updates:Ubuntu:{{ release }}-CLIENT-TOOLS:x86_64:update
+            Pin-Priority: 800
```
srbarrios, line 761: Stupid idea, if we will have exact same content in all "4.x-released" versions... we can try to have a unique elif statement for all of them.

srbarrios: Shit, it happened again, we had the same idea again, argh!

srbarrios (approved):

salt/repos/default.sls
Bischoff: :smile: no worries

juliogonzalez: No, that won't fly. We'll release SUSE Manager 5.0, and this won't work.

Right solution is what was proposed above, having first an elif for `uyuni-released` that adds the Uyuni Stable client tools, then another elif for `-released` that adds the SUSE Manager released client tools.

Seems that's what we do for other client tools, but can't tell for sure why we didn't do it for Ubuntu.

Bischoff:
Waiting for Can's approval to merge.

salt/repos/default.sls
srbarrios: YAGNI, when we have 5.0, we think about it.

juliogonzalez: We are going to have the first Alpha ~next month. So we are going to need this soon. It doesn't make sense to do a refactor now, to do another one in two months.

srbarrios: Rely on executing the proper if statement on the order we code the if statements doesn't seem a good practice to me.
Ultimately, I would rather use:
{% elif ('4.' in grains.get('product_version') or '5.' in grains.get('product_version')) and '-released' in grains.get('product_version') | default('', true) %}

juliogonzalez: Not saying it's a good practice, but AFAIK that's what we do in the other cases. 

So either we refactor the others, or we refactor this to do the same, but we better don't have exceptions doing different things if there are no good reasons for that.

But anyway @cbbayburt will decide when he checks :-)

srbarrios: Could you give an example of these other cases? I can't see anything similar.

juliogonzalez: That's what it seems Eric saw, as I asked him to review the other client tools. Maybe he was confused.

In the end all that matters that whatever we'll do, we try to standardize. That's the main point :-)

As I mentioned, Can will need to have a look and in the end propose a solution.

cbbayburt (approved):

PR #1413, opened 2023-10-16
Add SLE-HPC-Module to Salt Testenv 

agraul:
## What does this PR change?
We need libhttp_parser_2_7_1, which is a dependency of libgit2. It's available in our SUMA Server channel, but that's only for a single SLE SP. Since it's also in the HPC Module, we can use that repo.

### DIFF
--- a/salt/salt_testenv/init.sls
+++ b/salt/salt_testenv/init.sls
@@ -13,6 +13,17 @@ development_tools_repo_updates:
   pkgrepo.managed:
     - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-Development-Tools/{{ repo_path }}/x86_64/update/
     - refresh: True
+
+# needed for libhttp_parser_2_7_1, dependency of libgit2
+sle_module_hpc_repo_pool:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-HPC/{{ repo_path }}/x86_64/product/
+    - refresh: True
+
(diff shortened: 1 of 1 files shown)


meaksh (approved):
Thanks, LGTM :+1:

salt/salt_testenv/init.sls
```diff
@@ -13,6 +13,17 @@ development_tools_repo_updates:
...
+
+# needed for libhttp_parser_2_7_1, dependency of libgit2
+sle_module_hpc_repo_pool:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Products/SLE-Module-HPC/{{ repo_path }}/x86_64/product/
+    - refresh: True
+
+sle_module_hpc_repo_updates:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-HPC/{{ repo_path }}/x86_64/product/
```
meaksh, line 25: oops, there is a typo on the URL here:
```suggestion
    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de/ibs", true) }}/SUSE/Updates/SLE-Module-HPC/{{ repo_path }}/x86_64/update/
```

meaksh (changes requested):
I just realized of a typo on the update repo url

salt/salt_testenv/init.sls
agraul: Oops, copy paste fail :facepalm:. Thanks for catching it! I "tested" the state by setting up a salt_testenv node and the deployment installed the dependency correctly. I should have checked that the repos are correct as well.

meaksh (approved):

PR #1422, opened 2023-11-15
Remove useless repository for RES 8 and add ubuntu2204 tools repo

maximenoel8:
## What does this PR change?
To run the BV tests on AWS, we need the minion tool repositories. This list is the minimum you need to run the tests on AWS.
Update the minimum repository list to run BV on AWS with ubuntu2204 tool repo

### DIFF
--- a/salt/mirror/etc/minimum_repositories_testsuite.yaml
+++ b/salt/mirror/etc/minimum_repositories_testsuite.yaml
@@ -7,6 +7,9 @@ http:
   - url:  http://download.suse.de/ibs/SUSE/Updates/Ubuntu/20.04-CLIENT-TOOLS/x86_64/update/
     archs: [x86_64]
 
+  - url: http://download.suse.de/ibs/SUSE/Updates/Ubuntu/22.04-CLIENT-TOOLS/x86_64/update/
+    archs: [ x86_64 ]
+
   - url:  http://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/Ubuntu2004-Uyuni-Client-Tools/xUbuntu_20.04/
     archs: [x86_64]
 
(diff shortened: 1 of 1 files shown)


salt/mirror/etc/minimum_repositories_testsuite.yaml
```diff
@@ -46,9 +49,6 @@ http:
   - url:  http://download.suse.de/ibs/SUSE/Updates/RES/8-CLIENT-TOOLS/x86_64/update/
     archs: [x86_64]
 
-  - url:  http://download.suse.de/ibs/SUSE/Products/RES/8-CLIENT-TOOLS/x86_64/update/
-    archs: [x86_64]
-
```
srbarrios, line 51: If a package doesn't have any update, then the package will be only available in the Pool but not in the Updates.
Therefore, I'm not fully sure if we can rely only on Updates folder for RES8.

deneb-alpha: the point is more that the used URL is wrong...
http://download.suse.de/ibs/SUSE/Products/RES/8-CLIENT-TOOLS/x86_64/update/ -> this is a 404

the pool packages are at https://download.suse.de/ibs/SUSE/Products/RES/8-CLIENT-TOOLS/x86_64/product/

srbarrios: I see, so maybe the correct it will be to edit this line as:
`url:  http://download.suse.de/ibs/SUSE/Products/RES/8-CLIENT-TOOLS/x86_64/product/`

maximenoel8: This file is just a solution to have access to the default repositories needed for BV testing.
I just deploy BV on AWS using terraform and copy / paste the missing repositories in this list. The changes you are talking about are more related to salt/repos.

PR #1409, opened 2023-10-12
Debian 12 fixes

Bischoff:
## What does this PR change?

let debian 12 deploy in 4.3.9 build validation

PR #1406, opened 2023-10-05
Add support for Debian 12 clients

cbbayburt:
Adds support for Debian 12 clients.

PR #1423, opened 2023-11-19
Switch to Debian 12 client tools Stable repo

cbbayburt:
Follow-up to #1406 and #1409. Switch Debian 12 client tools repo from Master to Stable.


deneb-alpha (approved):
LGTM, thanks!

PR #1426, opened 2023-11-23
Add support to paygo images on controller

maximenoel8:
## What does this PR change?

Add the possibility to test paygo images on controller. 
To do so, add new bashrc variable related to paygo image

### DIFF
--- a/modules/controller/main.tf
+++ b/modules/controller/main.tf
@@ -60,10 +60,15 @@ module "controller" {
     server_http_proxy         = var.server_http_proxy
     custom_download_endpoint  = var.custom_download_endpoint
     pxeboot_image             = var.pxeboot_configuration["image"]
+    is_using_paygo_server     = var.is_using_paygo_server
     is_using_build_image      = var.is_using_build_image
     is_using_scc_repositories = var.is_using_scc_repositories
+    server_instance_id        = var.server_instance_id
     container_runtime         = lookup(var.server_configuration, "runtime", null)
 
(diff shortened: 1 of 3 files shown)


modules/controller/main.tf
```diff
@@ -60,10 +60,15 @@ module "controller" {
...
     pxeboot_image             = var.pxeboot_configuration["image"]
+    is_using_paygo_server     = var.is_using_paygo_server
     is_using_build_image      = var.is_using_build_image
     is_using_scc_repositories = var.is_using_scc_repositories
+    server_instance_id        = var.server_instance_id
     container_runtime         = lookup(var.server_configuration, "runtime", null)
 
+    sle12paygo_minion    = length(var.sle12paygo_minion_configuration["hostnames"]) > 0 ? var.sle12paygo_minion_configuration["hostnames"][0] : null
+    sle15paygo_minion    = length(var.sle15paygo_minion_configuration["hostnames"]) > 0 ? var.sle15paygo_minion_configuration["hostnames"][0] : null
+    sappaygo_minion      = length(var.sappaygo_minion_configuration["hostnames"]) > 0 ? var.sappaygo_minion_configuration["hostnames"][0] : null
```
srbarrios, line 71: It might be good to add more info on the sap one, maybe something like `sles4sap_paygo_minion` ?
Also for sle12 and sle15, I think it can make sense to split the paygo part, something like `sle12_paygo_minion`. Make sense for you?

maximenoel8: I don't know if we have any mechanisms around splitting the name with _ in two parts.

srbarrios (approved):

PR #1428, opened 2023-11-24
Add paygo product version condition to skip useless repositories

maximenoel8:
## What does this PR change?

Paygo instances on cloud ( testing with AWS ) already have the repositories by default.
Also in cloud, I try to avoid mirroring public repositories.
Using internal product repositories doesn't make sense for paygo testing.
Nevertheless, we still need the testsuite repositories.
So for server, I disable any repo calls and for minion, I only keep respo.testsuite call

I also removed a redundant repo call in testsuite.


salt/default/testsuite.sls
```diff
@@ -3,7 +3,6 @@
 
 include:
   - scc
-  - repos
```
srbarrios, line 6: Shouldn't we have here this?
```
- repos.testsuite
```

salt/minion/init.sls
```diff
@@ -1,6 +1,6 @@
 include:
   - scc.minion
-  {% if 'build_image' not in grains.get('product_version') | default('', true) %}
+  {% if 'build_image' and 'paygo' not in grains.get('product_version') | default('', true) %}
```
srbarrios, line 3: Can we check two values at same time?
I think it should be:
```suggestion
  {% if 'build_image' not in grains.get('product_version') and 'paygo' not in grains.get('product_version') | default('', true) %}
```

salt/minion/testsuite.sls
```diff
@@ -24,7 +24,7 @@ suse_minion_cucumber_requisites:
     - pkgs:
       - aaa_base-extras
       - ca-certificates
-    {% if 'build_image' not in grains.get('product_version') | default('', true) %}
+    {% if 'build_image' and 'paygo' not in grains.get('product_version') | default('', true) %}
```
srbarrios, line 27: Same than above.

salt/server/init.sls
```diff
@@ -74,7 +77,7 @@ ssl-building-ca-configuration:
 {% endif %}
 
 
-{% if '4' in grains['product_version'] and grains['osfullname'] != 'Leap' and not grains.get('server_registration_code') and 'build_image' not in grains.get('product_version') %}
+{% if '4' in grains['product_version'] and grains['osfullname'] != 'Leap' and not grains.get('server_registration_code') and 'paygo' not in grains.get('product_version') %}
```
srbarrios, line 80: Do we really want to replace the value, or to add an additional comparison?
Do we maybe want this?
```suggestion
{% if '4' in grains['product_version'] and grains['osfullname'] != 'Leap' and not grains.get('server_registration_code') and 'build_image' not in grains.get('product_version') and 'paygo' not in grains.get('product_version') %}
```

salt/repos/default.sls
```diff
@@ -340,7 +340,7 @@ os_ltss_repo:
 
 {% endif %} {# '15' == grains['osrelease'] #}
 
-{% if '15.1' == grains['osrelease'] and not ( grains.get('server_registration_code') or grains.get('proxy_registration_code') or grains.get('sles_registration_code')) %}
+{% if '15.1' == grains['osrelease'] and not ( grains.get('server_registration_code') or grains.get('proxy_registration_code') or grains.get('sles_registration_code') or 'paygo' not in grains.get('product_version') | default('', true)) %}
```
srbarrios, line 343: I don't think we need to default to true here. The product_version grain is always present, so it should be fine.

salt/server/init.sls
```diff
@@ -19,6 +19,8 @@ include:
   - server.salt_master
   - server.tcpdump
 
+
+{% if 'paygo' not in grains.get('product_version') | default('', true) %}
```
srbarrios, line 23: We don't need to default to true, for product_version grain.

salt/server/testsuite.sls
```diff
@@ -37,14 +37,14 @@ test_repo_debian_updates:
...
-      {% if 'build_image' not in grains.get('product_version') | default('', true) %}
+      {% if 'build_image' and 'paygo' not in grains.get('product_version') | default('', true) %}
       - pkg: testsuite_salt_packages
       {% endif %}
 
 # modify Cobbler to be executed from remote-machines..
 {% set products_using_new_cobbler_version = ["uyuni-master", "uyuni-released", "uyuni-pr", "head", "4.3-released", "4.3-nightly", "4.3-pr"] %}
 {% set cobbler_use_settings_yaml = grains.get('product_version') | default('', true) in products_using_new_cobbler_version %}
-
+{% if 'build_image' and 'paygo' not in grains.get('product_version') | default('', true) %}
```
srbarrios, line 47:
```suggestion
{% if 'build_image' not in grains.get('product_version') and 'paygo' not in grains.get('product_version') %}
```

salt/server/testsuite.sls
```diff
@@ -78,12 +79,12 @@ testsuite_packages:
       - aaa_base-extras
       - wget
       - OpenIPMI
-    {% if 'build_image' not in grains.get('product_version') | default('', true) %}
+    {% if 'build_image' and 'paygo' not in grains.get('product_version') | default('', true) %}
```
srbarrios, line 82:
```suggestion
    {% if 'build_image' not in grains.get('product_version') and 'paygo' not in grains.get('product_version') %}
```

salt/minion/testsuite.sls
srbarrios, line 27:
```suggestion
    {% if 'build_image' not in grains.get('product_version') and 'paygo' not in grains.get('product_version') %}
```

salt/default/testsuite.sls
maximenoel8: The repos.testsuite will be call during default > minimal. I try to remove useless call to repo.

salt/minion/init.sls
maximenoel8: I tested this expression and it works. Do you want to have it in our format ?

salt/server/init.sls
maximenoel8: I copy this expression from repo > default =>     {% if 'beta' in grains.get('product_version') | default('', true) %}
I need to check without the default.

salt/minion/init.sls
srbarrios: It works just because the first part it returns always true.

salt/default/testsuite.sls
srbarrios: If one of the salt states of this sls depends on another sls, I think it would make sense to include it here.

rjmateus: IIRC what is on minimal should also be present in the general highstate. Somaone can call the script `highstate.sh` which runs the highstate only, and should have the same behaviour.

salt/default/testsuite.sls
```diff
@@ -14,7 +13,7 @@ default_cucumber_requisites:
       - milkyway-dummy
       - virgo-dummy
     - require:
-      - sls: repos
+      - sls: repos.testsuite
```
rjmateus, line 16: why was this change?

salt/repos/default.sls
rjmateus: default to true, means it not present by default. I agree it may not be needed, bu I don' this as risky

salt/default/testsuite.sls
maximenoel8: In pkg.installed, the packages are coming from repos.testsuite. I'm not an salt expert but I find it easier to read if we point directly to the needed resource and not to a folder with 20+ files.
Also, for my paygo minion, I want those repositories but not the other repositories. 
Because of the sumaform architecture, it's hard to split those resources. Calling the specific salt file solves the issue without adding extra condition around the requirement.

maximenoel8:
Thanks for the reviews, I still need a new deployment to test the default value. I introduce this condition 2 years ago so I don't remember completely. I remember I struggle a bit with this condition.
I think because we use `not in`, we need this default. In the other example, we don't have default if the condition is `in` and we have default if the condition is `not in`

maximenoel8:
The default value is required or you have this error : 

```
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    Rendering SLS 'base:default.minimal' failed: Jinja error: argument of type 'NoneType' is not iterable
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0mTraceback (most recent call last):
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m  File "/usr/lib/python3.6/site-packages/salt/utils/templates.py", line 476, in render_jinja_tmpl
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    output = template.render(**decoded_context)
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m  File "/usr/lib/python3.6/site-packages/jinja2/asyncsupport.py", line 76, in render
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    return original_render(self, *args, **kwargs)
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m  File "/usr/lib/python3.6/site-packages/jinja2/environment.py", line 1008, in render
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    return self.environment.handle_exception(exc_info, True)
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m  File "/usr/lib/python3.6/site-packages/jinja2/environment.py", line 780, in handle_exception
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    reraise(exc_type, exc_value, tb)
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m  File "/usr/lib/python3.6/site-packages/jinja2/_compat.py", line 37, in reraise
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m    raise value.with_traceback(tb)
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0m  File "<template>", line 8, in top-level template code
14:53:34  [0m[1mmodule.mirror.module.mirror.module.host.null_resource.provisioning[0] (remote-exec):[0m [0mTypeError: argument of type 'NoneType' is not iterable
```

salt/default/testsuite.sls
srbarrios: Please before resolving this, let's finish the discussion.

rjmateus (approved):

salt/default/minimal.sls
```diff
@@ -5,8 +5,10 @@ include:
   - default.network
   - default.firewall
   - default.avahi
-  {% if 'build_image' not in grains.get('product_version') | default('', true) %}
+  {% if 'build_image' and 'paygo' not in grains.get('product_version') | default('', true) %}
```
agraul, line 8: `{% if 'build_image' %}` is the same as writing `{% if true %}`. That's very likely not what you want.
You can try this yourself on https://j2live.ttl255.com/

Input:
```
{% if 'build_image' %}
- repos
{% endif %}
```
Output:
```yaml

- repos
```

salt/default/testsuite.sls
agraul: From a Salt PoV, IIRC the require works as long as `repos.testsuite` (in this case) is part of the sls files that are looked at. But from a sumaform developer PoV, I would always include it where I need it. Relying on global state is a general code smell.

salt/default/testsuite.sls
agraul: Keep in mind that `require` is just for ordering the states. `pkg.installed` takes a `fromrepo` argument if you want to specify which repo to actually use when installing things. Of course, if there is only one repo that has a package, that's automatically used by `zypper`.

salt/server/init.sls
agraul: here you can also write ` 'paygo' not in grains.get('product_version', '')` if you want to give a default. I find the usage of `default` a bit unclear. Can you tell me what the first and the second arguments mean? I had to look up the docs for that.

PR #1430, opened 2023-11-30
Add support for the iam role needed for paygo server

maximenoel8:
## What does this PR change?
To deploy correctly payo server, you need an iam marketplace role `metering.marketplace.amazonaws.com`
This PR is creating the iam role when needed.


mcalmer (approved):
I think it is ok. But AFAIK the calling account needs proper permissions. Will be funny

PR #1437, opened 2023-12-06
Fix iam role

maximenoel8:
## What does this PR change?
The former iam role was not complete and needed the instance profile. 
Also we need to use it only once and not for each instance so move it to base.


backend_modules/aws/base/iam_instance_profile.tf
```diff
@@ -0,0 +1,29 @@
...
+      },
+    ],
+  })
+}
+
+
+resource "aws_iam_role_policy_attachment" "metering_full_access_policy_attachment" {
+  policy_arn = "arn:aws:iam::aws:policy/AWSMarketplaceMeteringFullAccess"
+  role       = aws_iam_role.metering_full_access_role.name
+}
```
nodeg, line 29: Small nitpick: A newline at the end is missing.

PR #1438, opened 2023-12-07
fix enabling of repos

mcalmer:
## What does this PR change?

Here in repos we need to check different as the negate is happening for all the grains check

`not (grains.get("...") or grains.get("...") or 'payg' in grains.get("...") | default('', true))`

This makes also the fix from @Bischoff obsolete (I had to revert it as it creates now a duplicate ID)


Bischoff (approved):
nice catch, thanks

PR #1443, opened 2023-12-11
use OpenSUSE 15.5 for uyuni master, released and prs

mbussolotto:
## What does this PR change?

use OpenSUSE 15.5 for uyuni master, released and prs

### DIFF
--- a/modules/proxy/main.tf
+++ b/modules/proxy/main.tf
@@ -15,8 +15,8 @@ variable "images" {
     "4.3-build_image"= "sles15sp4o"
     "head"           = "sles15sp4o"
     "uyuni-master"   = "opensuse155o"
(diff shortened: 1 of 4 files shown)


cbosdo (commented):
These are useless without changing https://github.com/mbussolotto/sumaform/blob/15.5/modules/server_containerized/variables.tf#L276 to `default`

srbarrios:
> These are useless without changing https://github.com/mbussolotto/sumaform/blob/15.5/modules/server_containerized/variables.tf#L276 to `default`

Well, not exactly useless, we need both ;)
And keep in mind that this will affect all the environments, so let's announce it so everyone is aware.

srbarrios (approved):

deneb-alpha (approved):
LGTM, thanks

PR #1447, opened 2023-12-13
Fix early avahi installation on controller

mtravitzky:
## What does this PR change?
Previously, since the controller always runs on Leap, the custom repo was not added in the early stage of the controller's setup. Because of this, the installation of avahi-lang would always fail if avahi is enabled in main.tf. This usually wasn't noticed, because a later step will also attempt to install this package, which would then succeed. However, there might be a need for avahi being available during the early stages.

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -4,7 +4,7 @@ include:
 {% if grains['use_avahi'] and grains.get('osmajorrelease', None) != None %}
 
 # TODO: remove the following state when fix to bsc#1163683 is applied to all the SLES <= SLES15SP4
-{% if grains['osfullname'] == 'SLES' and grains['osrelease'] != '15.5' %}
+{% if (grains['osfullname'] == 'SLES' and grains['osrelease'] != '15.5') or (grains['osfullname'] == 'Leap') %}
 custom_avahi_repo:
   pkgrepo.managed:
     - humanname: custom_avahi_repo
@@ -26,6 +26,8 @@ custom_avahi_repo:
     - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.7/SLE_15_SP3/
     {% elif grains['osrelease'] == '15.4' %}
     - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.8/15.4/
+    {% elif grains['osrelease'] == '15.5' and grains['osfullname'] == 'Leap' %}
+    - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.8/15.5/
     {% endif %}
     - enabled: True
     - refresh: True


nodeg:
For reference: https://github.com/uyuni-project/sumaform/pull/1353

salt/default/avahi.sls
```diff
@@ -26,6 +26,8 @@ custom_avahi_repo:
     - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.7/SLE_15_SP3/
     {% elif grains['osrelease'] == '15.4' %}
     - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.8/15.4/
+    {% elif grains['osrelease'] == '15.5' and grains['osfullname'] == 'Leap' %}
```
srbarrios, line 29: You don't need to include `and grains['osfullname'] == 'Leap' ` it has been already processed in the parent `if` statement.

Bischoff (approved):
lgtm, but see Oscar's comment

salt/default/avahi.sls
agraul: I suggested that to make it clear why 15.5 is a valid `osrelease` grain even though there is a `osrelease != 15.5` check in the outer `if`. But if you like it without both conditions, that also works.

nodeg (approved):

ktsamis (approved):

salt/default/avahi.sls
```diff
@@ -4,7 +4,7 @@ include:
 {% if grains['use_avahi'] and grains.get('osmajorrelease', None) != None %}
 
 # TODO: remove the following state when fix to bsc#1163683 is applied to all the SLES <= SLES15SP4
-{% if grains['osfullname'] == 'SLES' and grains['osrelease'] != '15.5' %}
+{% if (grains['osfullname'] == 'SLES' and grains['osrelease'] != '15.5') or (grains['osfullname'] == 'Leap') %}
```
srbarrios, line 7: What I read here https://bugzilla.suse.com/show_bug.cgi?id=1163683#c5
It doesn't match to what we compare in this statement 🤔 
````
SP3, SP4 and SP5 already have avahi 0.8, which is not affected
But older SPs are still affected.
````

@juliogonzalez is this `grains['osrelease'] != '15.5'` correct?

File: salt/default/avahi.sls (hunk 1 above)
Comment (mtravitzky) [reply salt/default/avahi.sls:29]: I did this on recommendation from @agraul for better readability.

File: salt/default/avahi.sls (hunk 2 above)
Comment (juliogonzalez) [reply salt/default/avahi.sls:7]: Acccording to my comment, it seems it's not correct, and in fact it seems lines 25, 26, 27 and 28 could go away.

Please start podman containers and verify the avahi version there is 0.8 in 15.3, 15.4 and 15. If so, I guess such lines can be removed.

@Bischoff you were the expert in avahi IIRC?

Comment (Bischoff) [reply salt/default/avahi.sls:7]: I contributed a bit to the avahi project yes, not sure that makes an expert out of me :trollface:

Comment (mtravitzky) [reply salt/default/avahi.sls:7]: Anything new regarding this?

Comment (nodeg) [reply salt/default/avahi.sls:7]: I did a quick check with `zypper info avahi` and got:
15.3: avahi 0.7
15.4: avahi 0.8
15.5 avahi 0.8

Comment (juliogonzalez) [reply salt/default/avahi.sls:7]: ACK, then it seems Leap >= 15.4 and SLE15 >= SP4 do not need the extra avahi, so you can cleanup the sls file to remove:

```
    {% elif grains['osrelease'] == '15.4' %}
    - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.8/15.4/
    {% elif grains['osrelease'] == '15.5' and grains['osfullname'] == 'Leap' %}
    - baseurl: http://download.opensuse.org/repositories/systemsmanagement:/sumaform:/tools:/avahi:/0.8/15.5/
 ```
And if everything is working fine after the PR is merged, 15.4 and 15.5 could have avahi building disabled should be removed at systemsmanagement:/sumaform:/tools at OBS, to save OBS resources.

juliogonzalez: I wonder why my comment said SP3 does not have it but now Leap 15.3 has it... maybe someone bumped the version in SLE15SP3 after the EoL of Leap 15.3

juliogonzalez: No, I checked and it seems my comment was wrong. SLE15SP3 has 0.7

mtravitzky:
It is possible that this issue might have already been fixed by https://github.com/uyuni-project/sumaform/commit/3abec076b9a2957339cf5ce72f7b54bbbcfc6d94 . It does not occur during deployment using the current code.

nodeg:
Since we use openSUSE Leap 15.5 as default on the controller, and you verified that the issue is not present anymore, FMPOV this PR can get closed.

jordimassaguerpla:
> Since we use openSUSE Leap 15.5 as default on the controller, and you verified that the issue is not present anymore, FMPOV this PR can get closed.

Agree on closing. Thanks anyway for the work :+1:

PR #1454, opened 2023-12-29
Additional disk for /var/lib/pgsql partition custom size

srbarrios:
## What does this PR change?

This is a PR to parametrize the partition size for /var/lib/pqsql, aiming to solve not enough disk space issues in our CI tests.

To use it, you need to set this variable (on GB) in your main.tf:

```
    server = {
      ...
      database_disk_size = 50
    }
```

### DIFF
--- a/salt/server/additional_disk.sls
+++ b/salt/server/additional_disk.sls
@@ -1,10 +1,10 @@
 include:
   - default
 
-{% if grains.get('repository_disk_size') > 0 %}
-
 parted:
   pkg.installed
+
+{% if grains.get('repository_disk_size') > 0 %}
 
 {% set fstype = grains.get('data_disk_fstype') | default('ext4', true) %}
 {% if grains['data_disk_device'] == "nvme1n1" %}
@@ -20,10 +20,36 @@ spacewalk_partition:
     - require:
       - pkg: parted
 
+www:
+  group.present:
+    - gid: 469
+    - system: True
+
+wwwrun:
+  group.present:
+    - gid: 466
+    - system: True
+  user.present:
+    - fullname: WWW daemon apache
+    - shell: /usr/sbin/nologin
+    - home: /var/lib/wwwrun
+    - uid: 466
+    - gid: 466
+    - groups:
+      - wwwrun
+      - www
+
 spacewalk_directory:
   file.directory:
     - name: /var/spacewalk
     - makedirs: True
+    - user: wwwrun
+    - group: www
+    - dir_mode: 775
+    - recurse:
+      - user
+      - group
+      - mode
   mount.mounted:
     - name: /var/spacewalk
     - device: {{partition_name}}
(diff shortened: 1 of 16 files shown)


salt/server/additional_disk.sls
```diff
@@ -36,3 +36,32 @@ spacewalk_directory:
...
+  cmd.run:
+    - name: /usr/sbin/parted -s /dev/vdc mklabel gpt && /usr/sbin/parted -s /dev/vdc mkpart primary 0% 100% && sleep 1 && /sbin/mkfs.{{fstype}} {{partition_name}}
+    - unless: ls {{partition_name}}
+    - require:
+        - pkg: parted
+
+pgsql_directory:
+  file.directory:
+    - name: /var/lib/pgsql
+    - makedirs: True
```
mcalmer, line 55: This dir requires special ownership and permissions. If you create it, you should also adapt this

mcalmer (changes requested):

salt/server/additional_disk.sls
srbarrios: Thanks for the remark, I will take a look.
(For now, I did not even tried to deploy using this code)

srbarrios:
```
chmod 750 /var/lib/pgsql
chown postgres:postgres /var/lib/pgsql
```

srbarrios: Ok, it seems I will not need it here.
It seems that it's adjusted later on, when we install the packages for both suse manager and postgres.

As a fact, we also don't set up a user and group for the partition `/var/spacewalk` but it's still correct at the end of the deployment.

```
drwxr-xr-x  6 wwwrun root 4096 nov 24 16:23 spacewalk
```

mcalmer: ok

mcalmer (approved):

salt/server/additional_disk.sls
srbarrios: Well, at the end, it seems it was necessary 🤷🏼‍♂️

Bischoff (approved):

salt/server/additional_disk.sls
```diff
@@ -36,3 +62,52 @@ spacewalk_directory:
       - cmd: spacewalk_partition
 
 {% endif %}
+
+{% if grains.get('database_disk_size') > 0 %}
+
+{% set fstype = grains.get('data_disk_fstype') | default('ext4', true) %}
+{% set partition_name = '/dev/vdc1' %}
```
maximenoel8, line 69: I'm not sure it's  a good idea to have hard written the partition_name. It will not be compatible with cloud deployment. 
Why not using the same method than data disk ? grains['data_disk_device'] ?

maximenoel8 (changes requested):

salt/server/additional_disk.sls
srbarrios: Because that will require a bigger refactor, I choose this faster approach.

data_disk_device is not at server module level but at a lower level, so it makes things harder, we might want to convert data_disk_device into an array and rename it as data_disk_devices, otherwise it might not make sense depending the module.

I agree we can address it, but I would do it in another PR, so we have this ready today to make stable our CI again, currently, once we try to backup the database it fails and doesn't recover.

srbarrios: Honestly, I'm not super happy with the naming... but just for the sake of supporting cloud setups... I made the easiest adjustments I could to have another data_disk_device, I called it second_data_disk_device to keep it agnostic as this is setup at host module.

PR #1465, opened 2024-01-18
Add new controller variable to enable / disable the catch timeout mes…

maximenoel8:
## What does this PR 

During build validation, we have server instability when bootstrapping the clients in parallel but also when running the smoke tests in parallel. 
To manage this situation and avoid having to run those stages manually, we added a mechanism in the testsuite to catch those timeouts and reload the page when possible.
This mechanism is now optional and is disable by default

This PR add the possibility to enable the catch timeout mechanism during CI deployment.

It creates a new variable used in the controller bashrc to enable the catch timeout mechanism.

### DIFF
--- a/modules/controller/variables.tf
+++ b/modules/controller/variables.tf
@@ -763,7 +763,7 @@ variable "no_mirror" {
 }
 
 variable "catch_timeout_message" {
-  description = "Specify to controller, enable the mechanism to catch timeout message we have during BV"
+  description = "Enable the mechanism to catch the timeout message shown by a server overloaded"
   default     = false
 }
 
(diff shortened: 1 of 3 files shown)


modules/controller/variables.tf
```diff
@@ -762,6 +762,11 @@ variable "no_mirror" {
   default     = false
 }
 
+variable "catch_timeout_message" {
+  description = "Specify to controller, enable the mechanism to catch timeout message we have during BV"
```
srbarrios, line 766: I would avoid mentioning anything in regard BV, as sumaform should not know about this concept.

Maybe something like:
"Enable the mechanism to catch the timeout message shown by a server overloaded"

srbarrios (approved):
Approving, but please consider to address my suggestion.

NamelessOne91 (approved):

PR #1471, opened 2024-01-24
Tune for large deployments

Bischoff:
## What does this PR change?

This PR hopefully solves SUSE/spacewalk#23207

It conforms to https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html#_environmental_variables

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -883,7 +883,7 @@ module "server" {
 
 ## Large deployments
 
-In the case of the Build Validation test suite, or when trying to reproduce situations with a large number of clients, it is adviced to use `large_deployment` option. This option is inspired by the documentation at https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html, and it will apply the following settings on the server:
+In the case of the Build Validation test suite, or when trying to reproduce situations with a large number of clients, it is advised to use `large_deployment` option. This option is inspired by the documentation at https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html, and it will apply the following settings on the server:
 
 ```
 ### /etc/rhn/rhn.conf
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -881,6 +881,33 @@ module "server" {
 }
 ```
 
+## Large deployments
+
+In the case of the Build Validation test suite, or when trying to reproduce situations with a large number of clients, it is adviced to use `large_deployment` option. This option is inspired by the documentation at https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html, and it will apply the following settings on the server:
+
+```
+### /etc/rhn/rhn.conf
+taskomatic.com.redhat.rhn.taskomatic.task.MinionActionExecutor.parallel_threads = 3
+hibernate.c3p0.max_size = 50
+
+### /etc/tomcat/server.xml
+changed `maxThreads` to 256
+
+### /var/lib/pgsql/data/postgresql.conf
+max_connections = 450
+work_mem = 10MB
+```
+
+An example follows:
+
+```hcl
+module "server" {
+   ...
+   large_deployment = true
+   ...
+}
+```
+
 ## Debugging facilities
 
 The `server` module has options to automatically capture more diagnostic information, off by default:
--- a/salt/server/testsuite.sls
+++ b/salt/server/testsuite.sls
@@ -155,6 +155,7 @@ tomcat:
     - watch:
       - file: enable_salt_content_staging_window
       - file: enable_salt_content_staging_advance
+      - file: enable_kiwi_os_image_building
 
 dump_salt_event_log:
     cmd.run:
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -29,7 +29,7 @@ locals {
   additional_packages       = { for host_key in local.hosts :
     host_key => lookup(var.host_settings[host_key], "additional_packages", []) if var.host_settings[host_key] != null }
   additional_grains       = { for host_key in local.hosts :
-  host_key => lookup(var.host_settings[host_key], "additional_grains", {}) if var.host_settings[host_key] != null }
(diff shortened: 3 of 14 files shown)


README_ADVANCED.md
```diff
@@ -881,6 +881,33 @@ module "server" {
...
+### /etc/rhn/rhn.conf
+taskomatic.com.redhat.rhn.taskomatic.task.MinionActionExecutor.parallel_threads = 3
+hibernate.c3p0.max_size = 50
+
+### /etc/tomcat/server.xml
+changed `maxThreads` to 256
+
+### /var/lib/pgsql/data/postgresql.conf
+max_connections = 450
+work_mem = 10MB
```
srbarrios, line 898: I wonder if we could speed up our reposynching process of metadata repositories with a higher value in work_mem. @rjmateus what do you think?

README_ADVANCED.md
```diff
@@ -881,6 +881,33 @@ module "server" {
...
+```
+### /etc/rhn/rhn.conf
+taskomatic.com.redhat.rhn.taskomatic.task.MinionActionExecutor.parallel_threads = 3
+hibernate.c3p0.max_size = 50
+
+### /etc/tomcat/server.xml
+changed `maxThreads` to 256
+
+### /var/lib/pgsql/data/postgresql.conf
+max_connections = 450
```
srbarrios, line 897: > Value recommendation =  2 * hibernate.c3p0.max_size + apache MaxClients + 70, if less than 1000

2 * 50 + 150 + 70 = 320

I think we need to increase apache MaxClients from the default (150) to 150+80 = 230.

Would that make sense for you?

srbarrios:
Just some comments in-line but it looks good to me.

If I can bring an idea (for later), I think it will be good to have these values adjusted dynamically depending our terraform file.
Meaning:
- We count the amount of `minion` or `ssh_minion` modules defined to be deployed on the main.tf
- We count the amount of images defined in the `images` list
- Something else?

From this information, we make some formulas to obtain the tweaked value.

💡  @rjmateus @admd We could even consider this as a feature in our SUSE Manager Server, as the server knows all these details, we can give that recommendation to our Administrator through a notification or similar.

srbarrios (approved):

Bischoff:
> If I can bring an idea (for later), I think it will be good to have these values adjusted dynamically depending our terraform file. 

Call me a control freak, but somehow I prefer to specify those values myself... :smile: 


> 💡 @rjmateus @admd We could even consider this as a feature in our SUSE Manager Server, as the server knows all these details, we can give that recommendation to our Administrator through a notification or similar.

:thinking: Isn't that a mission for the new sanity check tool? Maybe it's even already implemented in it?

README_ADVANCED.md
rjmateus: You don't need to change the MaxClients. We need to have it aligned with serverLimit and maxThreads.
I added a higher value to max connections because we can have more database connections. For example, imagine that someone is debugging and needs to connect to database, it will also be using a DB connection.

README_ADVANCED.md
rjmateus: I don't think so. This is usefull for complex queries, that uses a lot of data. reposync is a mix of CPU usage to extract data, and is insert-heavy. I don't think it uses huge queries with a lot of joins and sorts.
We could increase the max wal size, but that is already part of the default configuration. 

Postgresql docs: https://www.postgresql.org/docs/current/runtime-config-resource.html#GUC-WORK-MEM

salt/server/testsuite.sls
```diff
@@ -155,6 +155,7 @@ tomcat:
     - watch:
       - file: enable_salt_content_staging_window
       - file: enable_salt_content_staging_advance
+      - file: enable_kiwi_os_image_building
```
rjmateus, line 158: why is this in here?

rjmateus (approved):

salt/server/testsuite.sls
Bischoff: Piggy backing.

A dependency that we apparently forgot previously.

rjmateus:
> 
> From this information, we make some formulas to obtain the tweaked value.
> 

We have a tool already for that, the smdba: `smdba system-check autotuning --max_connections=400 --ssd`

However, this can be improved in the context of the check tool that is under development.

srbarrios:
> > From this information, we make some formulas to obtain the tweaked value.
> 
> We have a tool already for that, the smdba: `smdba system-check autotuning --max_connections=400 --ssd`
> 
> However, this can be improved in the context of the check tool that is under development.

Still... something more visual, like a notification sent to the WebUI or email, it might help to our users.
As a matter of fact, we don't even mention this command into the https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html doc page, which it will be an obvious page to place it.

README_ADVANCED.md
```diff
@@ -881,6 +881,33 @@ module "server" {
 }
 ```
 
+## Large deployments
+
+In the case of the Build Validation test suite, or when trying to reproduce situations with a large number of clients, it is adviced to use `large_deployment` option. This option is inspired by the documentation at https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html, and it will apply the following settings on the server:
```
vandabarata, line 886:
```suggestion
In the case of the Build Validation test suite, or when trying to reproduce situations with a large number of clients, it is advised to use `large_deployment` option. This option is inspired by the documentation at https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html, and it will apply the following settings on the server:
```

rjmateus:
Tunning a SUSE Manager server is not an exact science. It relates to the way/features in use. So it's hard to have an automation in this regards, because you may be adding more resources to something not used by the customer.

vandabarata (approved):

srbarrios:
> Tunning a SUSE Manager server is not an exact science. It relates to the way/features in use. So it's hard to have an automation in this regards, because you may be adding more resources to something not used by the customer.

That can of "AI" suggestions is what offers an extra value to customers nowadays, so.. if a person can figure out these values, a machine can do it too.
I can understand is not a matter for today, not even something can fit in the a medium-term roadmap, but.. we should think harder on that kind of features.

cbosdo (approved):
LGTM

PR #1478, opened 2024-01-25
SLE micro 5.1 and 5.2 use wicked

Bischoff:
## What does this PR change?

IPv6 does not need fixing in SLE Micro 5.1 and 5.2

### DIFF
--- a/salt/default/network.sls
+++ b/salt/default/network.sls
@@ -19,7 +19,7 @@ ipv6_accept_ra_{{ iface }}:
 {% endif %}
 {% endfor %}
 
-{% if grains['osfullname'] == 'SLE Micro' %}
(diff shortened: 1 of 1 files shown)


salt/default/network.sls
```diff
@@ -19,7 +19,7 @@ ipv6_accept_ra_{{ iface }}:
 {% endif %}
 {% endfor %}
 
-{% if grains['osfullname'] == 'SLE Micro' %}
+{% if grains['osfullname'] == 'SLE Micro' and grains['osrelease'] != '5.1' and grains['osrelease'] != '5.2' %}
```
ktsamis, line 22: you need an or statement between the versions no? It can't be both of them

ktsamis (changes requested):

ktsamis (approved):

NamelessOne91 (approved):

PR #1479, opened 2024-01-25
Create a variable for the main disk size

maximenoel8:
## What does this PR change?
Create a variable to be able to increase the main disk size


srbarrios (approved):

nodeg (approved):
Thank you, Maxime!

cbosdo (commented):
What would happen with an AWS backend since it doesn't have that variable?

PR #1481, opened 2024-01-26
Add `product_version` to cont. server validation

nodeg:
## What does this PR change?

Add the `product_version` grain to the containerized server tests. We also have this defined in the normal server tests.

 Necessary for e.g. https://github.com/uyuni-project/sumaform/pull/1475

### DIFF
--- a/.github/workflows/ci-validation/salt-server-containerized-validation
+++ b/.github/workflows/ci-validation/salt-server-containerized-validation
@@ -18,6 +18,7 @@ osrelease: '15.5'
 osrelease_info: [15, 5]
 
 hostname: sumaform-test-server
+product_version: head
 install_salt_bundle: true
 gpg_keys: ['testkey.pub']
 domain: tf.local


.github/workflows/ci-validation/salt-server-containerized-validation
```diff
@@ -18,6 +18,7 @@ osrelease: '15.5'
 osrelease_info: [15, 5]
 
 hostname: sumaform-test-server
+product_version: head
```
srbarrios, line 21: Here, in my opinion, we better use uyuni-master.

nodeg: Fine with me. The important part is that this variable/grain is mentioned. Otherwise the test will fail.

cbbayburt: Yes, the actual value doesn't matter for these validations. I usually have a look at the if-else conditions in the Salt jinjas, and pick a value that will enable the most if-else blocks.

cbbayburt (approved):

nodeg:
Ok, then I would leave it as is.

PR #1483, opened 2024-01-29
Support cloud-init on slemicro55o for not containerized servers

srbarrios:
## What does this PR change?

This PR refactor the cloud-init support for slemicro5.5, to be able to deploy it as a regular minion, instead of a containerized server.
It also removed an entry in the list of images, as we were duplicating the URL, and when deploying a sle micro 5.5 minion with `slemicro55-ign`, it seems there was some error in the image and we did not have the repositories in place.


cbosdo (approved):
LGTM

Bischoff (approved):

PR #1498, opened 2024-02-07
Make IPv6 changes to SLE micro permanent

Bischoff:
## What does this PR change?

Let IPv6 changes to SLE micro survive a reboot.

In Network Manager world, `nmcli device` is apparently transient, while `nmcli connection` seems to be permanent.


nodeg:
Thank you, Eric. I wonder if this will really do the trick for SLE Micro since we do not do/include it into a snapshot.

Bischoff:
> Thank you, Eric. I wonder if this will really do the trick for SLE Micro since we do not do/include it into a snapshot.

It will not be a problem for customers if they don't need predictable addresses.

And if they do need them, the answer is on slack exchange for everyone in that case :trollface: .

nodeg:
I did not have customers in mind, but only our own infrastructure for now :wink:

NamelessOne91 (approved):

PR #1503, opened 2024-02-13
Do not add custom avahi repo on SLES15 SP4

mtravitzky:
## What does this PR change?
Since SLES 15 SP4 already has the [fixed](https://bugzilla.suse.com/show_bug.cgi?id=1163683) avahi 0.8, this is not needed anymore. It was only needed for older SLES releases which use older avahi versions, since a necessary fix has been backported to them. Avahi 0.8 has this fix out of the box.

[This comment](https://bugzilla.suse.com/show_bug.cgi?id=1163683#c5) mentions SP3 also having avahi 0.8, so SP3 might also be a candidate for removal, but this isn't verified yet.

Edit: SP3 has avahi 0.7, so it still needs this.


ktsamis (approved):
Looks good to me, let the others also take a look before merging

nodeg (approved):
If you tested this, then it's fine to merge.

PR #1504, opened 2024-02-16
Allow to override the host name for libvirt too

Bischoff:
## What does this PR change?

This PR:
 * allows to override the host name for libvirt like it's already possible for AWS
 * removes the variable `overwrite_fqdn` definition from `host` and `server_containerized` modules,
    as `overwrite_fqdn` is only used as a hash key in the provider settings map
* documents the functionality
* fixes typos in external database section of documentation (piggyback)

This for example allows me to have both PRV and NUE test suites on the same ARM server without conflicts at libvirt level.

By the way, shouldn't we name this key `override_fqdn` rather than `overwrite_fqdn`? :thinking:

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -910,6 +910,67 @@ module "server" {
 }
 ```
 
+## Using a different FQDN
+
+Normally, the fully qualified domain name (FQDN) is derived from `name` variable. However, some providers, like AWS cloud provider, impose a naming scheme that does not always match this mechanism. You may also want a name for libvirt that differs from the hostname part of the FQDN. The `overwrite_fqdn` variable allows the FQDN to diverge from the value normally derived from the name.
+
+An AWS example is:
+
+```hcl
+module "cucumber_testsuite" {
+  source = "./modules/cucumber_testsuite"
+  ...
+  host_settings = {
+    ...
+    server = {
+      provider_settings = {
+        instance_type = "m6a.xlarge"
+        volume_size = "100"
+        private_ip = "172.16.3.6"
+        overwrite_fqdn = "uyuni-master-srv.sumaci.aws"
+      }
+    }
+    ...
+  }
+  ...
+}
+```
+
+A libvirt example is:
+
+```hcl
+module "opensuse155arm-minion" {
+  source = "./modules/minion"
+  ...
+  name = "nue-min-opensuse155arm"
+  ...
+  provider_settings = {
+    ...
+    overwrite_fqdn   = "suma-bv-43-min-opensuse155arm.mgr.suse.de"
+    ...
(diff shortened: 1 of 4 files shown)


README_ADVANCED.md
```diff
@@ -910,6 +910,44 @@ module "server" {
 }
 ```
 
+## Using a different FQDN
+
+Normally, the fully qualified domain name (FQDN) is derived from `name` variable. However, some providers, like AWS cloud provider, impose a naming scheme that does not always match this mechanism. You may also want a name for libvirt that differs from the hostname part of the FQDN. The `overwrite_fqdn` variable allows the FQDN to diverge from the value normally derived from the name.
```
szachovy, line 915: Could you give an example of the naming scheme from a cloud provider that does not match FQDN?

README_ADVANCED.md
```diff
@@ -910,6 +910,44 @@ module "server" {
...
+-----------------------------------------------------------
+ ...
+ 11   uyuni-bv-master-nue-min-opensuse154arm      running
+```
+
+and inside the VM:
+
+```bash
+# hostname -f
+uyuni-bv-master-min-opensuse154arm.mgr.suse.de
```
szachovy, line 947: Shouldn't the hostname FQDN match overwritten FQDN name?

README_ADVANCED.md
Bischoff: ok

README_ADVANCED.md
Bischoff: it does

nodeg: But not here in your example.
`suma-bv-43-min-opensuse155arm.mgr.suse.de` vs `uyuni-bv-master-min-opensuse154arm.mgr.suse.de`

Bischoff: fixed, please mark thread as "resolved" if you are satisfied with new version

README_ADVANCED.md
Bischoff: done, please resolve conversation if you are happy with the result

szachovy: I think giving it in italic parentheses just to be aware what domain crash would be sufficient, but current form is also ok.

szachovy (approved):

nodeg (commented):
I still do not get it why you want to remove the variable from the `null` backend. Could you please explain that a bit further?

nodeg:
Ok. Thank you for the clarification. Since mc introduced that, I would prefer a review from him, too.

Bischoff:
> Ok. Thank you for the clarification. Since mc introduced that, I would prefer a review from him, too.

I discussed this with MC in private.

PR #1505, opened 2024-02-19
Remove duplicate repository

Bischoff:
## What does this PR change?

Fix repository installed twice:
```
[2024-02-16T18:19:45.729Z] module.slemicro55-sshminion.module.sshminion.module.host.null_resource.provisioning[0] (remote-exec):
     Comment: Failed to configure repo 'test_repo_rpm_pool': Repository 'test_repo_rpm_pool' already exists as 'test_packages_pool'.
```

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -362,12 +362,6 @@ zypper:
       enabled: 1
       autorefresh: 1
       gpgcheck: false
-    - id: test_packages_pool
-      name: test_packages_pool
-      baseurl: http://${ use_mirror_images ? mirror : "download.opensuse.org" }/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Pool/rpm/
-      enabled: 1
-      autorefresh: 1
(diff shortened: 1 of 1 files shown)


nodeg (approved):
The part you remove was added by Cedric on 29th of January with https://github.com/uyuni-project/sumaform/pull/1475.
What I see is that we add this repo once during cloud-init (`test_packages_pool`) and the second time in `salt/repos/testsuite.sls` as `test_repo_rpm_pool`.
However, since this repository only contains our dummy packages, I think we do not need them that early in cloud-init.

Bischoff:
> The part you remove was added by Cedric on 29th of January with #1475. What I see is that we add this repo once during cloud-init (`test_packages_pool`) and the second time in `salt/repos/testsuite.sls` as `test_repo_rpm_pool`. However, since this repository only contains our dummy packages, I think we do not need them that early in cloud-init.

Yes. And another argument for removing it in cloud-init is that we do not do that for the other SLE micros.

NamelessOne91 (approved):

PR #1507, opened 2024-02-19
fix default repo for avahi installation

jordimassaguerpla:
## What does this PR change?

the default repo should have been in the "if" clause.

### DIFF
--- a/salt/default/avahi.sls
+++ b/salt/default/avahi.sls
@@ -64,9 +64,9 @@ avahi_pkg:
       - libavahi-core7
       {% endif %}
       {% endif %}
-{% endif %}
     - requires:
       - pkgrepo: os_pool_repo
+{% endif %}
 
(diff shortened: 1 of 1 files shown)


mtravitzky:
This change might bring back the issue mentioned here: https://github.com/uyuni-project/sumaform/pull/1447

because as the code is right now, with these lines outside the if-clause, it would have appended these lines to `custom_avahi_repo` in the case of SLES, which I think made the avahi installation work even if there was no custom avahi repo. But I might be completely wrong.

nodeg:
I am not sure why this was initially changed so FMPOV, we can revert it and merge this PR, and then we should consider https://github.com/uyuni-project/sumaform/pull/1447 again.

mtravitzky:
The issue in https://github.com/uyuni-project/sumaform/pull/1447 has not been brought back by this PR, so all is good.

PR #1520, opened 2024-02-28
handle main_disk_size for cucumber_testsuite module

mcalmer:
## What does this PR change?

Make it possible to configure the disk size in the cucumber_testsuite module


Bischoff (approved):
Thank you

It means that commit `3c8f550e` in susemanager-ci never worked: on the reference environment for Uyuni, we use `main_disk_size` for the server, from the cucumber module test suite too.

As a side note, at some point we will also need the other 2 variables:  `repository_disk_size` and `database_disk_size`. Which will not work anyway with a containerized proxy until we make `mgradm` capable of creating the symlinks from container volumes to real partitions.

PR #1524, opened 2024-03-04
DHCP and DNS host

Bischoff:
## What does this PR change?

This PR makes it possible to deploy a DHCP and DNS server inside the private network. This is useful for the containerized proxy because it stops playing that role in Retail context.

The proxy host still has to do the routing and NAT masquerading (for example via the branch network formula).

This PR also stops inserting the traditional client and the normal minion into the private network.

Some adaptation in the test suite will be needed, especially in containerized proxy context, but not only.

@rjmateus could you tell me what you think of the change to   `modules/base/main.tf`? It could be the sign of a problem with the decoupling of modules and backend modules.

### DIFF
--- a/modules/dhcp_dns/main.tf
+++ b/modules/dhcp_dns/main.tf
@@ -1,0 +1,147 @@
+module "dhcp_dns" {
+  source = "../host"
+
+  base_configuration            = var.base_configuration
+  name                          = var.name
+  quantity                      = var.base_configuration["additional_network"] != null ? var.quantity : 0
+
+  connect_to_base_network       = false
+  connect_to_additional_network = true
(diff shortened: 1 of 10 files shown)


modules/dhcp_dns/main.tf
```diff
@@ -0,0 +1,146 @@
+module "dhcp_dns" {
+  source = "../host"
+
+  base_configuration            = var.base_configuration
+  name                          = var.name
+  quantity                      = var.base_configuration["additional_network"] != null ? var.quantity : 0
+// pffff
```
srbarrios, line 7: 🤡

Bischoff:
Closing in favour of other version

PR #1525, opened 2024-03-05
DHCP and DNS host

Bischoff:
## What does this PR change?

This PR makes it possible to deploy a DHCP and DNS server inside the private network. This is useful for the containerized proxy because it stops playing that role in Retail context.

This PR also stops inserting the traditional client and the normal minion into the private network. They were only used for ping tests.

This PR also works around a bug transmitting the configuration from AWS :cry: .

Some adaptation in the test suite will be needed, especially in containerized proxy context, but not only.

## Results

```bash
dhcp-dns:~ # ip a s eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 52:54:00:de:7d:2d brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    altname ens3
    inet 192.168.1.53/24 brd 192.168.1.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::5054:ff:fede:7d2d/64 scope link 
       valid_lft forever preferred_lft forever
```

```bash
dhcp-dns:~ # cat /etc/resolv.conf
nameserver 127.0.0.1
search example.org
```

```bash
dhcp-dns:~ # cat /etc/dhcpd.conf
option domain-name "example.org";
option domain-name-servers 192.168.1.53;

subnet 192.168.1.0 netmask 255.255.255.0
{
  range 192.168.1.128 192.168.1.253;
  filename "boot/pxelinux.0";
  next-server 192.168.1.254;
}

host proxy
{
  hardware ethernet 52:54:00:F6:84:44;
  fixed-address 192.168.1.254;
}

host pxeboot
{
  hardware ethernet 52:54:00:12:CE:0B;
  fixed-address 192.168.1.4;
}
```

```bash
dhcp-dns:~ # cat /etc/named.conf.include
zone "1.168.192.in-addr.arpa" {
  type master;
  file "/var/lib/named/master/db.1.168.192.in-addr.arpa";

  notify no;
};

zone "example.org" {
  type master;
  file "/var/lib/named/master/db.example.org";

  notify no;
};
```

```bash
dhcp-dns:~ # cat /var/lib/named/master/db.1.168.192.in-addr.arpa
$ORIGIN 1.168.192.in-addr.arpa.

@        IN SOA dhcp-dns.example.org. admin@example.org. (2045010101 8600 900 86000 500)
@        IN NS  dhcp-dns.example.org.
dhcp-dns.example.org. IN A   192.168.1.53

53       IN PTR dhcp-dns.example.org.
254      IN PTR proxy.example.org.

4        IN PTR pxeboot.example.org.
```

```bash
dhcp-dns:~ # cat /var/lib/named/master/db.example.org
$ORIGIN example.org.

@        IN SOA dhcp_dns admin@example.org. (2045010101 8600 900 86000 500)
@        IN NS  dhcp_dns
dhcp-dns IN A   192.168.1.53

proxy    IN A   192.168.1.254

pxeboot  IN A   192.168.1.4
```

### DIFF
--- a/modules/dhcp_dns/main.tf
+++ b/modules/dhcp_dns/main.tf
@@ -1,0 +1,147 @@
+module "dhcp_dns" {
+  source = "../host"
+
+  base_configuration            = var.base_configuration
+  name                          = var.name
+  quantity                      = var.base_configuration["additional_network"] != null ? var.quantity : 0
+
+  connect_to_base_network       = false
+  connect_to_additional_network = true
+  roles                         = ["dhcp_dns"]
+  provision                     = false
+
+  image                         = var.image
+}
+
+# The DHCP and DNS server has no direct connection to the outside, so we must download the packages on its behalf
+locals {
+  add_net              = var.base_configuration["additional_network"] != null ? slice(split(".", var.base_configuration["additional_network"]), 0, 3) : []
+  prefix               = join(".", local.add_net)
+  reverse_prefix       = join(".", reverse(local.add_net))
+  zypper               = "/usr/bin/zypper --non-interactive --gpg-auto-import-keys"
+  # Currently only for openSUSE 15.5:
+  repo                 = "http://${var.base_configuration["mirror"] != null ? var.base_configuration["mirror"] : "download.opensuse.org"}/distribution/leap/15.5/repo/oss"
+}
+
+resource "null_resource" "standalone_provisioning" {
+
+  count = var.base_configuration["additional_network"] != null ? var.quantity : 0
+
+  connection {
+    host     = "${local.prefix}.53"
+    user     = "root"
+    password = "linux"
+    timeout  = "3m"
+  }
+
+  provisioner "local-exec" {
+    command = <<EOT
+mkdir -p /tmp/dhcp-dns/var/cache/zypp/packages/repo
+${local.zypper} --root /tmp/dhcp-dns addrepo ${local.repo} offline_repo ||:
+${local.zypper} --root /tmp/dhcp-dns install --download-only dhcp-server bind
+EOT
+  }
+
+  provisioner "file" {
+    source      = "/tmp/dhcp-dns/var/cache/zypp/packages/offline_repo"
+    destination = "/root"
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "echo ${var.name} > /etc/hostname",
+      "hostname ${var.name}",
+      "echo 'nameserver 127.0.0.1' > /etc/resolv.conf",
+      "echo 'search example.org' >> /etc/resolv.conf",
+      "${local.zypper} addrepo dir:/root/offline_repo offline_repo ||:",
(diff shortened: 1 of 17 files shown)


nodeg (commented):
I read through the PR, and the most difficult part for me is the new `dhcp_dns` module. What I can say, it LGTM, but I will not approve because of that.

cbbayburt (approved):
I have no idea about the networking details, but LGTM (as far as I can tell). Neat solution! :+1:

NamelessOne91:
LGTM but I know too little about network setups to really judge the configuration files created in the `dhcp_dns` module.

modules/dhcp_dns/variables.tf
```diff
@@ -0,0 +1,34 @@
...
+}
+
+variable "first_terminal_ip" {
+  description = "last digit of first terminal IP's address"
+  type        = number
+  default     = 1
+}
+
+variable "image" {
+  description = "an image name, e.g. sles12sp4 or opensuse155o"
```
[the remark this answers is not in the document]
Bischoff: right (it was a copy and paste)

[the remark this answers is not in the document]
Bischoff: done

modules/dhcp_dns/main.tf
```diff
@@ -0,0 +1,148 @@
...
+
+  provisioner "file" {
+    content = <<EOT
+option domain-name "example.org";
+option domain-name-servers ${local.prefix}.53;
+
+subnet ${local.prefix}.0 netmask 255.255.255.0
+{
+  range ${local.prefix}.128 ${local.prefix}.253;
+  filename "boot/pxelinux.0";
```
aaannz, line 72: FYI so that things are not completely ease, "normal" retail does use `boot/` prefix. But cobbler and containerized retail does not.

modules/dhcp_dns/main.tf
```diff
@@ -0,0 +1,148 @@
...
+  next-server ${local.prefix}.254;
+}
```
aaannz, line 74: When you are at it, you may want to add uefi and http uefi sections as well ([line 57](https://github.com/SUSE/salt-formulas/blob/master/dhcpd-formula/dhcpd/files/subnet.jinja#L57) and [line 72](https://github.com/SUSE/salt-formulas/blob/master/dhcpd-formula/dhcpd/files/subnet.jinja#L72))

Bischoff: I'm not sure. For the moment, we are not testing UEFI hosts.

modules/dhcp_dns/main.tf
Bischoff: Just for the record: the likely solution as of today is to create SSH keys on the proxy to let me access the dhcp-dns server and change this setting dynamically without too much hassle.

PR #1526, opened 2024-03-06
Update jdupes source

maximenoel8:
## What does this PR change?

jdupes package url changed.
I tested the new version with the mirror deployment and I had not compatibility issues.


mcalmer (approved):
LGTL but please do us a favor.
I think we have in our infrastructure gitlab repo the same state for our mirrors.
Would you do a PR for that as well please?

nodeg (approved):

NamelessOne91 (approved):

PR #1536, opened 2024-03-13
salt_testenv: start docker daemon after classic Salt testsuite installation

meaksh:
## What does this PR change?

In order to allow some Docker related tests to run, we want "docker" service is enable when testing classic Salt package.


agraul (approved):
Is the reason for not starting it for the Salt Bundle cases that we don't include a docker client library in the bundle?

meaksh:
> Is the reason for not starting it for the Salt Bundle cases that we don't include a docker client library in the bundle?

Not really, we do actually include the docker library in the Salt Bundle, but `docker` service is only installed on testing instances where we also test the classic package (as "docker" is a dependency only for "python3-salt-testsuite" package).

PR #1542, opened 2024-03-18
Support containerized proxy as part of private network

Bischoff:
## What does this PR change?

Support containerized proxy as part of private network

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -19,6 +19,8 @@ module "base" {
 
 locals {
   server_full_name          = "${var.name_prefix}srv.${var.domain}"
+  proxy_full_name           = "${var.name_prefix}pxy.${var.domain}"
+
   hosts                     = keys(var.host_settings)
   provider_settings_by_host = { for host_key in local.hosts :
     host_key => lookup(var.host_settings[host_key], "provider_settings", {}) if var.host_settings[host_key] != null }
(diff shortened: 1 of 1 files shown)


modules/cucumber_testsuite/main.tf
```diff
@@ -217,19 +217,22 @@ module "proxy_containerized" {
...
 
   quantity           = contains(local.hosts, "dhcp-dns") ? 1 : 0
   base_configuration = module.base.configuration
   image              = lookup(local.images, "dhcp-dns", "opensuse155o")
   name               = lookup(local.names, "dhcp-dns", "dhcp-dns")
-  private_hosts      = [ module.proxy[0].configuration, module.pxeboot-minion.configuration ]
+
+  private_hosts      = [ var.container_proxy ? module.proxy_containerized[0].configuration : module.proxy[0].configuration,
+                         module.pxeboot-minion.configuration
+                       ]
```
srbarrios, line 230:
```suggestion
 proxy              = var.container_proxy ? module.proxy_containerized[0].configuration : module.proxy[0].configuration
  private_hosts      = [ proxy, module.pxeboot-minion.configuration ]
```

Bischoff: this is what i tried... but you need either a global variable or local variable

Bischoff: ok i ended up adding it to the locals, and reusing that local in the controller module

modules/cucumber_testsuite/main.tf
```diff
@@ -152,6 +152,13 @@ module "server_containerized" {
   large_deployment              = lookup(local.large_deployment, "server_containerized", false)
 }
 
+locals {
+  proxy_full_name       = "${var.name_prefix}pxy.${var.domain}"
+  minimal_configuration = { hostname = contains(local.hosts, "proxy") ? local.proxy_full_name : local.server_full_name }
+  server_configuration  = var.container_server ? module.server_containerized[0].configuration : module.server[0].configuration
+  proxy_configuration   = var.container_proxy ? module.proxy_containerized[0].configuration : module.proxy[0].configuration
+}
```
srbarrios, line 160: Personally, I prefer to have a unique locals section, I would move this inside the section we have starting at line 20.

Bischoff: done

srbarrios (approved):

PR #1543, opened 2024-03-18
Support containerized proxy in private network

Bischoff:
## What does this PR change?

Support containerized proxy in private network


srbarrios (commented):
LGTM

PR #1544, opened 2024-03-20
Reduce default main disk size to 20GiB

maximenoel8:
## What does this PR change?

Currently, all the instances are by default deploy with the main disk size equal to 200GiB. 
We want to reduce this default size to 20GiB.

I also reduced proxy disk size ( previously 200GiB )

### DIFF
--- a/backend_modules/null/host/variables.tf
+++ b/backend_modules/null/host/variables.tf
@@ -108,7 +108,7 @@ variable "provider_settings" {
 
 variable "main_disk_size" {
   description = "Size of main disk, defined in GiB"
-  default     = 200
+  default     = 20
 }
 
 variable "additional_disk_size" {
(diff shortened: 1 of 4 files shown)


mcalmer:
What is the general strategy? Which system get which size of disk and how many disks?
At lease we need to adapt also the dedicated configurations.

rjmateus:
Why this need showed up? That size is the max size, the disk is dynamic and will allow space as needed. That will reduce the size for all machines. I think we should not change this, unless we have a specific reason.

srbarrios:
> Why this need showed up? That size is the max size, the disk is dynamic and will allow space as needed. That will reduce the size for all machines. I think we should not change this, unless we have a specific reason.

Agree on not changing it. But just a small remark, assuming the disk to have a dynamic size I think it only applies to libvirt but it might not be the case for AWS. To be confirmed.

rjmateus:
that's true, it only applies to libvirt

PR #1547, opened 2024-03-21
Change the default value of install_salt_bundle to true

srbarrios:
## What does this PR change?

This PR change the default value of _install_salt_bundle_ variable in the majority of modules.
It keeps untouched the host modules from other backends.

The aim of this refactor is to simplify our main.tf for most of our environments, where we were replacing the value of this variable in almost all the cases.


srbarrios:
Not tested yet.

Bischoff (approved):
approving although i would prefer a complete removal of everything non-bundle (including in the cloudinit code)

nodeg (approved):
Fine with me.

PR #1551, opened 2024-03-22
Install salt bundle, when required, through Combustion

srbarrios:
## What does this PR change?

Install salt bundle, when required, through Combustion.
For instance, we need it to bootstrap the Containerized Proxy Host when it's running in a SLE Micro 5.5


srbarrios:
I will not wait to merge it, I want to fix Head CI asap.

PR #1560, opened 2024-04-08
Use Leap Micro in Uyuni Proxy Containerized

srbarrios:
## What does this PR change?

Use Leap Micro in Uyuni Proxy Containerized.
This change the default image for all the uyuni product versions.

**WARNING**:
- We need to change the default image if we are under a K3S deployment. In such case, for now, we will keep using Leap Micro 5.5, to be aligned with the server, as we encountered some technical issues there.

**Related PR**:
- https://github.com/SUSE/susemanager-ci/pull/1190

### DIFF
--- a/modules/cucumber_testsuite/variables.tf
+++ b/modules/cucumber_testsuite/variables.tf
@@ -38,7 +38,7 @@ variable "name_prefix" {
 
 variable "images" {
   description = "list of images to be uploaded to the libvirt host, leave default for all"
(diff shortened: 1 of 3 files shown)


aaannz (approved):
Do I get it right that we are not changing base image for k3s automatically and it depends on main.tf override? If so LGTM.

jordimassaguerpla:
I would not switch to these defaults until we have the PR jobs green https://ci.suse.de/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests-reference/ .

jordimassaguerpla:
> I would not switch to these defaults until we have the PR jobs green https://ci.suse.de/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests-reference/ .

Jobs are green

jordimassaguerpla (approved):

PR #1562, opened 2024-04-10
Install expect on proxy

Bischoff:
## What does this PR change?

This PR installs `expect` package on the SLE micro proxy.

## Context

This is the third attempt to do the same thing.

The first 2 were done at test suite level:
 * the first one failed because it required too many reboots;
 * the second one failed because Foreign type proxies don't use the activation key.

Now trying from sumaform.

### DIFF
--- a/salt/proxy_containerized/init.sls
+++ b/salt/proxy_containerized/init.sls
@@ -61,6 +61,19 @@ install_mgr_tools:
       - podman
       - mgrpxy
       - mgrctl
+{% else %}
+{% if grains.get('testsuite') %}
+install_testsuite_helpers:
+  cmd.run:
+    - name: |
+        transactional-update -c run bash -c "zypper mr -e os_pool_repo && zypper -n install expect && zypper mr -d os_pool_repo"
+
+reboot:
+  module.run:
+    - name: system.reboot
+    - at_time: +2
+    - order: last
+{% endif %}
 {% endif %}
 
 # This will only work if the proxy is part of the cucumber_testsuite module, otherwise the server might not be ready


salt/proxy_containerized/init.sls
```diff
@@ -61,6 +61,21 @@ install_mgr_tools:
...
       - mgrpxy
       - mgrctl
+{% else %}
+{% if grains.get('testsuite') %}
+install_testsuite_helpers:
+  cmd.run:
+    - name: |
+        zypper mr -e os_pool_repo
+        transactional-update -c -n package install expect
+        zypper mr -d os_pool_repo
```
aaannz, line 71: Just curious, does this work as expected after reboot? Is repo indeed disabled?

Because, first zypper mr -e writes in the `/etc` snapshot being an overlay upper dir.
Then you do transaction update which makes currently mounted `/etc` as next snapshot lowerdir.
So zypper mr -d after transaction is essentially writing to the overlay lowerdir.

This may be underfined as is stated in [official documentation](https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt):
> Changes to the underlying filesystems while part of a mounted overlay
filesystem are not allowed.  If the underlying filesystem is changed,
the behavior of the overlay is undefined, though it will not result in
a crash or deadlock.

I think running this like:
```
zypper mr -e os_pool_repo
transactional-update -c run "zypper --non-interactive in expect; zypper mr -d os_pool_repo"
```

Bischoff: Apply your solution anyway, with small changes.

Bischoff: > I think running this like:
> 
> ```
> zypper mr -e os_pool_repo
> transactional-update -c run "zypper --non-interactive in expect; zypper mr -d os_pool_repo"
> ```

Does not work, it considers the string in quotes as a command to execute

Trying to find a solution

Bischoff: bash -c seems to do the trick.

Bischoff: ok it works again

rjmateus (approved):

aaannz (approved):

salt/proxy_containerized/init.sls
```diff
@@ -61,6 +61,19 @@ install_mgr_tools:
       - podman
       - mgrpxy
       - mgrctl
+{% else %}
+{% if grains.get('testsuite') %}
+install_testsuite_helpers:
+  cmd.run:
+    - name: |
+        transactional-update -c run bash -c "zypper mr -e os_pool_repo && zypper -n install expect && zypper mr -d os_pool_repo"
```
srbarrios, line 69: Reboot through the salt states it gave issues in the past.
If we already have a proven way to install these helpers from [combustion](https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/host/combustion#L61), I think that we better align the code having all at the same place.

Bischoff: Thanks for the snippets.

I agree that rebooting is calling for trouble. I don't like it much either. Although the "+2 minutes" should let the time for the Salt connection to be closed properly. And it was working in the tests.

One problem with these snippets is that they will install on all machines, not only on the proxy.

salt/proxy_containerized/init.sls
agraul: A cleaner way might be to use `zypper --plus-content X install Y` and leaving the repository disabled

salt/proxy_containerized/init.sls
srbarrios: OK, meaning you prefer to merge this approach or you will make the changes on this PR?

salt/proxy_containerized/init.sls
Bischoff: Interesting option, Alexander.

salt/proxy_containerized/init.sls
Bischoff: I will drop this PR and try your snippets.

Bischoff:
Closing in favour of Oscar's snippets.

PR #1563, opened 2024-04-10
Use Leap Micro in Uyuni Server Containerized

jordimassaguerpla:
## What does this PR change?

Use leap Micro in Uyuni Server Containerized
https://github.com/SUSE/spacewalk/issues/24037


srbarrios (commented):
LGTM

PR #1573, opened 2024-04-18
Fix condition in repo/default for debian or ubuntu

maximenoel8:
## What does this PR change?
Currently, part of repo/default is not executed for ubuntu and debian clients.
Fix the condition to correctly be executed for ubuntu and debian clients.

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -727,6 +727,7 @@ wait_until_apt_lock_file_unlock:
       - interval: 5
       - until: True
 
+# WORKAROUND : default repository from debian10 image are incorrect
 {% if grains['os_family'] == 'Debian' and grains['osmajorrelease']|int() == 10 %}
 debian10_archive_repo:
   file.replace:
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -812,7 +812,6 @@ remove_no_install_recommends:
     - name: /etc/apt/apt.conf.d/00InstallRecommends
 {% endif %}
 
-
 {# WORKAROUND: see github:saltstack/salt#10852 #}
 {{ sls }}_nop:
   test.nop: []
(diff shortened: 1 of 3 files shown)


nodeg (changes requested):
The condition seems correct and is only valid for Ubuntu. So no changes required.

Ubuntu:
```bash
root@suma-bv-50-minssh-ubuntu2204:~# venv-salt-call --local grains.get os_family
local:
    Debian
root@suma-bv-50-minssh-ubuntu2204:~# venv-salt-call --local grains.get os
local:
    Ubuntu 
```

```bash
Debian:
root@suma-bv-50-min-debian11:~# venv-salt-call --local grains.get os_family
local:
    Debian
root@suma-bv-50-min-debian11:~# venv-salt-call --local grains.get os
local:
    Debian
```

salt/repos/default.sls
```diff
@@ -694,7 +694,7 @@ clean_repo_metadata:
 
 {% endif %} {# grains['os_family'] == 'RedHat' #}
 
-{% if grains['os_family'] == 'Debian' and grains['os'] == 'Ubuntu' %}
+{% if grains['os_family'] == 'Debian' or grains['os'] == 'Ubuntu' %}
```
srbarrios, line 697: If we really want to run some salt states for both.
We can just use the family name, without filtering by os name.

```suggestion
{% if grains['os_family'] == 'Debian' %}
```

But I think the tools_update_repo must diverge by os name.

nodeg: Taking a look at the complete `if` statement, I think this is only for Ubuntu and not both. Everything only mentions Ubuntu. See
https://github.com/uyuni-project/sumaform/blob/fa37bc5f3d5862dda442170ba4c9d8bfa18e1eb2/salt/repos/default.sls#L697-L795

nodeg:
With https://github.com/uyuni-project/sumaform/pull/1575 being merged before this PR, we can merge this one as is.

nodeg:
I think we still need this change, @maximenoel8.

PR #1575, opened 2024-04-18
Add tools_update_repo for Debian systems

srbarrios:
## What does this PR change?

We noticed that we don't inject anywhere the tools_update_repo for Debian systems.

### DIFF
--- a/salt/repos/default.sls
+++ b/salt/repos/default.sls
@@ -694,7 +694,7 @@ clean_repo_metadata:
 
 {% endif %} {# grains['os_family'] == 'RedHat' #}
 
-{% if grains['os_family'] == 'Debian' and grains['os'] == 'Ubuntu' %}
+{% if grains['os_family'] == 'Debian' %}
 
 {% set release = grains.get('osrelease', None) %}
 {% set short_release = release | replace('.', '') %}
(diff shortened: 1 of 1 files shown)


salt/repos/default.sls
```diff
@@ -727,6 +727,13 @@ wait_until_apt_lock_file_unlock:
       - interval: 5
       - until: True
 
+{% endif %} {# grains['os_family'] == 'Debian' #}
```
nodeg, line 730: Doesn't this comment break the Jinja templating?

nodeg: Ok, according to https://jinja.palletsprojects.com/en/3.0.x/templates/#comments

```jinja
{# ... #} 
```
> for [Comments](https://jinja.palletsprojects.com/en/3.0.x/templates/#comments) not included in the template output

is fine.

salt/repos/default.sls
```diff
@@ -792,7 +792,71 @@ tools_update_repo_raised_priority:
...
+
+{% if grains['os'] == 'Debian' %}
+
+{% set release = grains.get('osrelease', None) %}
+
+tools_update_repo:
+  pkgrepo.managed:
+    - humanname: tools_update_repo
+    - file: /etc/apt/sources.list.d/tools_update_repo.list
+    # We only have one shared Client Tools repository, so we are using 4.3 for 4.2
```
nodeg, line 805: This comment needs adjustment.

nodeg: Furthermore we have no entry for 5.0?

salt/repos/default.sls
srbarrios: Yes, if you notice, we have it for every of these sections in the rest of family names, it helps us to navigate through this sls.

salt/repos/default.sls
srbarrios: Fixed, being generic, to don't need to adjust it next time.
For now, I think we only have Head right?

nodeg (approved):
Overall this LGTM. There is only the change, Maxime wanted to propose. But with this PR merged before, we can merge his PR as is.

vandabarata (approved):

Bischoff (approved):

NamelessOne91 (approved):

maximenoel8 (approved):

nodeg (approved):

PR #1576, opened 2024-04-22
Large deployment: Tweak the values

srbarrios:
## What does this PR change?

Adjusting some values for our Build Validation setup.
In order to perform all bootstraps and smoke tests in parallel.

### DIFF
--- a/salt/server/large_deployment.sls
+++ b/salt/server/large_deployment.sls
@@ -18,7 +18,7 @@ large_deployment_increase_hibernate_max_connections:
   file.replace:
     - name: /etc/rhn/rhn.conf
     - pattern: 'hibernate.c3p0.max_size = (.*)'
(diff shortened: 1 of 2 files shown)


NamelessOne91 (approved):

salt/server/large_deployment.sls
```diff
@@ -47,7 +47,7 @@ large_deployment_increase_database_max_connections:
   file.replace:
     - name: /var/lib/pgsql/data/postgresql.conf
     - pattern: 'max_connections = (.*)'
-    - repl: 'max_connections = 450'
+    - repl: 'max_connections = 400'
```
nodeg, line 50: Should we really decrease the value for the max. amount of connections?

rjmateus: I was incresing the value. When I entered the BV server the value was on 200, so I doubled it.

PR #1579, opened 2024-05-06
Make arch depend on grains instead of hard coded

ktsamis:
## What does this PR change?

Make arch depend on grains instead of hard coded. Makes sumaform work with MacOS. Basically I did `sed -i 's/x86_64/{{ grains.get("cpuarch") }}/g'` for each file here.

NOTE: This has not been tested except for server Uyuni which works. The rest I have to test, at least one of each kind of node, server, proxy, minion, client. It's a big, potentially destructive change. If we want we can only change the uyuni server and proxy and leave the rest intact as well, which would limit the scope and the potential for breakage as well.

### DIFF
--- a/backend_modules/aws/host/user_data.yaml
+++ b/backend_modules/aws/host/user_data.yaml
@@ -31,7 +31,7 @@ zypper:
   repos:
     - id: os_pool_repo
       name: os_pool_repo
-      baseurl: http://${mirror_url}/SUSE/Products/SLE-Module-Basesystem/15-SP1/x86_64/product
+      baseurl: http://${mirror_url}/SUSE/Products/SLE-Module-Basesystem/15-SP1/{{ grains.get("cpuarch") }}/product
       enabled: 1
       autorefresh: 1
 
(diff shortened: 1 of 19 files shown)


maximenoel8 (approved):
LGTM but I'm not sure those changes are going to be enough when not been in x86_64 arch. We are also setting up some temporary repositories during the cloud init to install some packages ( user_data.yaml ). We probably need to keep an eye on this.

cbbayburt:
> LGTM but I'm not sure those changes are going to be enough when not been in x86_64 arch. We are also setting up some temporary repositories during the cloud init to install some packages ( user_data.yaml ). We probably need to keep an eye on this.

I agree. I think the same change can be applied to the repositories defined in `user_data.yaml` files as well.

ktsamis:
> I agree. I think the same change can be applied to the repositories defined in `user_data.yaml` files as well.

This is done now, ready to be reviewed. Let me know if there is something else that is needed. I think it's ready to be merged.

cbbayburt (approved):

PR #1583, opened 2024-05-14
Add support for SL Micro 6.0 hosts, including Salt Shaker

meaksh:
## What does this PR change?

This PR includes some extra commits on top of https://github.com/uyuni-project/sumaform/pull/1550 in order to:

- Rename from SLEMicro6.0 to SLMicro6.0.
- Use sumaform image for SLMicro6 building in OBS.
- Use combustion instead of cloud init for SL Micro 6.0.
- Add support for Salt Shaker deployment.

/cc @ktsamis for awareness

PR #1587, opened 2024-05-17
Salt Shaker: Fix testsuite path and add GPG keys for SLMicro60

meaksh:
## What does this PR change?

After https://github.com/uyuni-project/sumaform/pull/1583, this PR fixes some detected issues:

- Add GPG keys for "saltstack" repos in OBS to avoid issues in combustion.
- Use a proper path to deploy testsuite on SL Micro 6.0, as content in `/var/tmp/` is not visible outside the transaction.


vzhestkov (approved):
LGTM

PR #1595, opened 2024-05-28
Disable beta channels

srbarrios:
## What does this PR change?

This PR aims to improve how we handle actions related to beta channels, depending if we want them enabled or disabled.

### DIFF
--- a/salt/server/init.sls
+++ b/salt/server/init.sls
@@ -162,8 +162,7 @@ extend_login_timeout:
         - cmd: server_setup
 {% endif %}
 
-# WORKAROUND: 4.4 is needed only until the branching of SUSE Manager 4.4 is completed
-{% if 'head' in grains.get('product_version') or '4.4' in grains.get('product_version') %}
+{% if 'head' in grains.get('product_version') and grains.get('beta_enabled') %}
 change_product_tree_to_beta:
   file.replace:
     - name: /etc/rhn/rhn.conf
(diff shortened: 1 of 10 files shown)


salt/server/init.sls
```diff
@@ -162,8 +162,7 @@ extend_login_timeout:
         - cmd: server_setup
 {% endif %}
 
-# WORKAROUND: 4.4 is needed only until the branching of SUSE Manager 4.4 is completed
-{% if 'head' in grains.get('product_version') or '4.4' in grains.get('product_version') %}
+{% if grains.get('beta_enabled') %}
```
Bischoff, line 165: rather:
```
{% if 'head' in grains.get('product_version') and  if grains.get('beta_enabled') %}
```
no ?

Bischoff (changes requested):
se suggested change

salt/server/init.sls
srbarrios: I'm not sure, could we have a case were the product version is different than head, but we still want to change this product_tree_tag to Beta?

For instance, we have a 'uyuni-pr' product_version that we use for those builts coming from branches that we test in our PR test suite. We were working in having the same option but for SUMA branches. In such cases, maybe we want to give more flexibility about the product version we are deploying.

mcalmer (approved):

salt/server/init.sls
Bischoff: answered in chat. I'd go for a minimal change and therefore keep the head condition

srbarrios: OK, I applied the suggestion.

Bischoff (approved):

PR #1596, opened 2024-05-28
Fix the salt_testenv module to properly run Salt Shaker tests in Leap

ycedres:
## What does this PR change?

Also fixes missing bits for testing the Salt package built in in `systemsmanagement:saltstack`.

### DIFF
--- a/salt/salt_testenv/init.sls
+++ b/salt/salt_testenv/init.sls
@@ -3,7 +3,7 @@ include:
 
 {% if grains['os'] == 'SUSE' %}
 
-{% if grains['osfullname'] == 'SLES' and grains['osrelease_info'][0] == 15  %}
+{% if grains['osfullname'] == 'SLES' and grains['osrelease_info'][0] == 15 %}
 {% set repo_path = "15" if grains["osrelease"] == 15 else "15-SP" + grains["osrelease_info"][1]|string %}
 development_tools_repo_pool:
   pkgrepo.managed:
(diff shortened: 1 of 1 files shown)


salt/salt_testenv/init.sls
```diff
@@ -156,6 +156,10 @@ start_docker_service:
     {% set salt_flavor_path = "saltstack:bundle:testing" %}
 {% elif grains["salt_obs_flavor"] == "saltstack:products:next" %}
     {% set salt_flavor_path = "saltstack:bundle:next" %}
+{% elif grains["salt_obs_flavor"] == "saltstack:testing" %}
```
meaksh, line 159: We do not have this flavor anymore, we removed `saltstack:testing` some time ago

salt/salt_testenv/init.sls
```diff
@@ -156,6 +156,10 @@ start_docker_service:
...
+    {% set salt_flavor_path = "saltstack:bundle:testing" %}
+{% elif grains["salt_obs_flavor"] == "saltstack:next}
```
meaksh, line 161: Same here for `saltstack:next`. We are not using it.

salt/salt_testenv/init.sls
```diff
@@ -3,7 +3,7 @@ include:
 
 {% if grains['os'] == 'SUSE' %}
 
-{% if grains['osfullname'] == 'SLES' and grains['osrelease_info'][0] == 15  %}
+{% if ((grains['osfullname'] == 'SLES' and grains['osrelease_info'][0] == 15) or (grains['osfullname'] == 'Leap')) %}
```
meaksh, line 6: If we set this, then Leap would configure the repositories that we set for SLE, which is not correct and probably not really needed.

meaksh (changes requested):

salt/salt_testenv/init.sls
ycedres: Now I just check that we install the bundle if we're running the Salt Shaker test suite in an environment different from the `systemsmanagement:saltstack` OBS project.

meaksh (approved):

PR #1597, opened 2024-05-29
configure registry instead of uyuni images location

mbussolotto:
## What does this PR change?

configure registry instead of uyuni images location


srbarrios (approved):

srbarrios:
@mbussolotto Approved.
Just for your info, @cbosdo guide me to use that variable, so pinging him so everyone is aware of that change.

rjmateus:
@srbarrios this has changed to have a consistent name between mgradm and mgrpxy.

PR #1598, opened 2024-05-29
configure registry using config.yaml

mbussolotto:
## What does this PR change?

configure registry using config.yaml.
This mean that `/etc/uyuni/config.yaml` should contains:
```
registry: REGISTRY_ADDRESS
````

### DIFF
--- a/salt/proxy_containerized/init.sls
+++ b/salt/proxy_containerized/init.sls
@@ -48,11 +48,12 @@ ssh_config_proxy_containerized:
 {% endif %}
 
 # Useful to setup the proxy through the tests
-env_var_bashrc_registry_proxy_httpd_image:
-  cmd.run:
-    - name: |
-        echo "export REGISTRY={{ container_repository }}" >> /root/.bashrc
-        source /root/.bashrc
+config_proxy_containerized:
+  file.managed:
+    - name: /etc/uyuni/config.yaml
+    - content: |
+      registry: {{ container_repository }}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by Bischoff — APPROVED** [review]

File: salt/proxy_containerized/init.sls [hunk 1]
```diff
@@ -48,11 +48,10 @@ ssh_config_proxy_containerized:
 {% endif %}
 
 # Useful to setup the proxy through the tests
-env_var_bashrc_registry_proxy_httpd_image:
+set_registry_proxy:
   cmd.run:
     - name: |
-        echo "export REGISTRY={{ container_repository }}" >> /root/.bashrc
-        source /root/.bashrc
+        echo "registry: {{ container_repository }}" > /etc/uyuni/config.yaml
```
mcalmer, line 54: Hmm, the directory might not exist

mcalmer (changes requested):
maybe better a file.managed with "makedirs: True" option?

mcalmer (approved):
should work now

salt/proxy_containerized/init.sls
Bischoff: yeah, and `file.managed` is better than `cmd.run` because it is stateful
(`cmd.run` is run every time and might overwrite an existing file).

Bischoff (approved):
lgtm with file.managed, thanks

PR #1601, opened 2024-05-29
Fix url for 15sp6 image

ktsamis:
## What does this PR change?

Fix url for 15sp6 image, the previous is no longer valid. I suspect that the GM final url might change later but for now it's valid.


Bischoff (approved):

ktsamis:
The test fails because of an error `Error: Resource not accessible by integration` nothing to do with the PR. Also updated the images here https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/898/

PR #1609, opened 2024-06-03
Following PR for the integration of SL Micro 6.0

srbarrios:
## What does this PR change?

- Removing wrong basesystem repo
- using a unique repo for client-tools
- filter test packages only for server/proxy
- Fixing sshd.sls issue

### DIFF
--- a/salt/default/init.sls
+++ b/salt/default/init.sls
@@ -8,7 +8,7 @@ include:
   {% if grains.get('reset_ids') | default(false, true) %}
   - default.ids
   {% endif %}
-  {% if not grains['osfullname'] == 'SLE Micro' %}
+  {% if not grains['osfullname'] in ['SLE Micro', 'SL-Micro'] %}
   # Dependencies already satisfied by the images
   # https://build.opensuse.org/project/show/systemsmanagement:sumaform:images:microos
   - default.testsuite
--- a/salt/default/sshd.sls
+++ b/salt/default/sshd.sls
@@ -1,9 +1,10 @@
 {% if 'client' in grains.get('roles') or 'minion' in grains.get('roles') or 'sshminion' in grains.get('roles') %}
-
+# WORKAROUND: Leap 15.6 and SL-Micro 6.0 are using a different sshd_config. To be reviewed.
+{% if not ( grains['osfullname'] in ['Leap', 'SL-Micro'] and grains['osrelease'] in ['15.6', '6.0'] ) %}
 sshd_change_challengeresponseauthentication:
   file.replace:
     - name: /etc/ssh/sshd_config
     - pattern: "^ChallengeResponseAuthentication.*"
     - repl: "ChallengeResponseAuthentication yes"
-
+{% endif %}
 {% endif %}
--- a/salt/repos/init.sls
+++ b/salt/repos/init.sls
@@ -19,7 +19,7 @@ include:
 {% if grains['os'] == 'SUSE' %}
 refresh_repos:
   cmd.run:
-{% if grains['osfullname'] == 'SLE Micro' %}
+{% if grains['osfullname'] in ['SLE Micro', 'SL-Micro'] %}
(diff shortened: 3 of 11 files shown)


srbarrios:
It does not work...
```
... micro/6.0/product/repo/Leap-Micro-6.0-x86_64-Media1/] Valid metadata not found 
```

salt/default/sshd.sls
```diff
@@ -1,9 +1,10 @@
 {% if 'client' in grains.get('roles') or 'minion' in grains.get('roles') or 'sshminion' in grains.get('roles') %}
-
+# WORKAROUND: SLES 15.6 and SL-Micro 6.0 are using a different sshd_config. To be reviewed.
+{% if grains['os'] == 'SUSE' and not grains['osrelease'] in ['15.6', '6.0'] %}
```
srbarrios, line 3: @meaksh I had to do this... make sense for you?
Otherwise, this salt state was failing because that file was not found.
And if I create this, in addition to `/etc/ssh/sshd_config.d/root.conf` then the ssh connection failed.

salt/default/init.sls
```diff
@@ -8,7 +8,7 @@ include:
   {% if grains.get('reset_ids') | default(false, true) %}
   - default.ids
   {% endif %}
-  {% if not grains['osfullname'] == 'SLE Micro' %}
+  {% if not grains['osfullname'] in ['SLE Micro', 'SL-Micro'] and not grains['osfullname'] == 'SL-Micro' %}
```
meaksh, line 11: Shouldn't this be just:

```suggestion
  {% if not grains['osfullname'] == 'SLE Micro' %}
```

salt/repos/init.sls
```diff
@@ -19,7 +19,7 @@ include:
 {% if grains['os'] == 'SUSE' %}
 refresh_repos:
   cmd.run:
-{% if grains['osfullname'] == 'SLE Micro' %}
+{% if grains['osfullname'] in ['SLE Micro', 'SL-Micro'] or grains['osfullname'] == 'SL-Micro' %}
```
meaksh, line 22: I removed the redundant part

```suggestion
{% if grains['osfullname'] in ['SLE Micro', 'SL-Micro'] %}
```

salt/default/sshd.sls
meaksh, line 3: I would try to use a better matching, based on `osfullname` grain instead.

srbarrios: Aha, ok.
But it seems to affect only to SL Micro 6.0 and Leap 15.6

salt/repos/init.sls
srbarrios: I replaced all the matches and it seems some where already refactored.

salt/default/init.sls
srbarrios: I replaced all the matches and it seems some where already refactored.

salt/default/sshd.sls
srbarrios: What about this?
```
{% if not ( grains['osfullname'] in ['Leap', 'SL-Micro'] and grains['osrelease'] in ['15.6', '6.0'] ) %}
```

salt/default/sshd.sls
```diff
@@ -1,9 +1,10 @@
 {% if 'client' in grains.get('roles') or 'minion' in grains.get('roles') or 'sshminion' in grains.get('roles') %}
-
+# WORKAROUND: SLES 15.6 and SL-Micro 6.0 are using a different sshd_config. To be reviewed.
+{% if not ( grains['osfullname'] in ['Leap', 'SL-Micro'] and grains['osrelease'] in ['15.6', '6.0'] ) %}
```
meaksh, line 3: Isn't `osfullname' for Leap actually `openSUSE Leap Micro` ?  And what about SLES ?

meaksh (changes requested):

meaksh (approved):

PR #1611, opened 2024-06-07
Update rocky owner

maximenoel8:
## What does this PR change?

Update rocky8 owner and add rocky9 to ami images.
Also remove sles15sp6 from the ami list ( the image is not available yet and block deployment)

### DIFF
--- a/backend_modules/aws/base/ami.tf
+++ b/backend_modules/aws/base/ami.tf
@@ -124,48 +124,6 @@ data "aws_ami" "sles15sp5-paygo" {
   }
 }
 
-data "aws_ami" "sles15sp6o" {
-  most_recent = true
-  name_regex  = "^suse-sles-15-sp6-byos-v"
-  owners      = ["679593333241"]
-
-  filter {
-    name   = "architecture"
-    values = ["x86_64"]
-  }
-
-  filter {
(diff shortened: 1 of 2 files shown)


nodeg (approved):
I do not understand why all those changes are necessary, but what I see LGTM.

maximenoel8:
I removed sles15sp6 because it's not available yet but instead of seeing it as deleted it shows modification...

mcalmer (approved):

backend_modules/aws/base/ami.tf
```diff
@@ -145,30 +145,10 @@ data "aws_ami" "sles15sp6o" {
   }
 }
 
-data "aws_ami" "sles15sp6-paygo" {
```
srbarrios, line 148: If the data is correct, and we will include it at some point soon, maybe we can just comment this section instead of delete it?

srbarrios (approved):

backend_modules/aws/base/ami.tf
mcalmer: I think we can bring this back when we really know how the name is.

mcalmer:
I need this, so I will merge this now.

PR #1613, opened 2024-06-13
Change urls to http without ssl

ktsamis:
## What does this PR change?

Change urls to http without ssl


srbarrios (approved):
Make sense. thanks.

PR #1617, opened 2024-06-17
Fix issue with Debian 10 and Debian Uyuni Client Tools and refactor Salt Shaker state

meaksh:
## What does this PR change?

This PR does a refactor on the Salt state to deploy the Salt Shaker in order to make it cleaner and also fix several detected issues:

- The Salt state to configure Salt Bundle and Salt classic package are now splited in two different files.
- The main `salt/salt_testenv/init.sls`  contains now the mapping between "OS -> Salt packages to install"
- Ensure the right Salt classic package is installed by forcing upgrade/downgrade from a specific repository to use.
- Disable rebootmgr to not reboot SLMicro 60 at night.
- Fix issues rendering the Salt Shaker state for some OSes.

Additionaly, this PR fixes some issues around Debian 10 backport repository, where repository has been archived but the image still contains references to the old repo, causing errors, and also fixes a wrong path to Uyuni Client Tools for Debian clients.

### DIFF
--- a/salt/salt_testenv/postinstallation.sls
+++ b/salt/salt_testenv/postinstallation.sls
@@ -1,0 +1,26 @@
+{% if grains['os_family'] == 'Suse' and grains['osfullname'] == 'SL-Micro' %}
+copy_salt_classic_testsuite:
+  cmd.run:
+    - name: transactional-update -c run cp -r /usr/lib/python3.{{ grains["pythonversion"][1] }}/site-packages/salt-testsuite /opt/salt-testsuite-classic
+
+copy_salt_bundle_testsuite:
+  cmd.run:
+    - name: transactional-update -c run cp -r /usr/lib/venv-salt-minion/lib/python3.{{ grains["pythonversion"][1] }}/site-packages/salt-testsuite /opt/salt-testsuite-bundle
+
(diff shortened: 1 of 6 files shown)


salt/salt_testenv/postinstallation.sls
```diff
@@ -0,0 +1,26 @@
+{% if grains['os_family'] == 'Suse' and grains['osfullname'] == 'SL-Micro' %}
```
agraul, line 1: Is there a condition that works for both `SLE Micro 5.5` and `SL Micro 6.0`? Right now we don't have `SLE Micro 5.5`, but maybe we want to change that? We can also defer this to when we add 5.5 though...

agraul (approved):
LGTM!

salt/salt_testenv/postinstallation.sls
meaksh: For now, we have other places in sumaform using: `grains['osfullname'] in ["SLE-Micro", "SL-Micro"]`.

But I'm not adding it here right now, because if we use `SLE-Micro` the state would have failed earlier as "Unsupported Salt Shaker OS". I'll just add it here once we add also the support for "SLE-Micro" :+1:

PR #1618, opened 2024-06-17
Salt Shaker: Fix issue with Leap as salt-minion is already preinstalled

meaksh:
## What does this PR change?

In Leap images, the "salt-minion" is always preinstalled, even if `install_salt_bundle = true`.

### DIFF
--- a/salt/salt_testenv/salt_classic_package.sls
+++ b/salt/salt_testenv/salt_classic_package.sls
@@ -95,7 +95,7 @@ install_salt_testsuite:
 {% else %}
   {# HACK: we call zypper manually to ensure right packages are installed regardless upgrade/downgrade #}
   cmd.run:
-    {% if grains["install_salt_bundle"] %}
+    {% if grains["install_salt_bundle"] and not grains['osfullname'] == 'Leap' %}
     - name: zypper --non-interactive in --force --from salt_testing_repo python3-salt salt python3-salt-testsuite
     {% else %}
     - name: zypper --non-interactive in --force --from salt_testing_repo python3-salt salt python3-salt-testsuite salt-minion


salt/salt_testenv/salt_classic_package.sls
```diff
@@ -88,14 +88,15 @@ salt_testing_repo:
...
+{% set salt_minion_is_installed = salt["pkg.info_installed"]("salt-minion").get("salt-minion", False) %}
 install_salt_testsuite:
 {% if grains['os_family'] == 'Suse' and grains['osfullname'] == 'SL-Micro' %}
   cmd.run:
     - name: transactional-update -c -n pkg in python3-salt-testsuite python3-salt-test python3-salt
 {% else %}
   {# HACK: we call zypper manually to ensure right packages are installed regardless upgrade/downgrade #}
   cmd.run:
-    {% if grains["install_salt_bundle"] %}
+    {% if not salt_minion_is_installed %}
```
agraul, line 99: The negation is a bit confusing. What cases do we differentiate?

meaksh: Yeah, it is weird to have the negation and then the else, this just evolved like this after some adjustments. Let me actually refactor this for a cleaner view.

We just need to check if `salt-minion` package needs to be explicitely included in the Zypper call.

m-czernek (approved):
Now looking at the diff, I'd prefer to build the list with the if/else statements and then call the `cmd.run` only once, but this LGTM too.

salt/salt_testenv/salt_classic_package.sls
agraul: It's still not clear to me why we don't need to include `salt-minion.rpm` in one of the branches. What case is that / when is that path followed?

PR #1621, opened 2024-07-05
Add support for 5.0 released product

Bischoff:
## What does this PR change?

This PR adds support for 5.0 released product


nodeg:
What about the other modules? This is only for the containerized proxy.

Bischoff:
> What about the other modules? This is only for the containerized proxy.

i did not find anything similar for containerized server.

nodeg (approved):
Then all good. Thank you.

NamelessOne91 (approved):

PR #1625, opened 2024-07-15
terracumber does not install MU repositories on containerized server

szachovy:
## What does this PR change?

Following the [SUMA interlock](https://confluence.suse.com/display/SUSEMANAGER/Engineering+Interlock+5.0+-+Sign+off?preview=%2F1359085797%2F1386250774%2FSUSE+Manager+5.0+Engnieering+Interlock+v5.0.pdf), the server containerized is deployed on top of SLE Micro 5.5. Therefore, given the grains, additional repositories are not added to the server in the later stages. This PR adds them to ensure their inclusion in the Jenkins 5.0 BV pipelines.

Server and server_containerized modules were treated with `fmt` afterwards.

### DIFF
--- a/modules/server_containerized/main.tf
+++ b/modules/server_containerized/main.tf
@@ -37,15 +37,15 @@ module "server_containerized" {
   volume_provider_settings      = var.volume_provider_settings
 
   grains = {
-    product_version        = var.product_version
-    container_runtime      = var.runtime
-    container_repository   = var.container_repository
-    container_tag          = var.container_tag
-    helm_chart_url         = var.helm_chart_url
-    cc_username            = var.base_configuration["cc_username"]
-    cc_password            = var.base_configuration["cc_password"]
-    mirror                 = var.base_configuration["mirror"]
-    server_mounted_mirror  = var.server_mounted_mirror
+    product_version                = var.product_version
+    container_runtime              = var.runtime
+    container_repository           = var.container_repository
+    container_tag                  = var.container_tag
+    helm_chart_url                 = var.helm_chart_url
+    cc_username                    = var.base_configuration["cc_username"]
+    cc_password                    = var.base_configuration["cc_password"]
+    mirror                         = var.base_configuration["mirror"]
+    server_mounted_mirror          = var.server_mounted_mirror
     server_username                = var.server_username
     server_password                = var.server_password
     java_debugging                 = var.java_debugging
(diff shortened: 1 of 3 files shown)


salt/repos/additional.sls
```diff
@@ -1,5 +1,5 @@
-# Skip Micro OSes, given the additional repos are set up by combustion/ignition/cloud-init
-{% if grains['osfullname'] not in ['SLE Micro', 'openSUSE Leap Micro'] %}
+# Skip Micro OSes, given the additional repos are set up by combustion/ignition/cloud-init, always add them to the server and server_containerized.
+{% if (grains['osfullname'] not in ['SLE Micro', 'openSUSE Leap Micro']) or ('server' in grains.get('roles')) or ('server_containerized' in grains.get('roles')) %}
```
srbarrios, line 2: Well...
Actually when you evaluate `'server' in grains.get('roles')` that gonna be true for both values 'server' and 'server_containerized', but ok, maybe just for the sake of readability.

srbarrios (approved):

modules/server_containerized/main.tf
```diff
@@ -67,6 +67,7 @@ module "server_containerized" {
     disable_auto_bootstrap         = var.disable_auto_bootstrap
     large_deployment               = var.large_deployment
     beta_enabled                   = var.beta_enabled
+    additional_repos               = var.additional_repos
```
Bischoff, line 70: nice catch, thanks

salt/repos/additional.sls
Bischoff: I vote for the
```
grains['roles'] in ['server', 'server_containerized']
```
and removal of parenthesis too.

Bischoff (approved):
thank you

salt/repos/additional.sls
szachovy: I am not sure of that because `grains.get('roles')` is a list not string.

szachovy:
> thank you

Initially I also had like that, but I noticed that `grains['roles']` is a list ([example](https://github.com/uyuni-project/sumaform/blob/80744d906bfb3326e159e318a3a156ba7f090e01/.github/workflows/ci-validation/salt-server-validation#L7))
Thus it (I guess) checks if **sublist** `roles` is in list `['server', 'server_containerized']` but in the list there are only **strings** so it fails of adding repos after primitive sumaform setup with `slemicro55o` server_containerized only.

If we are confident that `roles` list having server would have only one element I can try extract by index, but I am not sure of that. Current solution also seems lengthy to me.

salt/repos/additional.sls
srbarrios: I think you are right :)

PR #1640, opened 2024-07-30
fix additional repos for SLE Micro / openSUSE Micro

jordimassaguerpla:
## What does this PR change?

additional repos should not be installed with salt, because these are set up by combustion


agraul (commented):
Before: all systems that are not "SLE Micro" or "Leap Micro" get additional repos configured

After: all systems that are not "SLE Micro" or "Leap Micro" get additional repos configured, but only if they have the role "server" or "server_containerized".

Is that intended? Don't we lose the possibility to set up non-servers with additional repositories?

jordimassaguerpla:
> Before: all systems that are not "SLE Micro" or "Leap Micro" get additional repos configured
> 
> After: all systems that are not "SLE Micro" or "Leap Micro" get additional repos configured, but only if they have the role "server" or "server_containerized".
> 
> Is that intended? Don't we lose the possibility to set up non-servers with additional repositories?

:thinking: You are right! I think we want additional repos for any system, no matter the role, except for the Micro ones .

I wonder why we have the condition of being a containerized server ....

PR #1641, opened 2024-07-31
fix additional repos, only skip for Micro

jordimassaguerpla:
## What does this PR change?

additional repos should be added no matter the role, but only be skipped if the os is sle micro or opensuse micro


jordimassaguerpla:
This reverts https://github.com/uyuni-project/sumaform/pull/1625 . Additional repos in SLE micro are being added in combustion, otherwise, if you do it with salt, you need a reboot to have them "activated".

jordimassaguerpla:
> This reverts #1625 . Additional repos in SLE micro are being added in combustion, otherwise, if you do it with salt, you need a reboot to have them "activated".

The problem is that repos are being added twice, with combustion and with salt. Thus, applying the salt states gives an error stating that these repos are already added.

However, I see it was added for BV 5.5 ??

PR #1650, opened 2024-08-14
Add `product_version` for feilong Minions

nodeg:
## What does this PR change?

Add `product_version` for feilong Minions

### DIFF
--- a/backend_modules/null/host/variables.tf
+++ b/backend_modules/null/host/variables.tf
@@ -125,3 +125,8 @@ variable "volume_provider_settings" {
   description = "Map of volume-provider-specific settings, see the backend-specific README file"
   default     = {}
 }
+
+variable "product_version" {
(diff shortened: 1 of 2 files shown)


nodeg:
I hope I added the parameter to the correct place.

backend_modules/null/host/variables.tf
```diff
@@ -125,3 +125,8 @@ variable "volume_provider_settings" {
   description = "Map of volume-provider-specific settings, see the backend-specific README file"
   default     = {}
 }
+
+variable "product_version" {
+  description = "One of: 4.2-nightly, 4.2-released, 4.3-nightly, 4.3-released, 4.3-pr, 4.3-beta, 4.3-VM-nightly, 4.3-VM-released, head, uyuni-master, uyuni-released, uyuni-pr"
```
Bischoff, line 130: BTW we will soon have to add 5.0-xxx to all these comments :trollface: .

Bischoff (approved):
See small comment

backend_modules/null/host/variables.tf
nodeg: Yes. I think the description can be shortened and maybe a link to the docs should be added. Otherwise, we will need to edit with every new release.

PR #1651, opened 2024-08-14
SL Micro 6.0 tools for SUSE Manager

Bischoff:
## What does this PR change?

Use different tools for Uyuni and SUSE Manager for SL Micro 6.0.

### DIFF
--- a/backend_modules/libvirt/host/combustion
+++ b/backend_modules/libvirt/host/combustion
@@ -77,7 +77,16 @@ zypper ar http://${ use_mirror_images ? mirror : "download.opensuse.org" }/updat
 
 %{ if image == "slmicro60o" }
 zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SL-Micro/6.0/x86_64/product/ os_pool_repo
+
+# SL Micro 6.0 is being used in both Uyuni and SUSE Manager
+# we do not support yet Leap Micro 6.0, even in Uyuni
+%{ if product_version == "uyuni-master" || product_version == "uyuni-released" || product_version == "uyuni-pr" }
 zypper ar http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Master:/SLMicro6-Uyuni-Client-Tools/SL-Micro6/ client_tools_repo
+%{ else }
+zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-Manager-Tools/15/x86_64/product/ tools_pool_repo
+zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Updates/SLE-Manager-Tools/15/x86_64/update/ tools_update_repo
+%{ endif }
+
 %{ if testsuite }
 zypper ar http://${ use_mirror_images ? mirror : "download.opensuse.org" }/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Pool/rpm/ test_repo_rpm_pool
 %{ endif }


maximenoel8:
Maybe just a question, why are you creating this is_uyuni varaible and not just doing this ? 
```bash
%{ if product_version contains "uyuni" }
```
or
```bash
 if [[ $product_version =~ uyuni ]]; then
```

Bischoff:
> Maybe just one question, why are you creating `is_uyuni` variable and not just doing this ?
> 
> ```shell
> %{ if product_version contains "uyuni" }
> ```

I do not see a "contains" operator in https://developer.hashicorp.com/terraform/language/expressions/operators


> or
> 
> ```shell
>  if [[ $product_version =~ uyuni ]]; then
> ```

The code above is shell, and `product_version` is a terraform variable.

You COULD pass the terraform variable to shell by using this
```
  if [[ "${{  product_version }}" =~ "uyuni" ]]; then
```
but this would get expanded to e.g.
```
  if [[ "4.3-stable" =~ "uyuni" ]]; then
```
which is not really elegant.


More discussions about my attempts here: https://suse.slack.com/archives/C02D12TNYLS/p1723642166193729

Bischoff:
Tested to work on suma-05 with both SUSE Manager and Uyuni products.

backend_modules/libvirt/host/combustion
```diff
@@ -77,7 +77,16 @@ zypper ar http://${ use_mirror_images ? mirror : "download.opensuse.org" }/updat
 
 %{ if image == "slmicro60o" }
 zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SL-Micro/6.0/x86_64/product/ os_pool_repo
+
+# SL Micro 6.0 is being used in both Uyuni and SUSE Manager
+# we do not support yet Leap Micro 6.0, even in Uyuni
+%{ if is_uyuni }
```
nodeg, line 83: Why not use the `product_version` variable here? We should have it also available, e.g. https://github.com/uyuni-project/sumaform/pull/1651/files#diff-e30c08771bf51de88a84878bb3b80efa7ffbb028051231d87a003b84432dfce4R93

Bischoff: It is available indeed, but then you have no terraform function like `regex` or `regexall` to use it.

The problem is explained here https://github.com/hashicorp/terraform-provider-template/blob/master/template/datasource_template_file.go#L125 .

nodeg: But if you use the product version, according to your code, you do not need those terraform functions. AFAICS, you only use them with your newly introduced variable:
```hcl
is_uyuni            = length(regexall("uyuni", local.product_version)) > 0
```
which - again - could be done via the `product_version` variable.

Bischoff: The problem is that terraform and the terraform file provider do not have the same functions.

Reason is that the terraform file provider use whatever functions at the time it was vendored. Not the terraform functions that come with the terraform that you have installed.

See for example https://github.com/hashicorp/terraform-provider-template/issues/82

I know I am just working around terraform weirdness...

nodeg: I see. I have the following solution which does not throw an error on `terraform apply`:

```sls
%{ if product_version == "uyuni-master" || product_version == "uyuni-released" || product_version == "uyuni-pr" }
zypper ar http://${ use_mirror_images ? mirror : "download.opensuse.org"}/repositories/systemsmanagement:/Uyuni:/Master:/SLMicro6-Uyuni-Client-Tools/SL-Micro6/ client_tools_repo
%{ else }
zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Products/SLE-Manager-Tools/15/x86_64/product/ tools_pool_repo
zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Updates/SLE-Manager-Tools/15/x86_64/update/ tools_update_repo
%{ endif }
```
I am running a `terraform apply` right now and check the repository once finished.

Bischoff: ok, by abandoning the idea of a single test, but listing all the cases

why not, i'll apply it

Bischoff: done

maximenoel8 (approved):

nodeg:
I can confirm the new changes work:

```bash
uyuni-min-sles15:~ # zypper lr --uri
Repository priorities are without effect. All enabled repositories share the same priority.

# | Alias              | Name               | Enabled | GPG Check | Refresh | URI
--+--------------------+--------------------+---------+-----------+---------+-------------------------------------------------------------------------------------------------------------------
1 | ca_suse            | ca_suse            | No      | ----      | ----    | http://download.opensuse.org/repositories/SUSE:/CA/15.5/
2 | client_tools_repo  | client_tools_repo  | No      | ----      | ----    | http://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Master:/SLMicro6-Uyuni-Client-Tools/SL-Micro6/
3 | os_pool_repo       | os_pool_repo       | No      | ----      | ----    | http://download.suse.de/ibs/SUSE/Products/SL-Micro/6.0/x86_64/product/
4 | test_repo_rpm_pool | test_repo_rpm_pool | No      | ----      | ----    | http://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Pool/rpm/
uyuni-min-sles15:~ #

uyuni-min-sles15:~ # salt-call --local grains.item product_version
/usr/lib/python3.11/site-packages/salt/utils/http.py:8: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
  import cgi
/usr/lib/python3.11/site-packages/salt/utils/jinja.py:9: DeprecationWarning: 'pipes' is deprecated and slated for removal in Python 3.13
  import pipes
/usr/lib/python3.11/site-packages/salt/grains/core.py:2711: DeprecationWarning: 'locale.getdefaultlocale' is deprecated and slated for removal in Python 3.15. Use setlocale(), getencoding() and getlocale() instead.
  ) = locale.getdefaultlocale()
local:
    ----------
    product_version:
        uyuni-master
```

nodeg (approved):

PR #1653, opened 2024-08-14
Force installation of docker instead of podman-docker in salt_testenv

ycedres:
## What does this PR change?

In the context of the `salt_testenv` module, for some OSes, zypper resolves the `docker` symbol to the `podman-docker` package instead of the `docker` package, which causes the docker service not being available. This happens during the application of the `install_salt_testuiste` state (`podman-docker` is resolved as a dependency of `python3-salt-testsuite` for some OSes).

This PR forces the installation of docker beforehand.

### DIFF
--- a/salt/salt_testenv/salt_classic_package.sls
+++ b/salt/salt_testenv/salt_classic_package.sls
@@ -96,10 +96,15 @@ install_salt_testsuite:
 {% else %}
   {# HACK: we call zypper manually to ensure right packages are installed regardless upgrade/downgrade #}
   cmd.run:
+    {# We install docker first to ensure zypper does not resolve this dependency to podman-docker #}
     {% if salt_minion_is_installed %}
-    - name: zypper --non-interactive in --force --from salt_testing_repo python3-salt salt python3-salt-testsuite salt-minion
+    - name: |
+        zypper --non-interactive in --force docker
+        zypper --non-interactive in --force --from salt_testing_repo python3-salt salt python3-salt-testsuite salt-minion
     {% else %}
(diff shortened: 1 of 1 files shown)


salt/salt_testenv/salt_classic_package.sls
```diff
@@ -96,10 +96,15 @@ install_salt_testsuite:
 {% else %}
   {# HACK: we call zypper manually to ensure right packages are installed regardless upgrade/downgrade #}
   cmd.run:
+    {# We install docker first to ensure zypper does not resolve this dependency to podman-docker #}
     {% if salt_minion_is_installed %}
-    - name: zypper --non-interactive in --force --from salt_testing_repo python3-salt salt python3-salt-testsuite salt-minion
+    - name: |
+        zypper --non-interactive in --force docker
```
agraul, line 102: Can we be sure that `docker` is not resolved to `podman-docker` here?

ycedres: I don't know if there's a way to have a guarantee. When I tried installing directly `docker` I saw zypper selected the desired package. 

Anyway I tried again the hold mechanism you proposed and it worked. This morning I tried it and it didn't work but I might have done something wrong with  the requisites. I have just forced-pushed a new implementation based on holding the `podman-docker` installation.

agraul: `zypper in docker` will install `docker` and not `podman`, `zypper` first matches package names and only if it does not find anything will fall back to matching capabilities.

vzhestkov (approved):
Let's try it :+1:

PR #1665, opened 2024-08-30
Force GPG import in containerized server

nodeg:
## What does this PR change?

Since we cannot confirm anything during the sumaform setup, we need to force the import. Otherwise, sumaform does not succeed because it will wait for confirmation forever.
image


ktsamis (commented):
all good

PR #1667, opened 2024-08-30
roles is a list and we need to check for all exact values we expect

mcalmer:
## What does this PR change?

Using "in" for a list, requires exact matching. This is different to use in match on a string


srbarrios (approved):

nodeg (approved):

nodeg:
Thank you, mc!

PR #1679, opened 2024-09-06
Always force dhcp dns host setup

jordimassaguerpla:
## What does this PR change?
dhcp dns host setup is not happening, so we add a trigger to make it run always.


Bischoff (approved):
thank you, let's try

PR #1680, opened 2024-09-06
Non-interactive mode for transactional-update

Bischoff:
## What does this PR change?

Do not remain blocked on:
```
Resolving package dependencies...

The following package is going to be upgraded:
  venv-salt-minion

1 package to upgrade.
Overall download size: 31.1 MiB. Already cached: 0 B. After the operation, additional 48.9 KiB will be used.

Backend:  classic_rpmtrans
Continue? [y/n/v/...? shows all options] (y):
```


nodeg (approved):
LGTM. I wonder why it worked before...

Bischoff:
> LGTM. I wonder why it worked before...

I wonder that too. Maybe we never had `venv-salt-minion` updates before...

PR #1689, opened 2024-09-10
Defer DHCP on eth1 to fasten deployment

Bischoff:
## What does this PR change?

Defer DHCP on eth1 to fasten deployment


jordimassaguerpla:
SLE micro does not use combustion

Bischoff:
> SLE micro does not use combustion

But it does use network manager. The tests in the second part are about NM.

Ideally, we should also defer in cloud-init case.

jordimassaguerpla (approved):

PR #1693, opened 2024-09-11
Add dummy AppStream repository

NamelessOne91:
## What does this PR change?

Adds a Salt state, for the standard server, setting up a local dummy AppStream repository to be used by the testsuite.


Bischoff (approved):

Bischoff:
@NamelessOne91 do we need a port to 5.0? Probably, no?

PR #1694, opened 2024-09-12
Remove SUSE Manager 4.2

Bischoff:
## What does this PR change?

This PR removes SUSE Manager 4.2

Note: grafana module changes base product and image. I don't think those values were really used, though.


witekest (approved):
The change looks good to me. I think I have never used `grafana` configuration.

modules/grafana/main.tf
```diff
@@ -10,10 +10,10 @@ module "grafana" {
     mirror          = var.base_configuration["mirror"]
     server          = var.server_configuration["hostname"]
     locust          = var.locust_configuration["hostname"]
-    product_version = "4.2-nightly"
+    product_version = "4.3-nightly"
```
srbarrios, line 13: Ok, fine for now.
But I think we need a second refactor in grafana, providing in variables.tf the product_version and the image.
So we can align it with other modules, and use 

```
    product_version           = var.product_version
```
```
  image                    = var.image == "default" ? var.images[var.product_version] : var.image
```

modules/proxy/main.tf
```diff
@@ -1,15 +1,13 @@
...
-    "4.2-nightly"    = "sles15sp3o"
-    "4.2-build_image"= "sles15sp3o"
     "4.3-released"   = "sles15sp4o"
     "4.3-nightly"    = "sles15sp4o"
     "4.3-pr"         = "sles15sp4o"
     "4.3-beta"       = "sles15sp4o"
     "4.3-build_image"= "sles15sp4o"
     "4.3-VM-nightly" = "sles15sp4o"
     "4.3-VM-released"= "sles15sp4o"
+    # Uyuni non-podman deprecated in September 2024:
```
srbarrios, line 10: I would not introduce this comment. It might seems useful now, but it will lose sense in few months and it might be kept as leftover for longer time.

salt/mirror/etc/minima.yaml
```diff
@@ -121,28 +121,16 @@ scc:
     - SLE-Module-Containers15-SP4-Pool
     - SLE-Module-Containers15-SP4-Updates
     # SUSE Manager Server
-    - SLE-Product-SUSE-Manager-Server-4.2-Pool
```
srbarrios, line 124: Would you please take also care of cleaning the already populated content in our mirrors, so we make some disk space?

modules/grafana/main.tf
Bischoff: Given that @witekest "never used this module", maybe we just nuke this module instead?

srbarrios (approved):

Bischoff:
> The change looks good to me. I think I have never used `grafana` configuration.

Then we might as well get rid of this module? Discussion in the other thread please.

salt/mirror/etc/minima.yaml
Bischoff: That's on my TODO already :) . And same for openSUSE 15.4 removal, and same for Ubuntu 24.04 addition. Big works on the mirror ahead :) .

modules/proxy/main.tf
Bischoff: I already have several of them. I will use them as helpers when we remove uyuni-non-podman, so I don't forget any of them.

modules/grafana/main.tf
srbarrios: In our test suite, we install and configure grafana and prometheus through salt formulas.
From the other side, SUMA/Uyuni supports to install it manually, which might be covered this particular automation. But this module it has not been updated to be aligned with the new product changes.
So, from my point of view, I'd agree on removing grafana module in sumaform, in another PR.

Bischoff: ok

If even the specialist (Witold) does not use it, there's little point in maintaining it, indeed.

PR #1696, opened 2024-09-13
Temporary workaround until we backport to 5.0

Bischoff:
## What does this PR change?

Temporary workaround until we backport to 5.0

I chose to add a second case, rather than using a OR, so it becomes very visible, and will also be easier to remove later on. Also, it's more flexible.

### DIFF
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -57,8 +57,40 @@ cucumber_requisites:
       - rubygem-twopence
     - require:
       - sls: repos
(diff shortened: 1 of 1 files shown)


maximenoel8 (approved):
Probably not the most beautiful way to manage it but at lease we clearly separate 5.0 and 4.3.

Bischoff:
> Probably not the most beautiful way to manage it but at lease we clearly separate 5.0 and 4.3.

yes, that was the idea

PR #1700, opened 2024-09-23
Add a file for salt output

maximenoel8:
## What does this PR change?

Write down the salt deployment output in var log file to more easily debug when salt states are failing without the need of rerunning it.

### DIFF
--- a/salt/first_deployment_highstate.sh
+++ b/salt/first_deployment_highstate.sh
@@ -22,7 +22,7 @@ fi
 
 echo "starting first call to update salt and do minimal configuration"
 
-${SALT_CALL} --local --file-root=$FILE_ROOT/ --log-level=info $MODULE_EXEC state.sls default.minimal ||:
+${SALT_CALL} --local --file-root=$FILE_ROOT/ --log-level=info --out-file=/var/log/salt-deployment.log $MODULE_EXEC state.sls default.minimal ||:
 
 NEXT_TRY=0
 until [ $NEXT_TRY -eq 10 ] || ${SALT_CALL} --local test.ping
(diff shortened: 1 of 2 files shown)


salt/first_deployment_highstate.sh
```diff
@@ -22,7 +22,7 @@ fi
 
 echo "starting first call to update salt and do minimal configuration"
 
-${SALT_CALL} --local --file-root=$FILE_ROOT/ --log-level=info $MODULE_EXEC state.sls default.minimal ||:
+${SALT_CALL} --local --file-root=$FILE_ROOT/ --log-level=info --out-file=/var/log/salt-deployment.log $MODULE_EXEC state.sls default.minimal ||:
```
meaksh, line 25: Hmm, I'm not sure here. This would make the actual return from the Salt command execution to not be displayed in the screen but only written in the log file, so if there is an error during deployment I'm force to download that file to see the error. Is this what we really want? 

Maybe we can print the output afterward? So we preserve our actual output but also get it written in the log file? Something like:

```suggestion
${SALT_CALL} --local --file-root=$FILE_ROOT/ --log-level=info --out-file=/var/log/salt-deployment.log $MODULE_EXEC state.sls default.minimal ||:

# Display the output from the Salt execution
cat /var/log/salt-deployment.log
```

What do you think?

meaksh (dismissed):

salt/first_deployment_highstate.sh
maximenoel8: I just checked and indeed it removed the output. I'm fine with the solution.

PR #1708, opened 2024-09-27
New syntax for mirror

Bischoff:
## What does this PR change?

This PR accomodates for the new syntax of the mirror as developed by @NamelessOne91 .

It also fixes the architecture of the Debian and Ubuntu tools to `amd64`.

**Note:** Honestly, I would simply remove this piece of code that allows to create a minima server "out of the box", but that no one uses, and that is probably full of errors.


maximenoel8 (approved):
LGTM, thanks

NamelessOne91 (approved):

PR #1709, opened 2024-09-30
dnf command not available on centos7

maximenoel8:
## What does this PR change?

dnf command not available on centos7 : 
```bash
[root@suma-bv-50-centos7-minion log]# dnf -y install venv-salt-minion avahi nss-mdns qemu-guest-agent
-bash: dnf: command not found
```


maximenoel8:
@srbarrios I need this change for my testing, if it's incorrect let's revert it.

srbarrios (approved):

PR #1710, opened 2024-10-01
Enable salt service ONLY if auto_connect to master is true.

maximenoel8:
## What does this PR change?
Don't start venv-salt-minion or salt-minion service when auto_connect_to_master is set to false.
The file master_configuration is only set when auto connect to master is true so it doesn't make sense to start salt minion service when file is not set.

### DIFF
--- a/salt/minion/init.sls
+++ b/salt/minion/init.sls
@@ -66,6 +66,7 @@ master_configuration:
           - susemanager
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


NamelessOne91 (approved):

srbarrios (approved):

Bischoff (approved):
surprising

meaksh (approved):
LGTM, thanks!

PR #1713, opened 2024-10-02
Move product version declaration to base

maximenoel8:
## Context
Currently in BV, we need to copy/paste for all modules ( server, proxy and minions ) the `product version` variable.
The `product version` is always the same and don't bring information to minion or client modules.
After discussion, we want to declare the product version to only one place and the best place seems to be at the **base module** level.
From my understanding, the product version is only useful for server, proxy and controller.
I didn't really find use case for the product version for minions / clients.

## What does this PR change?
 - Move `product version` declaration at **base module** level.
 - The `product version` is going to be added to **grain template** at the **host module** level and not at the **individual module** level.
 - Modify proxy and server to use the **base configuration** value if `product version` is not set at the **individual module** level.
 - Keep the possibility to overwrite the `product version` at **individual module** level.
 - Update cucumber testsuite module to impact those changes.

*Note*: **individual modules** are modules like `modules/server/main.tf`, `modules/minion/main.tf`, `modules/proxy/main.tf` where **base module** is `modules/base/main.tf` and **host module** is `backend_modules/libvirt/host/main.tf`

## Other possibility

Declare the product version at the server level and use the the server_configuration to spread it to all other modules.
It will require more changes because the server_configuration variable is only declare in minion modules.


modules/server_containerized/main.tf
```diff
@@ -11,6 +11,10 @@ variable "images" {
   }
 }
 
+locals {
+  product_version = var.product_version != null ? var.product_version : var.base_configuration["product_version"]
+}
+
```
srbarrios, line 17: I'm not sure to understand the case where we could need to have the variable in all the modules, in addition to base. We will ever need to override the value in the base?

Shouldn't be enough to have it defined in base, so we use `var.base_configuration["product_version"]` directly in the rest of modules?

maximenoel8: The only situation where I had two product versions was in cloud.
I had server as 4.3-paygo and proxy as 4.3-byos or 4.3-build_image. But it's only related to cloud.
I agree we could simplify and completely block the possibility to overwrite.

meaksh: I would keep this as flexible as possible - as there might be situations where you want to deploy an environment with nodes having different products versions.

meaksh (approved):
Looks good to me. Let's give it a try. Thanks!

modules/server_containerized/main.tf
srbarrios: Fair enough, let's go that way then.

srbarrios (approved):

PR #1721, opened 2024-10-03
Always run test_repo_debian_updates_script_copy salt state

srbarrios:
## What does this PR change?

Not only run this cp command when the `download_ubuntu_repo.sh` file change, as sometimes we might have a new container image even if that file did not changed, and so, we will not copy the file inside the container and the next salt state will fail.


Bischoff (approved):
I don't see a simple way to detect that the container changed, so ok

PR #1724, opened 2024-10-04
Ubuntu 24.04 tools are not released yet (4.3)

Bischoff:
## What does this PR change?

Ubuntu 24.04 tools are not released yet (4.3)

Note: this assumes we get devel tools for 4.3 too. This is currently not the case.


ktsamis (approved):

srbarrios (approved):
Approved.
But let's please wait a bit to clarify it with releng, before merging

Bischoff:
> Approved. But let's please wait a bit to clarify it with releng, before merging

Raul asked Abid if he should create the missing project.

In the meantime, I will merge. If we won't get the project, I'll hack a symlink as you suggested.

PR #1729, opened 2024-10-06
Fix liberty linux9

maximenoel8:
## What does this PR change?

Fix liberty linux 9 deployment issues.
Add new repository to solve perl and wget package resolution.
Use uuidgen to generate the machine ID ( dbus-uuidgen not installed on Liberty Linux 9 )

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -749,6 +749,11 @@ yum_repos:
     enabled: true
     gpgcheck: false
     name: tools_pool_repo
+  ssl_as_updates:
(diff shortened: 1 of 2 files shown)


Bischoff (approved):

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -749,6 +749,11 @@ yum_repos:
     enabled: true
     gpgcheck: false
     name: tools_pool_repo
+  ssl_as_updates:
```
srbarrios, line 752: We should not require this channel.
When exactly this is need?

PR #1731, opened 2024-10-07
Clarify between 43 and 50

Bischoff:
## What does this PR change?

* Cosmetical changes
* Use 5.0 shared tools even on 4.3 branch (Debian and Ubuntu only, for now)
* 5.0 branch should not point to 4.3

Issue:  https://github.com/SUSE/spacewalk/issues/25428

SLE 15, SLE 12, and RHEL 7 + clones need 2 repos instead of one. This will be addressed in another PR.

### DIFF
--- a/salt/repos/client_tools.sls
+++ b/salt/repos/client_tools.sls
@@ -408,7 +408,8 @@ tools_update_repo:
 {% elif 'beta' in grains.get('product_version') | default('', true) %}
 {% set tools_repo_url = 'http://' + grains.get("mirror") | default("download.suse.de/ibs", true) + '/SUSE/Updates/Ubuntu/' + release + '-CLIENT-TOOLS-BETA/x86_64/update/' %}
 {% elif '4.3-nightly' in grains.get('product_version') | default('', true) %}
-{% set tools_repo_url = 'http://' + grains.get("mirror") | default("download.suse.de", true) + '/ibs/Devel:/Galaxy:/Manager:/4.3:/Ubuntu' + release + '-SUSE-Manager-Tools/xUbuntu_' + release %}
+{# 5.0 is intentional here (shared tools) #}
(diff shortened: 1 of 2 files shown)


salt/repos/client_tools.sls
```diff
@@ -84,14 +84,24 @@ tools_pool_repo:
...
     - refresh: True
     - priority: 98
 
+{# TODO due to shared client tools, we also need 5.0 tools in 4.3 branch #}
+
+{% elif '5.0-nightly' in grains.get('product_version') | default('', true) %}
+
+tools_additional_repo:
+  pkgrepo.managed:
+    - baseurl: http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/Devel:/Galaxy:/Manager:/5.0:/SLE12-SUSE-Manager-Tools/SLE_12/
```
szachovy, line 101: SLE12?

szachovy (changes requested):

salt/repos/client_tools.sls
Bischoff: yes, SLE 12 SP 5 LTSS is still supported, even in 5.0

https://documentation.suse.com/suma/5.0/en/suse-manager/client-configuration/supported-features.html

szachovy: Ok, I was thinking you would rather have http://download.suse.de/ibs/Devel:/Galaxy:/Manager:/5.0:/SLE15-SUSE-Manager-Tools/SLE_15/ here.

Closing

ktsamis (approved):
I see nothing wrong but I also am not sure if there is any other point we should be touching with these changes. All this seems good

salt/repos/client_tools.sls
Bischoff: No, SLE 12 and SLE 15 tools are different.

On my side, I was assuming we would have https://download.suse.de/ibs/Devel:/Galaxy:/Manager:/5.0:/SLE12-SUSE-Manager-Tools/images/repo/SLE-12-Manager-Tools-POOL-x86_64-Media1/ , but this repo does not exist. Julio is aware of the issue and we will ask Marina when she's back.

PR #1733, opened 2024-10-08
Add 4.3 devel tools on traditional clients

Bischoff:
## What does this PR change?

This PR declares 4.3 devel tools on traditional clients, in addition to the devel tools shared with 5.0


srbarrios (approved):

ktsamis (approved):

salt/repos/client_tools.sls
```diff
@@ -159,13 +166,20 @@ tools_pool_repo:
...
 {% if '4.3-nightly' in grains.get('product_version') | default('', true) %} {# Devel Tools Repos #}
 
-tools_additional_repo:
+{% if 'client' in grains.get('roles') %}
+tools_trad_repo:
   pkgrepo.managed:
   - baseurl: http://{{ grains.get("mirror") | default("download.suse.de", true) }}/ibs/Devel:/Galaxy:/Manager:/4.3:/SLE15-SUSE-Manager-Tools/images/repo/SLE-15-Manager-Tools-POOL-x86_64-Media1/
   - refresh: True
   - priority: 98
+{% endif %}
```
maximenoel8, line 175: This part should be a else otherwise you will get both repositories for the client

salt/repos/client_tools.sls
```diff
@@ -349,8 +363,21 @@ tools_update_repo:
     {% else %}
       - cmd: suse_res7_key
     {% endif %}
+{% endif %}
```
maximenoel8, line 366: Same here.

maximenoel8 (changes requested):
For me, we will have both 4.3 and 5.0 tool repo for the clients.

maximenoel8:
This file is really complex... What about splitting it into product version file in a folder and having a clienttools.sls calling those files depending on the product version ? So we will have only if for the client type.

salt/repos/client_tools.sls
Bischoff: that's the idea

maximenoel8 (approved):

Bischoff:
> This file is really complex... What about splitting it into product version files in a folder and having a clienttools.sls calling those files depending on the product version ? So we will have only condition in those product version files depending on the client type and we avoid those super nested conditions. This three to four level of conditions will trigger rubocop :-p.

Great minds meet each other: https://github.com/SUSE/spacewalk/issues/25438

Bischoff:
> For me, we will have both 4.3 and 5.0 tool repo for the clients.

For traditional, yes. See https://suse.slack.com/archives/C02DDLYLFMX/p1728389516609459

PR #1737, opened 2024-10-11
Fix problem propagating the "product_version" attribute

meaksh:
## What does this PR change?

I've noticed there was a problem where the `product_version` was not set properly on the deployed nodes when not having a `product_version` definition in the "base" module but instead inside each different nodes.

Like in this example:

```
module "base" {
[... no product_version defined here ...]
}

module "suma-43-srv" {
  source = "./modules/server"

  name = "suma-43-srv"
  product_version = "4.3-nightly"
}

module "suma-43-min-build" {
  source = "./modules/build_host"

  name = "suma-43-min-build"
  product_version = "4.3-released"
  server_configuration = module.suma-43-srv.configuration
  image = "sles15sp4o"
}

module "suma-43-min-host" {
  source = "./modules/host"

  name = "suma-43-min-host"
  image = "sles15sp4o"
}
```

Before this PR, i.a. the `suma-43-srv` node was not deploying correctly, as it got `product_version = ""` and therefore not installing the SERVER stack.

This PR should fix this problem and use the `product_version` from the node if defined, if note take the one from "base" module.

### DIFF
--- a/modules/build_host/main.tf
+++ b/modules/build_host/main.tf
@@ -17,6 +17,7 @@ module "build_host" {
   connect_to_additional_network = true
   roles                         = ["build_host"]
   disable_firewall              = var.disable_firewall
+  product_version               = var.product_version
   grains = {
     mirror                 = var.base_configuration["mirror"]
     server                 = var.server_configuration["hostname"]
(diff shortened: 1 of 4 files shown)


srbarrios (approved):
Thank you!

maximenoel8 (approved):
LGTM, I missed getting the variable to backend host. Sorry .

PR #1742, opened 2024-10-25
Exclude openSUSE Leap Micro from avahi_pkg

mackdk:
## What does this PR change?

This PR avoid applying the `avahi_pkg` state on openSUSE Leap Micro.


agraul:
What I'm currently missing is why we want to skip installing avahi. Is it pre-installed?

PR #1743, opened 2024-10-28
Leap: add dependency for mgr-push

NamelessOne91:
## What does this PR change?

Due to killing the Leap 15.5 reposync in the CI, we need those
dependencies already fulfilled.

See SUSE/spacewalk#24979


Bischoff (approved):
LGTM. I understand there are no ports because this is on Leap only, therefore uyuni.

PR #1746, opened 2024-11-13
Explicitly configure salt-test for Salt Shaker

m-czernek:
## What does this PR change?

On Salt Shaker, we install both non-bundled and bundled Salt test. The bundled salt-test binary has higher priority in `alternatives`. Consequently, non-bundled Salt Shaker pipelines use bundled salt-test, and experience various test fails.

In this PR, we don't rely on the `alternatives` priority, but configure the `salt-test` binary location explicitly.


m-czernek:
Closing in favor of https://github.com/openSUSE/salt-test/pull/5

PR #1748, opened 2024-11-19
The workaround (for bsc#1229740) should no longer be needed

renner:
## What does this PR change?

This patch is removing the `--registry` argument from the call to `mgradm install` that was only in place due to a problem that meanmwhile has been solved (bsc#1229740). Otherwise the fallback to the SUSE registry would not work as expected.


salt/server_containerized/install_mgradm.sls
```diff
@@ -10,10 +10,9 @@ mgradm_config:
 {% set runtime = grains.get('container_runtime') | default('podman', true) %}
 {% set install_cmd = 'kubernetes' if runtime == 'k3s' else 'podman' %}
 
-# WORKAROUND: Setting the registry here via --registry until https://bugzilla.suse.com/show_bug.cgi?id=1229740 is fixed
 mgradm_install:
   cmd.run:
-    - name: mgradm install {{ install_cmd }} --logLevel=debug --config /root/mgradm.yaml --registry {{ grains.get('container_repository') }} {{ grains.get("fqdn") }}
```
mbussolotto, line 16: maybe `{{ grains.get("fqdn") }}` can be kept... `mgradm` should recover correctly the host fqdn, but this might prevents cases when something it's misconfigured.

mbussolotto (approved):
just a little nitpick as comment, but it looks good

salt/server_containerized/install_mgradm.sls
renner: Oh sure, sorry I overlooked it, thanks for checking! I was reading it too fast and misunderstood.

PR #1753, opened 2024-12-04
Add Berlin timezone for container deployment.

maximenoel8:
## What does this PR change?

Currently, the container have a different timezone than server host or controller.
This differences are created some issues (getting the last log not working with the testsuite, using spacewalk is scheduling the event one hour later, etc)
Force the timezone to Berlin timezone.

Related to: https://github.com/SUSE/spacewalk/issues/25607

### DIFF
--- a/salt/server_containerized/mgradm.yaml
+++ b/salt/server_containerized/mgradm.yaml
@@ -38,3 +38,5 @@ admin:
   firstName: Admin
   lastName: Admin
   email: galaxy-noise@suse.de
+tz:
+  Europe/Berlin


salt/server_containerized/mgradm.yaml
```diff
@@ -38,3 +38,5 @@ admin:
   firstName: Admin
   lastName: Admin
   email: galaxy-noise@suse.de
+tz:
+  Europe/Berlin
```
srbarrios, line 42: OK. But.. I'd have choose UTC, if that is available.
Also, did you double-check if that could affect to our environment deployed in other timezones like SLC?

PR #1754, opened 2024-12-06
Add paygo images

maximenoel8:
## What does this PR change?

 - Add the new image searches for AWS to select 5.0 payg server (x86_64 and arm64), 5.0 byos proxy, sles15sp6 byos and  payg images.
 - Add the possibility to directly specify an ami-image
 - Add the possibility to describe address group for the firewall (we will need to add /32 for individual address). This give the possibility to declare only one address for jenkins worker, mirror, etc in NUE (195.135.223.0/24)
 - Disable all salt states for paygo image


modules/proxy_containerized/variables.tf
```diff
@@ -134,3 +134,19 @@ variable "additional_grains" {
   description = "custom grain string to be added to this minion's configuration"
   default     = {}
 }
+
+variable "use_os_released_updates" {
+  description = "Apply all updates from SUSE Linux Enterprise repos"
+  default     = true
```
srbarrios, line 140: Didn't we agree that in our test environments we want to validate using pristine images?
So, not updating to the latest system packages?

maximenoel8: I'm not sure about what we want to do. Look at next comment. This value was set to true in the main.tf file. I just move it to variables.

modules/proxy_containerized/main.tf
```diff
@@ -23,7 +23,7 @@ module "proxy_containerized" {
   base_configuration            = var.base_configuration
   image                         = var.image == "default" || local.product_version == "head" ? var.images[local.product_version] : var.image
   name                          = var.name
-  use_os_released_updates       = true
```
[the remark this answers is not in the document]
srbarrios: If we want to follow this last agreement https://confluence.suse.com/display/SUSEMANAGER/QE+Team+Agreements#QETeamAgreements-PristineImages I would set it as false by default.

**Note**: Before reaching this agreement, I always had in my mind the opposite strategy, so have the systems up-to-date. That might be the reason why this value it was set as true.

PR #1760, opened 2024-12-16
update GPG keys to extended one

mcalmer:
## What does this PR change?

The Uyuni GPG keys was extended. We need to import the extended one now


juliogonzalez (approved):
Seems you took it from `osc signkey systemsmanagement:Uyuni:Master`, so seems to be OK

ktsamis (approved):

PR #1766, opened 2025-01-10
Exclude server proxy from clientool

maximenoel8:
## Context
Deploying 5.0 paygo server and byos proxy is not possible. I'm now working and getting the configuration working.
To do so, I need to be able to run the salt stages. 
Those changes are fixing some of the failing states.

Adding client tool repository in cloud is too complicated for server and proxy. It required to mirror to much data. Plus those repositories are not necessary. They are required for the clients not server and proxy.

## What does this PR change?

Exclude proxy and server from client tools.
Don't disable the repositories in a paygo context ( break the paygo image )


NamelessOne91 (approved):
Thanks Maxime

PR #1771, opened 2025-01-16
Add support for SL Micro 6.1 hosts

szachovy:
## What does this PR change?

This is the continuation of https://github.com/uyuni-project/sumaform/commit/aadfe2e20d74ca1a6ee20c320b91f077eaca0c82 to make it more aligned with https://github.com/uyuni-project/sumaform/pull/1583 fulfilling requirements in https://github.com/SUSE/spacewalk/issues/26216.

To be verified after deployment:
* `salt/default/sshd.sls`
* ...


srbarrios (approved):

ktsamis (approved):

NamelessOne91 (approved):

backend_modules/libvirt/host/user_data.yaml
maximenoel8, line 1: slemicro61o is not supported by user_data.yaml but combustion. Those changes need to be impacted in combustion (if needed). https://github.com/uyuni-project/sumaform/blob/aadfe2e20d74ca1a6ee20c320b91f077eaca0c82/backend_modules/libvirt/host/main.tf#L7

modules/controller/variables.tf
```diff
@@ -62,6 +62,14 @@ variable "slemicro_minion_configuration" {
   }
 }
 
+variable "slmicro_minion_configuration" {
```
maximenoel8, line 65: Do we need a minion for slmicro ? Can't we reuse the slemicro minion. Just a question.

modules/cucumber_testsuite/main.tf
```diff
@@ -332,6 +332,28 @@ module "slemicro_minion" {
   provider_settings = lookup(local.provider_settings_by_host, "slemicro_minion", {})
 }
 
+module "slmicro_minion" {
```
maximenoel8, line 335: Do we need a minion for slmicro ? Can't we reuse the slemicro minion. Just a question.

modules/cucumber_testsuite/main.tf
```diff
@@ -505,6 +528,7 @@ output "configuration" {
     proxy = var.container_proxy ? module.proxy_containerized[0].configuration : module.proxy[0].configuration
     suse_client = module.suse_client.configuration
     slemicro_minion = module.slemicro_minion.configuration
```
maximenoel8, line 530: I checked the controller, we are parsing those slemicro and slmicro minion variable but actually doing nothing with it. 
The slemicro variable is call nowhere.
https://github.com/uyuni-project/sumaform/blob/aed8598bf8f7dbd30324260714823b0142b028d0/modules/controller/variables.tf#L57-L63
If you search in controller/main.tf, we don't use this variable.

We don't declare it in bashrc controller.
https://github.com/uyuni-project/sumaform/blob/f76bf6cc3ce48b6ac042dbbb2bd00c6a149cdea5/salt/controller/bashrc#L5-L13

backend_modules/libvirt/host/user_data.yaml
maximenoel8: Double check if what is already in combustion is not enough.

salt/repos/testsuite.sls
```diff
@@ -17,7 +17,7 @@ uyuni_key_for_fake_packages:
     - name: transactional-update -c run rpm --import http://{{ grains.get("mirror") | default("minima-mirror-ci-bv.mgr.prv.suse.net", true) }}/uyuni.key
 {% endif %}
 
-{% if not (grains['osfullname'] in ['SL-Micro'] and grains['osrelease'] in ['6.0']) and not (grains['osfullname'] in ['openSUSE Leap Micro'] and grains['osrelease'] in ['5.5']) %}
```
maximenoel8, line 20: Should we support both 6.0 and 6.1 ?

maximenoel8 (changes requested):

modules/controller/variables.tf
srbarrios: We can reuse it indeed.
Maybe in a future PR, we can also rename it, to something like suse_micro_minion.

salt/repos/testsuite.sls
srbarrios: Right. But in fact here, what we do is to NOT support this salt state if we are on a transactional system.
So, I think we need to refactor it to cover any kind of transactional system version.

modules/controller/variables.tf
szachovy: There are two options:

1. We treat SLE and SL minions as completely separate entities (like Deblike and RHlike minions), ensuring they share as few formulas as possible between each other.
2. We reuse some parts of the program to simplify the implementation.

Both approaches have pros and cons, but I would choose option number 1. This is because, later on, it might become unclear what is shared and what is not. Additionally, someone unfamiliar with the setup might create a separate PR for it in the future.

modules/cucumber_testsuite/main.tf
szachovy: https://github.com/uyuni-project/sumaform/pull/1771#discussion_r1922032913

modules/cucumber_testsuite/main.tf
szachovy: Isn't it that `MINION` is any suse (SLE, SL, Leap, Tumbleweed etc...) minion and their distinction is specified in BV testsuite? https://github.com/uyuni-project/sumaform/blob/42db68faf74fdbe0003351128444dfccbea761a7/salt/controller/bashrc#L60

modules/controller/variables.tf
srbarrios: SLE Micro and SL Micro it's just a product name rebranding, there is no technical diversion between them. So, we need to conside SLE Micro and SL Micro as the same product, and so the same module on Sumaform.

modules/cucumber_testsuite/main.tf
srbarrios: As far as I'm aware, we don't deploy/test any SL/SLE Micro on CI Testsuite, we only do it as part of the BV Testsuite.
So, it might not make sense to have this entry on the controller, still it might still interesting to have the ability to deploy it through sumaform or we can even consider cover sl micro in our CI testsuite.

backend_modules/libvirt/host/user_data.yaml
```diff
@@ -421,6 +421,93 @@ runcmd:
...
+  - [ transactional-update, --continue, --non-interactive, pkg, install, mgrpxy, ca-certificates-suse ]
+%{ endif }
+%{ if install_salt_bundle }
+  - [ transactional-update, --continue, --non-interactive, pkg, install, qemu-guest-agent, venv-salt-minion, avahi ]
+%{ else }
+  - [ transactional-update, --continue, --non-interactive, pkg, install, qemu-guest-agent, salt-minion, avahi ]
+%{ endif }
+  - [ reboot ]
+%{ endif }
+
```
nodeg, line 510: As already mentioned, all those "cloud-init" changes need to go to the combustion configuration and not to `user_data.yaml`, since starting with SL Micro 6.0, combustion is used instead of `cloud-init`. See https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/host/combustion

backend_modules/libvirt/host/user_data.yaml
szachovy: Seems like with combustion, changes for transactional systems in `user_data.yaml` applies the same changes for the second time _(?)_

fix done

backend_modules/libvirt/host/user_data.yaml
szachovy: Done https://github.com/uyuni-project/sumaform/pull/1771#discussion_r1922483388

salt/repos/testsuite.sls
szachovy: refactored

backend_modules/libvirt/host/user_data.yaml
maximenoel8: If it's the case, something wrong in sumaform.
In sumaform logic we clearly disable user_data when using combustion
https://github.com/uyuni-project/sumaform/blob/aadfe2e20d74ca1a6ee20c320b91f077eaca0c82/backend_modules/libvirt/host/main.tf#L49
https://github.com/uyuni-project/sumaform/blob/aadfe2e20d74ca1a6ee20c320b91f077eaca0c82/backend_modules/libvirt/host/main.tf#L138

modules/cucumber_testsuite/main.tf
maximenoel8: Be careful not to mix CI and BV. What you are pointing is a BV variable ( using minion modules). What I was pointing was CI configuration (using cucumber testsuite module). You may now understand why I want to normalize CI and BV to have the same approach. All those different behaviors mixed in one file is confusing...

Currently, if you declare sle_micro in your cucumber module, this variable will do nothing because it's not present in bashrc.

Some explanation:

In the module output:
https://github.com/uyuni-project/sumaform/blob/fb4381b63bae9d2644b692b46fe25c0e6d113d12/modules/cucumber_testsuite/main.tf#L504-L519
We see suse_minion (MINION in bashrc, I know, we need some standardization) declare along side slemicro_minion. For me, having the possibility to declare those two variables should result with the support of this two minions and not a replacement of MINION.

My approach will be, don't add sl_micro variable to cucumber testsuite. Create an new issue to address the sle micro refactor needed (don't mix it in this PR).

modules/controller/variables.tf
maximenoel8: You should remove this variable and readdress it in different issue (that will also tackle sle_micro variable).
For the differences, like Oscar say. Same type of product different name.

modules/cucumber_testsuite/main.tf
maximenoel8: I will remove this variable that just add noise to cucumber module.

modules/cucumber_testsuite/main.tf
```diff
@@ -471,6 +493,7 @@ module "controller" {
   minion_configuration           = contains(local.hosts, "suse_minion") ? module.suse_minion.configuration : { hostnames = [], ids = [], ipaddrs = [], macaddrs = [], private_macs = [] }
   sshminion_configuration        = contains(local.hosts, "suse_sshminion") ? module.suse_sshminion.configuration : { hostnames = [], ids = [], ipaddrs = [], macaddrs = [], private_macs = [] }
   slemicro_minion_configuration  = contains(local.hosts, "slemicro_minion") ? module.slemicro_minion.configuration : { hostnames = [], ids = [], ipaddrs = [], macaddrs = [], private_macs = [] }
+  slmicro_minion_configuration   = contains(local.hosts, "slmicro_minion") ? module.slmicro_minion.configuration : { hostnames = [], ids = [], ipaddrs = [], macaddrs = [], private_macs = [] }
```
maximenoel8, line 496: Same here

modules/cucumber_testsuite/main.tf
szachovy: done, re-review please

modules/cucumber_testsuite/main.tf
szachovy: done, re-review please

modules/controller/variables.tf
szachovy: Done, I don't want to mix renaming with adding support in one PR.
Here is the separate card for this purpose: https://github.com/SUSE/spacewalk/issues/26268

maximenoel8 (approved):
Thanks for the work !

PR #1778, opened 2025-02-20
Salt Shaker: allow testing "python311-salt" in SLMicro and Tumbleweed

meaksh:
## What does this PR change?

This PR just enhances the mechanism to install the classic Salt package on Salt Shaker environments, so Python 3.11 is used for SLMicro and Tumbleweed

### DIFF
--- a/salt/salt_testenv/salt_classic_package.sls
+++ b/salt/salt_testenv/salt_classic_package.sls
@@ -89,10 +89,19 @@ salt_testing_repo:
     - gpgkey: http://{{ grains.get("mirror") | default("download.opensuse.org", true) }}/repositories/systemsmanagement:{{ grains["salt_obs_flavor"] }}/{{ repo_path }}/repodata/repomd.xml.key
 
 {% set salt_minion_is_installed = salt["pkg.info_installed"]("salt-minion").get("salt-minion", False) %}
+
+{% if grains['os_family'] == 'Suse' and grains['osfullname'] in ['SL-Micro', 'openSUSE Tumbleweed'] %}
+{% set salt_classic_pkgs = "python311-salt-testsuite python311-salt" %}
+{% set salt_test_executor = "python311-salt-test" %}
+{% else %}
+{% set salt_classic_pkgs = "python3-salt-testsuite python3-salt" %}
+{% set salt_test_executor = "python3-salt-test" %}
+{% endif %}
+
 install_salt_testsuite:
 {% if grains['os_family'] == 'Suse' and grains['osfullname'] == 'SL-Micro' %}
   cmd.run:
-    - name: transactional-update -c -n pkg in python3-salt-testsuite python3-salt-test python3-salt
+    - name: transactional-update -c -n pkg in {{ salt_classic_pkgs }} {{ salt_test_executor }}
 {% else %}
   {# HACK: we call zypper manually to ensure right packages are installed regardless upgrade/downgrade #}
   cmd.run:
@@ -100,11 +109,11 @@ install_salt_testsuite:
     {% if salt_minion_is_installed %}
     - name: |
         zypper --non-interactive in --force docker
-        zypper --non-interactive in --force --from salt_testing_repo python3-salt salt python3-salt-testsuite salt-minion
+        zypper --non-interactive in --force --from salt_testing_repo {{ salt_classic_pkgs }} salt salt-minion
     {% else %}
     - name: |
         zypper --non-interactive in --force docker
-        zypper --non-interactive in --force --from salt_testing_repo python3-salt salt python3-salt-testsuite
+        zypper --non-interactive in --force --from salt_testing_repo {{ salt_classic_pkgs }} salt
     {% endif %}
     - fromrepo: salt_testing_repo
 {% endif %}
@@ -115,7 +124,7 @@ install_salt_testsuite:
 install_salt_tests_executor:
 {% if grains['os_family'] == 'Suse' and grains['osfullname'] == 'SL-Micro' %}
   cmd.run:
-    - name: transactional-update -c -n pkg in python3-salt-test
+    - name: transactional-update -c -n pkg in {{ salt_test_executor }}
 {% else %}
   pkg.installed:
     - pkgs:


agraul (changes requested):
I would prefer an approach that does not require manual changes when Tumbleweed or MicroOS changes the primary Python version (assuming we'll start testing MicroOS as well at some point).

Zypper can use `--capability` to work on rpm provides, for package installations that's the default fallback and does not require these changes. `transactional-update pkg install` might take the same flag, that's something we can check.

meaksh:
@agraul I just took this other approach to not hardcode the primary interpreter version (which we are already using at https://github.com/uyuni-project/sumaform/blob/master/salt/salt_testenv/postinstallation.sls). For now, I just want to fix the current failures we have in our Jenkins jobs for SLMicro 6.0. What do you think?

agraul:
I still don't get the need to specify any version. Can you point me to a deployment failure log where I can see what's being installed and why that's not the right package?

meaksh:
The current problem on SLMicro60, is that `python311-salt` is now part of the VM image (it is preinstalled). The current preinstalled package is our initial iteration on the singlespec approach, that got released already, which contains an `Obsoletes: python3-salt < %{version}`.

So, given this `python311-salt` is preinstalled, then the sumaform deployment fails here when trying to install `python3-salt` here. We see:

```console
(remote-exec): Problem: 1: the installed python311-salt-3006.0-9.1.x86_64 obsoletes 'python3-salt < 3006.0-9.1' provided by the to be installed python3-salt-3006.0-8.1.x86_64
(remote-exec):  Solution 1: Following actions will be done:
(remote-exec):   deinstallation of python311-salt-3006.0-9.1.x86_64
(remote-exec):   downgrade of salt-3006.0-9.1.x86_64 to salt-3006.0-8.1.x86_64
(remote-exec):   downgrade of salt-minion-3006.0-9.1.x86_64 to salt-minion-3006.0-8.1.x86_64
(remote-exec):   downgrade of salt-transactional-update-3006.0-9.1.x86_64 to salt-transactional-update-3006.0-8.1.x86_64
(remote-exec):  Solution 2: do not install python3-salt-3006.0-8.1.x86_64
(remote-exec): Choose from above solutions by number or cancel [1/2/c/d/?]      
```
Jenkins job: https://ci.suse.de/user/manager/my-views/view/Salt%20Shaker/job/manager-salt-shaker-products-testing-slmicro60/

agraul:
If I understand that output correctly, our problem is that we have an older release in our devel repository. That raises the question: don't we want to enforce that our package is installed? With your proposed changes, which Salt package will be installed at the end? The pre-installed package or the package from our `salt_testing_repo`?

agraul:
> > If I understand that output correctly, our problem is that we have an older release in our devel repository.
> 
> Not exactly... indeed we have a higher version-release in our `salt_testing_repo`, but the thing is that, before our singlespec changes, the `salt_testing_repo` was only building `python3-salt` package for SLMicro 6.0, while now our singlespec package builds `python311-salt` only for SLMicro 6.0.
> 

Are you sure the version in `salt_testing_repo` is higher? What you posted says that the to-be-installed package has a lower version-release.

> Problem: 1: the installed python311-salt-3006.0-9.1.x86_64 obsoletes 'python3-salt < 3006.0-9.1' provided by the to be installed python3-salt-3006.0-8.1.x86_64

> Yes, sure, with my proposal I just want to get the python311-salt installed coming from the salt_testing_repo (instead of "python3-salt"), for SLMicro 6.0 and TW.

And this does not work with `--capability`? That's too bad, this workaround is brittle and will require future changes.

meaksh:
> > > If I understand that output correctly, our problem is that we have an older release in our devel repository.
> > 
> > 
> > Not exactly... indeed we have a higher version-release in our `salt_testing_repo`, but the thing is that, before our singlespec changes, the `salt_testing_repo` was only building `python3-salt` package for SLMicro 6.0, while now our singlespec package builds `python311-salt` only for SLMicro 6.0.
> 
> Are you sure the version in `salt_testing_repo` is higher? What you posted says that the to-be-installed package has a lower version-release.

Yes, even it this sounds strange, it is actually the higher version:
```console
salt-shaker-products-testing-slmicro60:~ # zypper search -s python311-salt
Loading repository data...
Reading installed packages...

S  | Name                     | Type    | Version     | Arch   | Repository
---+--------------------------+---------+-------------+--------+---------------------------------
v  | python311-salt           | package | 3006.0-78.1 | x86_64 | salt_testing_repo
i  | python311-salt           | package | 3006.0-9.1  | x86_64 | alp_sources_repo
i  | python311-salt           | package | 3006.0-9.1  | x86_64 | os_pool_repo
   | python311-salt-test      | package | 1.4-1.2     | noarch | salt_testsuite_dependencies_repo
   | python311-salt-testsuite | package | 3006.0-78.1 | x86_64 | salt_testing_repo
   | python311-salt-testsuite | package | 3006.0-9.1  | x86_64 | alp_sources_repo

    Note: For an extended search including not yet activated remote resources please use 'zypper
    search-packages'.
```
> 
> > Problem: 1: the installed python311-salt-3006.0-9.1.x86_64 obsoletes 'python3-salt < 3006.0-9.1' provided by the to be installed python3-salt-3006.0-8.1.x86_64
> 
> > Yes, sure, with my proposal I just want to get the python311-salt installed coming from the salt_testing_repo (instead of "python3-salt"), for SLMicro 6.0 and TW.
> 
> And this does not work with `--capability`? That's too bad, this workaround is brittle and will require future changes.

It doesn't work with `--capability python3-salt`, I must set the exact version to get the resolution to work as we expect. I can prepare a patch that would work that way.

It does not work with:
```console
# transactional-update -c -n pkg in --capability python3-salt-testsuite python3-salt-test python3-salt
-failed, it says nothing to update-
```
```console
# transactional-update -c -n pkg in --capability python3-salt-testsuite-3006.0-78.1 python3-salt-test python3-salt-3006.0-78.1
-successfull installation of python311-3006.0-78.1 using capabilities-
```

I'll push an extra commit with an alternative approach based on `--capability` usage. Thanks  :+1:

meaksh:
@agraul any chance to review the latest changes here? TIA!

agraul:
> @agraul with this latest change, we use `--capability` and it works, but it is really hacky the way to get the appropiated salt version, as jinja2 variables are rendered before we setup all the respositories, so we cannot really use it.

The way I understood this message was that you don't want to use this approach. I don't like this more than the previous implementation, for me they are on similar levels of hacky.

If I may suggest a last alternative, what happens if we first remove the pre-installed Salt packages and then do an unversioned `zypper install --capability python3-salt-testsuite`? Shouldn't that install newest version from our devel repo (assuming we set up the priorities correctly)?

m-czernek:
I've implemented and tested Alex's suggestion in https://github.com/uyuni-project/sumaform/pull/1834

Seems to work for me, simply adding `-C` means that zypper won't force the installation of the 3.6 version of salt since another version of salt is already installed.

meaksh:
> I've implemented and tested Alex's suggestion in #1834
> 
> Seems to work for me, simply adding `-C` means that zypper won't force the installation of the 3.6 version of salt since another version of salt is already installed.
> 
> @meaksh WDYT about closing this PR in favor of #1834 ? Is that solution acceptable for you?

Thanks Marek for looking into this!

I remember trying with `--capability` and it didn't work back then to me, as I mentioned in https://github.com/uyuni-project/sumaform/pull/1778#issuecomment-2681653176.

However, if this is now working and we are sure we get the right package version (coming from `salt_testing_repo`) installed, then I'm definitely fine going in favor of this PR.

meaksh:
> I've implemented and tested Alex's suggestion in #1834
> 
> Seems to work for me, simply adding `-C` means that zypper won't force the installation of the 3.6 version of salt since another version of salt is already installed.
> 
> @meaksh WDYT about closing this PR in favor of #1834 ? Is that solution acceptable for you?

Thanks Marek for looking into this!

I remember trying with `--capability` and it didn't work back then to me, as I mentioned in https://github.com/uyuni-project/sumaform/pull/1778#issuecomment-2681653176.

However, if this is now working and we are sure we get the right package version (coming from `salt_testing_repo`) installed, then I'm definitely fine going in favor of this PR.

m-czernek:
You are correct, I missed the first part of Alex's suggestion. I've fixed it and tested it, the installed version is now correct.

PR #1780, opened 2025-02-25
QE: Remove debian 11 because is deprecated

Pablogoliva:
## What does this PR change?

In this PR debian 11 is remove from our sumaform for its end of life
Fixes: https://github.com/SUSE/spacewalk/issues/26311


ktsamis (approved):

jordimassaguerpla:
The failure we see in "Check updated image URLs" is not a blocker. That workflow writes a comment to the PR, but given is a fork , it can't write to this PR because of not having permissions. So , you can ignore it.

PR #1789, opened 2025-03-09
fix Multi-Linux-Manager repository URLs after product rename

mcalmer:
## What does this PR change?

The products were renamed from "SUSE-Multi-Linux-Manager-*" into "Multi-Linux-Manager-*".
This effect also the repository URLs

### DIFF
--- a/backend_modules/libvirt/host/combustion
+++ b/backend_modules/libvirt/host/combustion
@@ -110,10 +110,10 @@ zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/SUSE/Pr
 zypper ar http://${ use_mirror_images ? mirror : "download.opensuse.org" }/repositories/systemsmanagement:/Uyuni:/Master:/ContainerUtils/openSUSE_Leap_16.0/ container_utils
 %{ else }
 %{ if container_server }
-zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/Devel:/Galaxy:/Manager:/Head/images/repo/SUSE-Multi-Linux-Manager-Server-5.1-x86_64/ container_utils
+zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/Devel:/Galaxy:/Manager:/Head/images/repo/Multi-Linux-Manager-Server-5.1-x86_64/ container_utils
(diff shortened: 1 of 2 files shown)


backend_modules/libvirt/host/combustion
```diff
@@ -122,7 +122,11 @@ zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/Devel:/
 zypper ar http://${ use_mirror_images ? mirror : "download.opensuse.org" }/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Pool/rpm/ test_repo_rpm_pool
 
 # Leap repos are required to install expect
-zypper ar --no-gpgcheck http://${ use_mirror_images ? mirror : "download.opensuse.org" }/distribution/leap/16.0/repo/oss/ leap_pool_repo
+%{ if use_mirror_images }
+zypper ar --no-gpgcheck http://${ mirror }/distribution/leap/16.0/repo/oss/ leap_pool_repo
+%{ else }
+zypper ar --no-gpgcheck http://download.opensuse.org/distribution/leap/16.0/repo/oss/x86_64 leap_pool_repo
```
Bischoff, line 128: I think this is wrong, there's no x86_84

look at http://download.opensuse.org/distribution/leap/16.0/repo/oss/

Bischoff (changes requested):
I think there's an error

Bischoff (commented):
I don't think we should change our paths because of that funny case.

I think it's likely to be an error from OBS, we should clarify it with them.

backend_modules/libvirt/host/combustion
```diff
@@ -122,7 +122,11 @@ zypper ar http://${ use_mirror_images ? mirror : "download.suse.de/ibs"}/Devel:/
...
+%{ endif }
```
Bischoff, line 129: See my comment.

mcalmer: Well, this exists and it is currently the only way how I can make it work.
I have added Lubos to this slack thread and we will find out.

mcalmer:
@Bischoff I have removed the PR with the arch repo

Bischoff (approved):
happy to see the hack away :grin:

PR #1792, opened 2025-03-12
Fix beta_enabled variable on cucumber_testsuite module + Refactor rhn.sls on server_containerized

srbarrios:
## What does this PR change?

Card: https://github.com/SUSE/spacewalk/issues/26373

Fix `beta_enabled` variable on cucumber_testsuite module.
With those changes, the aim is to provide `beta_enabled` variable at cucumber_testsuite module, so you don't need to define it twice in `controller` and `server_containerized` submodules.

Additionally, we fixed the rhn.sls that was previously NOT triggered, refactoring it so it use mgrctl to run some `sed` commands to modify RHN configuration file.

Note for MLM team:
This PR must be follow up by a PR to adapt our internal terraform files in susemanager-ci repository.
Add this variable at cucumber_testsuite level.

### DIFF
--- a/modules/cucumber_testsuite/main.tf
+++ b/modules/cucumber_testsuite/main.tf
@@ -73,8 +73,6 @@ locals {
     host_key => lookup(var.host_settings[host_key], "repository_disk_use_cloud_setup", null) if var.host_settings[host_key] != null }
   scc_access_logging        = { for host_key in local.hosts :
     host_key => lookup(var.host_settings[host_key], "scc_access_logging", false) if var.host_settings[host_key] != null }
-  beta_enabled              = { for host_key in local.hosts :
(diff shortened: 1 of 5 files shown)


modules/cucumber_testsuite/main.tf
```diff
@@ -105,7 +105,7 @@ module "server" {
   forward_registration           = false
   monitored                      = true
   use_os_released_updates        = false
-  beta_enabled                   = lookup(local.beta_enabled, "server", false)
```
NamelessOne91, line 108: If we no longer read it from local, can we get rid of the beta_enabled mapping at line 76 ?

NamelessOne91 (approved):

srbarrios:
```
oscar-server:~ # venv-salt-call --local grains.get beta_enabled
local:
    True
```
```
oscar-controller:~ # venv-salt-call --local grains.get beta_enabled
local:
    True
```

PR #1795, opened 2025-03-12
Changes needed for separate database container

aaannz:
## What does this PR change?

* postgresql container uses `reportdb` hostname which is also default so do not set it in the mgradm.yaml

### DIFF
--- a/salt/server_containerized/mgradm.yaml
+++ b/salt/server_containerized/mgradm.yaml
@@ -1,8 +1,6 @@
 {% set runtime = grains.get('container_runtime') | default('podman', true) %}
 db:
   password: spacewalk
-reportdb:
-  host: localhost
 ssl:
   password: spacewalk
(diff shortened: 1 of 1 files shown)


aaannz:
It is enough to remote this as:
* 5.0 defaults to `localhost`, see https://github.com/SUSE/uyuni-tools/blob/053c629100f652654e1ae05fcc64bbc87e7af891/mgradm/cmd/install/shared/shared.go#L118
* 5.1 defaults to `reportdb`, see https://github.com/uyuni-project/uyuni-tools/pull/482/files#diff-dc8c8e232fe2324f7f50ff66417bdd40179eb0bf728a53be216d927df21ac9d5R128

srbarrios (approved):
LGTM. Thank you

nodeg (approved):
Thank you, Ondrej.

PR #1798, opened 2025-03-18
Pin all GitHub actions to a commit

juliogonzalez:
## What does this PR change?

One of the recommendations of the [Good security practices for using GitHub Actions features](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions?learn=getting_started).

For each action I went to the repository and used the last released version for the major version we were using, and specified the exact version with a comment.

If we were using `latest`, I pinned to the latest commit for the most recent version released.


nodeg (approved):
LGTM, thank you, Julio.

PR #1801, opened 2025-03-20
QE: Removing ubuntu20.04 from uyuni HEAD

Pablogoliva:
## What does this PR change?

REmoves ubuntu 20.04 just from uyuni HEAD


backend_modules/aws/host/user_data.yaml
```diff
@@ -218,41 +218,4 @@ runcmd:
 packages: ["venv-salt-minion"]
 %{ endif }
 
-%{ if image == "ubuntu2004" }
```
Bischoff, line 221: Are we sure AWS is run only with head and uyuni ?

Bischoff (commented):
see question

backend_modules/aws/host/user_data.yaml
Pablogoliva: The source link indicates that this just for uyuni. Am I right or should I leave it??

ktsamis: No this is is not from head, this will affect all testsuites. We use for example if 4.3 BV AWS here https://github.com/SUSE/susemanager-ci/blob/master/terracumber_config/tf_files/SUSEManager-4.3-build-validation-AWS.tf#L328

Pablogoliva: Write it back in the code

ktsamis (approved):
Now it looks good, the only thing is I didn't check if there are more places to remove stuff

PR #1803, opened 2025-03-20
Add support to 5.1 released

maximenoel8:
## What does this PR change?

Add 5.1 released product version declaration.
Add 5.1 repository to combustion to install the correct uyuni tool.

### DIFF
--- a/modules/proxy_containerized/main.tf
+++ b/modules/proxy_containerized/main.tf
@@ -6,6 +6,8 @@ variable "images" {
     "uyuni-master"   = "leapmicro55o"
     "uyuni-released" = "leapmicro55o"
     "uyuni-pr"       = "leapmicro55o"
+    "5.1-released"   = "slmicro61o"
+    "5.1-nightly"    = "slmicro61o"
   }
 }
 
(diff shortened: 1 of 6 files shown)


NamelessOne91 (approved):

modules/proxy_containerized/main.tf
```diff
@@ -6,6 +6,8 @@ variable "images" {
     "uyuni-master"   = "leapmicro55o"
     "uyuni-released" = "leapmicro55o"
     "uyuni-pr"       = "leapmicro55o"
+    "5.1-released"   = "slmicro61o"
+    "5.1-nightly"    = "slmicro61o"
```
nodeg, line 10: Small nitpick. I would move these below the `5.0` values as you did with the controller `main.tf.

modules/server_containerized/main.tf
```diff
@@ -8,6 +8,8 @@ variable "images" {
     "uyuni-master"   = "leapmicro55o"
     "uyuni-released" = "leapmicro55o"
     "uyuni-pr"       = "leapmicro55o"
+    "5.1-released"   = "slmicro61o"
+    "5.1-nightly"    = "slmicro61o"
```
nodeg, line 12: Small nitpick. I would move these below the `5.0` values as you did with the controller `main.tf.

srbarrios (approved):
LGTM. But I wonder if we should have a 5.1-beta too. 🤔

maximenoel8:
> LGTM. But I wonder if we should have a 5.1-beta too. 🤔

It makes sense to have a beta tag if there is something to add behind. For now I only see one totest repo.

PR #1808, opened 2025-03-30
change controller default image to Leap 15.6

mcalmer:
## What does this PR change?

Leap 15.5 is EOL and the ruby repo was already removed


mcalmer:
Tested with ref-head. It works.

Bischoff (approved):
We would have preferred to delay this as much as we could

But if ruby packages leave us no choice...

PR #1810, opened 2025-03-31
Liberty 9: put the IDs in the correct files

Bischoff:
## What does this PR change?

Fix commit 04b274851dae1f3f783c9dd8a55c82f9fbf55227

On Liberty 9, Put the IDs in the correct files:
 * the output of `uuidgen` to `/var/lib/dbus/machine-id`
 * the output of `systemd-machine-id-setup` to `/etc/machine-id`


ktsamis (approved):

Pablogoliva (approved):
LGTM

meaksh (approved):

Bischoff:
> Looks good, thanks! Nitpick: maybe worth considering the refactor of the "if-else" on the SLS to a simpler form, to avoid duplicated code. Something like: (...)

Thanks, applied your suggestion.

agraul:
Looking at https://www.freedesktop.org/software/systemd/man/253/systemd-machine-id-setup.html, `systemd-machine-id-setup` copies over the d-bus uuid. So these two should be the same with this change, is that correct?

Bischoff:
> Looking at https://www.freedesktop.org/software/systemd/man/253/systemd-machine-id-setup.html, `systemd-machine-id-setup` copies over the d-bus uuid. So these two should be the same with this change, is that correct?

No, they aren't:
```
[root@suma-bv-50-liberty9-sshminion ~]# rm -f /etc/machine-id && rm -f /var/lib/dbus/machine-id && mkdir -p /var/lib/dbus && uuidgen > /var/lib/dbus/machine-id && systemd-machine-id-setup && touch /etc/machine-id-already-setup
Initializing machine ID from VM UUID.

[root@suma-bv-50-liberty9-sshminion ~]# cat /etc/machine-id
ff9abf5f8b0b4eb0b21d7b868e78d3ac

[root@suma-bv-50-liberty9-sshminion ~]# cat /var/lib/dbus/machine-id 
2c2781ac-04de-4d2b-bbe4-bd97c8607e5c
```

ktsamis (approved):

salt/default/ids.sls
```diff
@@ -1,16 +1,15 @@
...
-systemd_machine_id:
-  cmd.run:
-    - name: rm -f /etc/machine-id && rm -f /var/lib/dbus/machine-id && mkdir -p /var/lib/dbus && uuidgen | sudo tee /etc/machine-id && touch /etc/machine-id-already-setup
-    - creates: /etc/machine-id-already-setup
-    - onlyif: test -f /usr/bin/systemd-machine-id-setup -o -f /bin/systemd-machine-id-setup
-{% else %}
+install_dbus_uuidgen:
+  pkg.installed:
+    - pkgs:
+      - dbus-tools
```
srbarrios, line 5: `dbus-tools` is a package available in all Linux distros?
If that's the case, we don't need the if statement, as this is declarative language; if it's already installed, will do nothing.

Bischoff: Short answer: no :smile_cat: 
```
ebischoff@kool:~> rpm -qf /usr/bin/dbus-uuidgen
dbus-1-1.12.2-150400.18.8.1.x86_64
```

maximenoel8 (approved):

srbarrios (approved):

PR #1813, opened 2025-04-03
Change download.suse.de by dist.nue.suse.com

srbarrios:
## What does this PR change?

Our hosts in SLC can't resolve download.suse.de due to firewall defaults, instead we should use dist.nue.suse.com


ktsamis (approved):

Bischoff (approved):
LGTM

PR #1818, opened 2025-04-17
Disabling automatic daily system updates

srbarrios:
## What does this PR change?

This PR disable the automatic daily system updates and its reboot, in all images initialized with Combustion.

- `ln -sf /dev/null` reliably disables the service even without `systemctl`.
- Compatible with early boot stages of MicroOS / Leap Micro.
- No dependency on D-Bus or systemd being fully available.

### DIFF
--- a/backend_modules/libvirt/host/combustion
+++ b/backend_modules/libvirt/host/combustion
@@ -187,5 +187,9 @@ zypper --non-interactive install $PACKAGES
 zypper --non-interactive update --no-recommends iptables
 %{ endif }
 
+# Disabling this timer safely by masking it (systemctl might still not available at Combustion phase)
+echo "Disabling automatic daily system updates"
+ln -sf /dev/null /etc/systemd/system/transactional-update.timer
+
 # Leave a marker
 echo "Configured with combustion" > /etc/issue.d/combustion
--- a/backend_modules/libvirt/host/combustion
+++ b/backend_modules/libvirt/host/combustion
@@ -187,7 +187,7 @@ zypper --non-interactive install $PACKAGES
 zypper --non-interactive update --no-recommends iptables
 %{ endif }
 
-# Disabling this timer safely by masking it (systemctl might still not available at Combustion phase)
+# Disabling this timer safely by masking it (systemctl might still be unavailable during the Combustion phase)
 echo "Disabling automatic daily system updates"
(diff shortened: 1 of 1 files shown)


szachovy (approved):

nodeg:
> Automatic updates are controlled by systemd.timer that runs once per day. This applies all updates and informs rebootmgrd that the machine should be rebooted. You may adjust the time when the update runs, see systemd.timer(5) documentation.

For reference: https://documentation.suse.com/smart/systems-management/html/Micro-transactional-updates/index.html#tr-up-automatic-updates-disable

Maybe we should also do something later on e.g. in the test suite with `systemctl`?!

backend_modules/libvirt/host/combustion
```diff
@@ -187,5 +187,9 @@ zypper --non-interactive install $PACKAGES
 zypper --non-interactive update --no-recommends iptables
 %{ endif }
 
+# Disabling this timer safely by masking it (systemctl might still not available at Combustion phase)
+echo "Disabling automatic daily system updates"
+ln -sf /dev/null /etc/systemd/system/transactional-update.timer
+
```
Bischoff, line 193: That's the Kalashikov method :rofl:

Bischoff (approved):
I prefer in the test suite than in sumaform, but ok.

backend_modules/libvirt/host/combustion
srbarrios: Yeah... forced to do that way, because systemctl is not available at Combustion phase.

srbarrios:
> > Automatic updates are controlled by systemd.timer that runs once per day. This applies all updates and informs rebootmgrd that the machine should be rebooted. You may adjust the time when the update runs, see systemd.timer(5) documentation.
> 
> For reference: https://documentation.suse.com/smart/systems-management/html/Micro-transactional-updates/index.html#tr-up-automatic-updates-disable
> 
> Maybe we should also do something later on e.g. in the test suite with `systemctl` (`sudo systemctl --now disable transactional-update.timer`)?!

No we don't need.
What I did here is the same that a `systemctl mask` that will have the same effect that disabling it with `sudo systemctl --now disable transactional-update.timer` just that it works even if systemctl still not ready at Combustion time.

PR #1819, opened 2025-04-17
New env.var. on controller that points to our build sources

srbarrios:
## What does this PR change?

Related to https://github.com/uyuni-project/uyuni/pull/10211/ proposal

### DIFF
--- a/salt/controller/bashrc
+++ b/salt/controller/bashrc
@@ -20,6 +20,7 @@ export VIRTHOST_KVM_PASSWORD="linux" {% else %}# no KVM host defined {% endif %}
 # base goodies
 {% if grains.get('additional_network') | default(false, true) %}export PRIVATENET="{{ grains.get('additional_network') }}" {% else %}# no private network defined {% endif %}
 {% if grains.get('mirror') | default(false, true) %}export MIRROR="yes" {% else %}# no mirror used {% endif %}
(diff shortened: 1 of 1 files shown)


Bischoff (changes requested):
No.

We already use a mirror whenever needed.

In the PR you are referring to, we explicitely want an external resource, even when we use a mirror (look at the title of the step).

And even if you remove the test for the mirror, this is not used for sources, this is only a test of external connectivity.

srbarrios:
> No.
> 
> We already use a mirror whenever needed.
> 
> In the PR you are referring to, we explicitely want an external resource, even when we use a mirror (look at the title of the step).
> 
> And even if you remove the test for the mirror, this is not used for sources, this is only a test of external connectivity.

I don't get your "No." All what you explained after it, is related to your PR, not to this one.
This PR only adds an additional environment variable, so we can know the fqdn of the mirror in case we have one, and otherwise we have the fqdn of our default source in NUE/PRG, which I believe it should be enough.

Bischoff:
> I don't get your "No." All what you explained after it, is related to your PR, not to this one. This PR only adds an additional environment variable, so we can know the fqdn of the mirror in case we have one, and otherwise we have the fqdn of our default source in NUE/PRG, which I believe it should be enough.

After clarification in huddle:
 * OK to check what we will be using at reposync time
 * use `build_sources` as variable name
 * handle all 3 cases: mirror, MLM with download.nue.suse.com, Uyuni with opensuse.org

Bischoff (approved):
thanks

PR #1821, opened 2025-04-22
Fix systemd machine id journal not present

maximenoel8:
## What does this PR change?

Copying the journal log into the new journal log folder is failing for rocky8, leap 15.6, server and deblike minion.
There are two causes:
 - folder /var/log/journal doesn't exist
 - cannot copy a folder in itself

```
Moving logs to /var/log/journal/5fb2f6d9f0af9f07dab8756f6806eca3
mv: cannot move '/var/log/journal/5fb2f6d9f0af9f07dab8756f6806eca3' to a subdirectory of itself, '/var/log/journal/5fb2f6d9f0af9f07dab8756f6806eca3/5fb2f6d9f0af9f07dab8756f6806eca3'

```

Fix this issue with checking the folder journal

### DIFF
--- a/salt/default/ids.sls
+++ b/salt/default/ids.sls
@@ -11,19 +11,36 @@ systemd_machine_id:
   cmd.run:
     - name: >
         bash -c '
-        rm -f /etc/machine-id &&
-        rm -f /var/lib/dbus/machine-id &&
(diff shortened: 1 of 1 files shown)


srbarrios:
Follow up of https://github.com/uyuni-project/sumaform/pull/1820

PR #1822, opened 2025-04-22
use new MLMTools repos for tools_additional_repo when setting up Head

mcalmer:
## What does this PR change?

For Devel:Galaxy:Manager:Head we have new MLM Client Tools repos which must be used when setting up "head".


mcalmer:
here is something missing, but I cannot push.

mcalmer:
> here is something missing, but I cannot push.

Ok, ready to review now.
This is running as run 2009 in test-orion

srbarrios (approved):
LGTM. Remember to adapt the mirror YAML too

mcalmer:
> LGTM. Remember to adapt the mirror YAML too

This was adapted already beginning of March

PR #1825, opened 2025-04-24
Enable by default the tweaks for server performance

srbarrios:
## What does this PR change?

Enable by default the tweaks for server performance.
If we don't want this change in some particular deployment, we can manually disable it.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -928,27 +928,28 @@ module "server" {
 
 ## Large deployments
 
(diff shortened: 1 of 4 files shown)


README_ADVANCED.md
```diff
@@ -928,27 +928,28 @@ module "server" {
...
-hibernate.c3p0.max_size = 50
+hibernate.c3p0.max_size = 100
 
 ### /etc/tomcat/server.xml
 changed `maxThreads` to 256
 
 ### /var/lib/pgsql/data/postgresql.conf
-max_connections = 450
-work_mem = 10MB
+max_connections = 400
```
nodeg, line 943: Why the reduction here?

srbarrios: To align the doc with the code.

nodeg: I see. Thank you for the clarification.

PR #1832, opened 2025-05-09
Update minima version

NamelessOne91:
## What does this PR change?

Updates Salt states so that they all use the latests (at the time of writing) available minima version.


ktsamis (approved):

maximenoel8 (approved):

srbarrios (approved):

srbarrios (changes requested):
Hi @NamelessOne91, I would like that we discuss about this PR.
The title only describe an update of minima version in sumaform, but in fact, now we are injecting a minima mirror as part of our regular server modules. In particular, I dislike and discourage the idea of downloading minima mirror binary inside the container and run it inside the container, instead, I think that the way containers prefers to work is with volumes, and I think we need to handle minima mirror from the server host, and include an additional volume on the container pointing to the content.

salt/server_containerized/test_repositories.yaml
```diff
@@ -0,0 +1,8 @@
+{% set host = grains.get('mirror') | default('download.opensuse.org', true) %}
+storage:
+  type: file
+  path: /srv/www/htdocs/pub/
+
+http:
+  - url: http://{{ host }}/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Updates/rpm
+  - url: http://{{ host }}/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Appstream/rhlike
```
srbarrios, line 8: IMHO, this YAML file should be part of mirror folder, as it aims to be used by a minima mirror. Also, this would benefit of not having the file duplicated in `server` and `server_containerized`.

PR #1834, opened 2025-05-14
[salt-shaker] Specify python version for classic salt

m-czernek:
Some OSes enable Python 3.11. When we first install `python3-salt`, this defaults to `python311-salt`. In such case, if we try to install `python3-salt` later again, this causes conflicts with the already installed python311-salt.

Related to: https://github.com/SUSE/spacewalk/issues/27141


agraul:
I think this is a duplicate of https://github.com/uyuni-project/sumaform/pull/1778

meaksh (approved):
Let's give it a try :+1:

PR #1836, opened 2025-05-15
fix channels option for server_containerized

mcalmer:
## What does this PR change?

With server_containerized the channels option is not working. This PR fix the first things


srbarrios (approved):
LGTM. Thank you for the fix.

PR #1837, opened 2025-05-19
QE: Allowing communication between build_host and server via ssh

Pablogoliva:
## What does this PR change?

This PR modifies the `build_host` and the `server` salt states to allow the communication from the `build_host` to the `server` via ssh.


maximenoel8 (approved):

salt/server_containerized/testsuite.sls
```diff
@@ -99,6 +99,13 @@ suse_staging_key_import:
     - name: "mgradm gpg add -f /tmp/suse_staging.key"
     - onchanges:
       - file: suse_staging_key_copy_host
+
+authorized_keys_controller:
```
srbarrios, line 103:
```suggestion
authorized_keys_buildhost:
```

srbarrios: One interesting thing from Salt states, is that the names must be unique, otherwise it will fail.

Pablogoliva: Solved

salt/build_host/init.sls
```diff
@@ -7,6 +7,24 @@ certificate_authority_certificate:
...
     - makedirs: True
 
+ssh_private_key:
+  file.managed:
+    - name: /root/.ssh/id_ed25519
+    - source: salt://build_host/keys/id_ed25519
+    - makedirs: True
+    - user: root
+    - group: root
+    - mode: 700
```
NamelessOne91, line 17: 400 (owner readonly) is probably better and makes sure that your keys are accepted when used by the SSH client.
Still, if you've tried the SSH connection and it works: all good.

NamelessOne91 (approved):
LGTM

Just a note about permissions set for the keys' files.

salt/build_host/init.sls
Pablogoliva: It works yes. But from a security point of view is a good idea.

Pablogoliva: Done

srbarrios (approved):

PR #1853, opened 2025-06-12
Increase default RAM and vcpu for containerized server

maximenoel8:
## What does this PR change?
Update default value for vcpu and ram for the containerized server.


backend_modules/libvirt/host/main.tf
```diff
@@ -28,9 +28,9 @@ locals {
     },
     contains(local.x86_64_v2_images, var.image) ? { cpu_model = "host-model", xslt = file("${path.module}/cpu_features.xsl") } : {},
     contains(var.roles, "server") ? { memory = 4096, vcpu = 2 } : {},
-    contains(var.roles, "server_containerized") ? { memory = 4096, vcpu = 2 } : {},
+    contains(var.roles, "server_containerized") ? { memory = 20480, vcpu = 4 } : {},
```
srbarrios, line 31: I would prefer to use the minimum requirements as default:
image

srbarrios, line 31:
```suggestion
    contains(var.roles, "server_containerized") ? { memory = 16384, vcpu = 4 } : {},
```

Bischoff (commented):
Agree with Oscar's comments

cbbayburt:
IIUC, we wanted this value to be the recommended default for production servers. But does it make sense to have this much in sumaform where it's only used for development (except testsuite, which has its well-defined custom configuration). So in sumaform default values are the ones you use when you're quickly firing up a server to test/develop something with a few clients and a couple of repos.

I’ve been using sumaform with 8GB servers for many years and versions including 5.1, and never had any issue in the development context. So I think 8G is a sensible default in sumaform for the server.

PR #1857, opened 2025-06-17
set large deployment setup to false by default

NamelessOne91:
## What does this PR change?

See the title: I really think these tweaks should be opt-in.


nodeg:
Fine with me, however we should use them in BV. So not sure if we explicitly set the value there, but we need to verfiy this. Otherwise LGTM.

nodeg (approved):

PR #1862, opened 2025-06-25
Decompress xz images before volumes creation

NamelessOne91:
## What does this PR change?

Adds a null resource using local-exec to decompress .xz images before they are used to create libvirt volumes.

The set of `compressed_images` can be retrieved dinamically, but TBH I think it's more readable to have it hardcoded and keep it up to date when more are added.


NamelessOne91:
> I wonder how much time it takes this decompression? In any case, LGTM, it does not seems we have another choice.

It doesn't take long for openEuler, the image is not that big - I can time it,  but it's probably (way) less than 30 secs.
The whole decompression thing can turn into a problem if we have to deploy lots of VMs based on (big) compressed images, but TBH I don't see that happening anytime soon.

maximenoel8 (approved):
LGTM. Do we know why one of the github actions is failing ?

PR #1870, opened 2025-07-02
Update devel repositories for MLM 5.1 on sles15sp7

maximenoel8:
## What does this PR change?
Update devel repositories for MLM 5.1 on sles15sp7.
Add missing repositories for sles15sp7.


srbarrios (commented):
LGTM. But I prefer if releng approves it.

PR #1871, opened 2025-07-02
Change url for openeuler image to working one

ktsamis:
## What does this PR change?

Downgrade the openeuler image to use the one that works.


srbarrios:
I don't have all the context, but if it helps, you can take a look inside suma-toolbox to understand what we support:

https://gitlab.suse.de/search?group_id=171&nav_source=navbar&project_id=3426&repository_ref=master&search=openeuler&search_code=true

ktsamis:
After discussions this is definitely not needed.

PR #1878, opened 2025-07-17
Allow to define custom attributes for the database container

meaksh:
## What does this PR change?

This PR adds support to define custom "registry/image/tag" for the database containers, which is particularly helpful for developers when testing devel server and db images.


maximenoel8 (approved):
LGTM Thanks !

PR #1880, opened 2025-07-24
Add AlmaLinux 10 images to make Salt Shaker working

vzhestkov:
## What does this PR change?

NOTE: It's not the complete adding AlmaLinux 10 to the testing environment, but the first step to make Salt Shaker working with the bundle on Alma 10


meaksh (commented):
I think we would also need a change at `backend_modules/libvirt/host/user_data.yaml` to configure the instance.

vzhestkov:
@meaksh do you mean repos for client tools? they don't exist yet and we are just preparing the Bundle for it, that's why I mentioned that it's not complete yet in the description. For this case I just need to add the image to make salt shaker job working.

meaksh:
> @meaksh do you mean repos for client tools? they don't exist yet and we are just preparing the Bundle for it, that's why I mentioned that it's not complete yet in the description. For this case I just need to add the image to make salt shaker job working.

You need to have a `venv-salt-minion` installed by cloudinit in the deployed instance, so the highstate from sumaform is able to run on the new VM (even in the context of Salt Shaker).

So, maybe we would need to have the `venv-salt-minion` at `Uyuni:Master:EL10-Client-Tools` (later we move it to `Stable:`) so we can really drive a deployment for Salt Shaker.

srbarrios (approved):

meaksh (approved):
LGTM

PR #1923, opened 2025-09-30
Install mgradm before configuring the additional_disks

maximenoel8:
## What does this PR change?

For MLM 5.1 on sles15sp7, the salt state `server_containerized.additional_disks` is failing because `mgr-storage-server` is not yet installed (done during combustion for slmicro61).
Move the uyuni tools installation outside mgradm installation state and call it before configuring the additional storages.


maximenoel8:
Yes True. I was confused by it too.

PR #1926, opened 2025-10-02
Update generated mgradm.yaml

NamelessOne91:
## What does this PR change?

Update generated mgradm.yaml and mgrpxy.yaml config files to use the new flags of uyuni-tools.


maximenoel8:
Are you sure we need to specify the image, mgradm will not be able to pick it by default?
Be sure to test those changes before merging.

NamelessOne91:
> Are you sure we need to specify the image, mgradm will not be able to pick it by default? Be sure to test those changes before merging.

I had a better look.

The default **--image**  used by the tools now include all the path after the registry host (e.g. devel/galaxy/manager/5.0/containerfile/suse/manager/5.0/x86_64/server) and the default -**-registry-host** now is registry.suse.de

**--registry** should now be just an alias for **--registry-host**

We can therefore just force the container tag to latest and use default values, at least for CI. The actual needed changes are in the main.tf files for the pipelines.
We have to still keep **container_repository** with a specific value set for the podman_login Salt state, otherwise it will fail. I think Michele mentioned this Salt state being redundant because the tools should automatically try to use SCC credentials if they're provided, maybe we can remove it later on.

```
podman_login:
  cmd.run:
    - name: podman login -u {{ grains.get('cc_username') }} -p {{ grains.get('cc_password') }} {{ grains.get("container_repository") }}
{% endif %}
```

maximenoel8 (approved):

Bischoff (approved):

PR #2070, opened 2026-03-12
During first ssh connection, wait for routable ip

maximenoel8:
## Problem
During the provisioning phase of test deployments, OpenTofu/Terraform was intermittently failing with the error:
dial tcp [fe80::...]:22: connect: invalid argument

This occurred because the libvirt_domain resource often reports the IPv6 Link-Local address (fe80::/10) via the QEMU agent before the Global IPv6 or DHCP IPv4 addresses are fully assigned. Since Link-Local addresses are not reachable from Jenkins worker, the connection is failing.

Additionally, the previous logic could potentially pick up IPv4 Link-Local (APIPA - 169.254.0.0/16) addresses, which would lead to connection timeouts.

# Why the loop

To resolve the "empty host" race condition without relying on arbitrary sleep timers, a dynamic waiter has been introduced.

 - Transition to Dynamic Waiting: Replaced the static time_sleep with a local-exec polling loop. This loop queries the Libvirt QEMU agent until a routable (non-link-local) address is detected.
 - Why we still keep the Regex: While the waiter ensures a routable IP exists, the Libvirt metadata still returns a list containing all detected IPs (including fe80::). The regex remains necessary in the host field to explicitly select the routable address from that list and avoid the "Invalid Argument" error.
 - Efficiency: The provisioning phase now starts the moment the network is ready, reducing total deployment time in CI compared to a fixed sleep.
 - Fail-Fast: If a routable IP is not assigned within the timeout period (e.g., DHCP failure), the waiter exits with an error, providing a clear failure point in Jenkins rather than a generic SSH timeout.

# What does this PRs
Updated the connection block in the terraform_data.provisioning resource to use a strict filter for the host attribute.

Logic: The new logic iterates through all addresses reported by the VM's first network interface and excludes any string starting with fe80 (IPv6 Link-Local) or 169.254 (IPv4 Link-Local).

Result: The provisioner now only attempts to connect to routable Global IPv6 or IPv4 addresses.

Safety: Removed the fallback to 127.0.0.1. If no routable address is found, the host now evaluates to null. This prevents the provisioner from "masking" the failure by attempting to SSH into the local runner/bastion host.

```
16:06:02  │ Error: file provisioner error
16:06:02  │ 
16:06:02  │   with module.build_validation_module.module.server[0].module.server.module.host.terraform_data.provisioning[0],
16:06:02  │   on /home/jenkins/workspace/manager-4.3-qe-mi-validation-sles/results/sumaform/backend_modules/libvirt/host/main.tf line 278, in resource "terraform_data" "provisioning":
16:06:02  │  278:   provisioner "file" {
16:06:02  │ 
16:06:02  │ timeout - last error: dial tcp [fe80::a8b2:93ff:fe02:3d1]:22: connect:
16:06:02  │ invalid argument
16:06:02  ╵
script returned exit code 1
```

## Depends on https://github.com/SUSE/susemanager-ci/pull/1934


backend_modules/libvirt/host/main.tf
```diff
@@ -208,23 +207,19 @@ resource "libvirt_domain" "domain" {
     type           = "pty"
     target_port    = "0"
     target_type    = "serial"
-    source_host    = null
```
maximenoel8: Why those lanes were remove ?

Answer:
They are redundant defaults that can occasionally cause validation warnings or clutter in modern OpenTofu/Terraform providers

Bischoff:
> Instead of forcing IPv4, I would like first to understand and have an explanation of why using IPv6 fails that particular environment.

Linux kernel would always prefer IPv6 over IPv4 when there is a choice.

But I would like to understand better the situation too.

Bischoff (changes requested):
Please do not merge this, at least for now.

From what I see after some quick debugging, there is a problem at network level I need to solve.

backend_modules/libvirt/host/rng.xsl
```diff
@@ -0,0 +1,19 @@
...
+            </rng>
+        </xsl:copy>
+    </xsl:template>
+
+    <xsl:template match="node()|@*">
+        <xsl:copy>
+            <xsl:apply-templates select="node()|@*"/>
+        </xsl:copy>
+    </xsl:template>
+</xsl:stylesheet>
```
Bischoff: Missing final linefeed

Bischoff (changes requested):
just a doubt about one line, plus one nitpick

Bischoff (approved):
thanks for the regex changes !

NamelessOne91:
I am not sure if this is strictly related, but I remember that when we updated the terraform-libvirt-provider we had similar issues around IPv6 and link-local addresses being used. Especially for retail.

Technically, the provider we build in OBS should be patched to avoid considering an interface ready if that's the only available address.
See https://build.opensuse.org/projects/systemsmanagement:sumaform/packages/terraform-provider-libvirt/files/ipv6.patch?expand=1

Can you double check that:

1) we are indeed using a provider binary installed from our repo and RPM. Not one pulled from the public registry.
2) if that DEBUG log pops up and there's a match with the IP causing troubles

PR #2094, opened 2026-03-31
Switch to 5.2-released (currently 5.2 beta 1)

Bischoff:
## What does this PR change?

In 5.2 Build Validation, now we start to have usable 5.2 branches, it's time to drift away from Head.

This PR adds support for 5.2-released product (currently 5.2 beta 1).

Piggyback: use `%{ elif product_version ==` in jinja instead of `%{ else } %{ if product_version ==` .


ktsamis (approved):
without checking if the urls are valid, the schemas look good and it seems correct the logic

NamelessOne91 (approved):

Issue #2091, opened 2026-03-29
server_containerized: wait_for_mgr_sync.sh loops indefinitely when no SCC credentials are provided


libkush:
When deploying a containerized Uyuni server without SCC credentials (which is the normal case for Uyuni, SCC is only required for SUSE Manager), the `scc_data_refresh` Salt state runs `wait_for_mgr_sync.sh` which loops indefinitely, blocking tofu apply forever.

## Environment

- **Host OS**: openSUSE Tumbleweed
- **Backend**: libvirt (local)
- **product_version**: uyuni-master
- **OpenTofu version**:  v1.11.5

## `main.tf`:

```terraform
terraform {
 required_version = ">= 1.6.0"
 required_providers {
   libvirt = {
     source = "dmacvicar/libvirt"
     version = "0.8.3"
   }
 }
}

provider "libvirt" {
  uri = "qemu:///system"
}

module "base" {
  source = "./modules/base"
  product_version = "uyuni-master"
  domain = "local"
  ssh_key_path    = "~/.ssh/id_ed25519.pub"
  images = ["opensuse156o", "tumbleweedo"]
}

module "server" {
  source = "./modules/server_containerized"
  base_configuration = module.base.configuration
  name = "server"
  create_sample_bootstrap_script = false

  provider_settings = {
    memory = 4096
    vcpu   = 4
  }
}

module "minion" {
  source = "./modules/minion"
  base_configuration = module.base.configuration

  name = "minion"
  image = "opensuse156o"
  server_configuration = module.server.configuration

  provider_settings = {
    memory = 1024
    vcpu   = 2
  }
}
```

## Steps to reproduce

```bash
cp main.tf.libvirt.example main.tf
# edit main.tf as above — note: no cc_username or cc_password set
cd modules && ln -sfn ../backend_modules/libvirt backend
tofu init
tofu apply
```

**Observed behavior**:

After `mgradm install` completes successfully, the provisioning hangs indefinitely with output like:
```
(remote-exec): Waiting for mgr-sync refresh to finish...
(remote-exec): ...not finished yet...
(remote-exec): Refreshing Channel families
(remote-exec):   Error: <Fault 10104: 'redstone.xmlrpc.XmlRpcFault: No SCC organization credentials found.'>
(remote-exec):              [DONE]
(remote-exec): Refreshing SUSE products                       [FAIL]
(remote-exec): ...not finished yet...
(remote-exec): Refreshing Channel families
[...]

PR #2100, opened 2026-04-01
fix: skip scc_data_refresh if cc_username is not set

libkush:
## What does this PR change?

Fixes the issue #2091 tofu apply hangs indefinitely when deploying a containerized Uyuni server without SCC credentials.

## Problem

The `scc_data_refresh` state runs `wait_for_mgr_sync.sh` unconditionally. This script polls `mgr-sync` in a loop waiting for channel and product synchronization to complete. Without SCC credentials (which is the normal case for Uyuni deployments), the sync always fails with

```
Error: <Fault 10104: 'redstone.xmlrpc.XmlRpcFault: No SCC organization credentials found.'>
```

causing the loop to never exit and `tofu apply` to hang forever until manually killed.

### DIFF
--- a/salt/server_containerized/initial_content.sls
+++ b/salt/server_containerized/initial_content.sls
@@ -29,6 +29,7 @@ scc_data_refresh:
     - args: "{{ server_username }} {{ server_password }}"
     - require:
       - cmd: mgradm_install
+    - unless: test -z "{{ grains.get('cc_username') or '' }}"
 
 {% if grains.get('channels') %}
 add_channels:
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

libkush:
I changed it so that it uses a Jinja `if` block instead of shell interpolation.

nodeg:
Taking a look at the changes, this could be related to https://github.com/uyuni-project/sumaform/pull/1988 from @wombelix.

PR #2116, opened 2026-04-16
Enable java_debugging also in BV

ilmanzo:
## What does this PR change?

Enables java_debugging also in BV 

```
mlm-bv-sandbox-server:~ # ss -ln | grep 800
tcp   LISTEN 0      4096                   0.0.0.0:8002             0.0.0.0:*
tcp   LISTEN 0      4096                   0.0.0.0:8003             0.0.0.0:*
tcp   LISTEN 0      4096                   0.0.0.0:8001             0.0.0.0:*
tcp   LISTEN 0      4096                   0.0.0.0:9800             0.0.0.0:*
```

### DIFF
--- a/modules/build_validation/main.tf
+++ b/modules/build_validation/main.tf
@@ -79,7 +79,7 @@ module "server" {
   database_disk_size   = 150
 
   server_mounted_mirror          = var.platform_location_configuration[var.location].mirror
-  java_debugging                 = false
+  java_debugging                 = true
   auto_accept                    = false
   monitored                      = true
   disable_firewall               = false
@@ -124,7 +124,7 @@ module "server_containerized" {
   container_tag                  = "latest"
   beta_enabled                   = false
   server_mounted_mirror          = var.platform_location_configuration[var.location].mirror
-  java_debugging                 = false
+  java_debugging                 = true
   auto_accept                    = false
   disable_firewall               = false
   allow_postgres_connections     = false


NamelessOne91 (approved):

ktsamis (commented):
I would explicitly turn this to true so it's visible that the variable is there and true instead of defaulting to true because then we have to search for it

ilmanzo:
> I would explicitly turn this to true so it's visible that the variable is there and true instead of defaulting to true because then we have to search for it

ok... explicitly set now

modules/build_validation/main.tf
```diff
@@ -124,7 +124,7 @@ module "server_containerized" {
   container_tag                  = "latest"
   beta_enabled                   = false
   server_mounted_mirror          = var.platform_location_configuration[var.location].mirror
-  java_debugging                 = false
+  java_debugging                 = true
```
szachovy, line 127: How is it reflected? It's adding `debug = True` to yml during `mgradm install` or something else?

ilmanzo: I assume the value is used in some salt configuration, like

https://github.com/uyuni-project/sumaform/blob/master/salt/server/spacewalk-search.sls

szachovy: seems ok: https://github.com/uyuni-project/sumaform/blob/eee5edce22da2de0e7ac8dbaa59d340ffdcca9b7/salt/server_containerized/mgradm.yaml#L59

approving

szachovy (approved):

PR #2118, opened 2026-04-17
QE: Remove useless variables

Pablogoliva:
## What does this PR change?

Mixing to variables in one to reduce problems in the main.tfs. The OCI were spitted to write them properly in the Chart.yaml, but I decided to do it via salt to reduce problems writing of the OCI.

### DIFF
--- a/main.tf.libvirt.rke.example
+++ b/main.tf.libvirt.rke.example
@@ -104,11 +104,8 @@ module "cucumber_testsuite" {
       # Override where to get the containers from
       # container_repository = "registry.opensuse.org/systemsmanagement/uyuni/master/containers/uyuni"
 
-      # Name of the helm chart. The last sting of the OCI url.
-      # helm_chart_name = "server-helm"
-
       # Helm oci
-      # helm_chart_url = "oci://registry.opensuse.org/systemsmanagement/uyuni/master/charts/uyuni"
+      # helm_chart_url = "oci://registry.opensuse.org/systemsmanagement/uyuni/master/charts/uyuni/server-helm"
       
       # Override the default image names for Server and Database containers
       # container_image = "server"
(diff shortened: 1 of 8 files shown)


ktsamis (commented):
You are removing the name variable but keep the url but also use repository which is split from url, same as the new name. Why? What does this change instead of constructing the final url at the end? What difference does it make if the url is given with the name together and then you split it or when they are given separately and then put together? I feel it's better to have a common url and then the names separately just in case we want to edit any of these later in jenkins and overwrite than construct them in the code. Does this fail to apply now and that's why you change it?

Pablogoliva:
> You are removing the name variable but keep the url but also use repository which is split from url, same as the new name. Why? What does this change instead of constructing the final url at the end? What difference does it make if the url is given with the name together and then you split it or when they are given separately and then put together? I feel it's better to have a common url and then the names separately just in case we want to edit any of these later in jenkins and overwrite than construct them in the code. Does this fail to apply now and that's why you change it?

Yes that's what I'm doing having just one URL. I did it that way in the past to save time, but I had this improvement as technical debt.
The url of the oci is: oci://registry.suse.de/devel/galaxy/manager/main/mlm-beta-products-sle15/charts/suse/multi-linux-manager/5.2/server-helm but in the Chart.yaml is written separated in name and repository like in the picture  below. In the beginning I was separating it by hand, but I have decided in this PR to write the whole URL and separate it via SW in sumaform, I didn't do it in the past because there were many things to do and I decided to put it together later. 
image

ktsamis:
> > You are removing the name variable but keep the url but also use repository which is split from url, same as the new name. Why? What does this change instead of constructing the final url at the end? What difference does it make if the url is given with the name together and then you split it or when they are given separately and then put together? I feel it's better to have a common url and then the names separately just in case we want to edit any of these later in jenkins and overwrite than construct them in the code. Does this fail to apply now and that's why you change it?
> 
> Yes that's what I'm doing having just one URL. The url of the oci is: oci://registry.suse.de/devel/galaxy/manager/main/mlm-beta-products-sle15/charts/suse/multi-linux-manager/5.2/server-helm but in the Chart.yaml is written separated in name and repository like in the picture below, separated in name and repository. In the beginning I was separating it by hand by I decided to write the whole URL and separate it via SW in sumaform. 

But this is what I don't understand, why change it now? If we have default value for name and url and they get written separately in the chart.yaml anyway, why change that and first put them together to separate them later? This PR changes a behavior in a way that I don't get, it was working before right? What's the benefit in merging this? It seems to me cosmetic at best. I don't want to block this if there is a reason for it or you simply prefer it this way, I am only trying to understand it

Pablogoliva:
After a discussion, we decided that is not necessary.

PR #2120, opened 2026-04-20
QE: Configure apparmor and get the name of the uyuni pod via script

Pablogoliva:
## What does this PR change?

Configure apparmor in SLES 15 SP7 and switch the variable server_pod for a script, because sometimes the name of the pod changes during runtime and I prefer having a script to reduce flakyness.

### DIFF
--- a/salt/server_kubernetes/get_server_pod_name
+++ b/salt/server_kubernetes/get_server_pod_name
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+source /etc/profile.d/rke2_vars.sh
+kubectl get pods -n uyuni --no-headers | grep uyuni | grep -v setup | awk '{print $1}'
+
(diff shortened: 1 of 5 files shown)


salt/server_kubernetes/get_server_pod_name
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+source /etc/profile.d/rke2_vars.sh
+kubectl get pods -n uyuni --no-headers | grep uyuni | grep -v setup | awk '{print $1}'
```
Bischoff, line 4: Can the pod name change ? Or can we just hardcode it ?

Bischoff (commented):
a question

salt/server_kubernetes/get_server_pod_name
Pablogoliva: Yes, it can change. Sometimes the magic of kubernetes changes the name of the pod for unknown reasons, that's the reason why I decided to get this name via a script instead saving it into a variable.

Pablogoliva: Kubernetes is a black magic that from time to time decides to restart the pod, but is not a bad behavior this is how it works. It has happened to me just once after some weeks of deploys, but let's get the name via a script to avoid flakyness.

Bischoff: Thanks for the explanation

maximenoel8 (approved):

PR #2127, opened 2026-04-23
Augment ulimits and mmaps on the controller

Bischoff:
## What does this PR change?

Augment ulimits and mmaps on the controller: the ulimits by a factor 32, and the mmaps by a factor 20.

Mostly useful for full BV. Tailored for 50 chrome processes.

Tested on 5.1.3 full BV:
```
mlm-bv-51-sles-controller:~ # cat /proc/sys/vm/max_map_count 
524288
mlm-bv-51-sles-controller:~ # ulimit -n 
32768
```

### DIFF
--- a/salt/controller/limits.conf
+++ b/salt/controller/limits.conf
@@ -1,0 +1,6 @@
+# Per-user open files limits
+root           soft   nofile         32768
+root           hard   nofile         32768
+
+admin          soft   nofile         32768
+admin          hard   nofile         32768
--- a/salt/controller/init.sls
+++ b/salt/controller/init.sls
@@ -265,3 +265,15 @@ node_exporter_service:
     - enable: True
     - require:
       - pkg: node_exporter
+
+augment_ulimits:
+  file.managed:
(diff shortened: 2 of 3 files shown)


salt/controller/limits.conf
```diff
@@ -0,0 +1,3 @@
+# Per-user open files limits
+root           soft   nofile         32768
```
srbarrios: We still don't use it...
But I added an "admin" user on the controller a pair of months ago.

Bischoff: OK

Honestly, that would be an old dream becoming true... :trollface:

ktsamis (approved):

salt/controller/limits.conf
Bischoff: Done.

srbarrios (approved):

NamelessOne91 (approved):

srbarrios (approved):

maximenoel8 (approved):

PR #2134, opened 2026-04-28
Use DHCP4 for ipv4 and configure SLAAC for IPV6 in salt

maximenoel8:
### Summary

Reduce VM provisioning time by ~35 seconds by configuring DHCPv4-only networking during cloud-init when SLAAC IPv6 is enabled, eliminating unnecessary DHCPv6 timeout delays.

### Changes

  - backend_modules/libvirt/host/main.tf - Pass IPv6 configuration to network template
  - backend_modules/libvirt/host/network_config.yaml - Conditionally use dhcp4 when SLAAC is active

 ### What's Fixed

The network setup was waiting 35 seconds for DHCPv6 responses during VM boot, even when using SLAAC (Stateless Address Autoconfiguration via Router Advertisement) for IPv6. This delay occurred before the Terraform connection block could establish.

Now when ipv6.accept_ra = true, the cloud-init network configuration uses IPv4-only DHCP (dhcp4) instead of dual-stack DHCP, eliminating the timeout.

### DIFF
--- a/backend_modules/libvirt/host/network_config.yaml
+++ b/backend_modules/libvirt/host/network_config.yaml
@@ -19,6 +19,6 @@ network:
       - type: static
         address: ${dhcp_dns_address}
 %{ else }
-      - type: dhcp
+      - type: dhcp4
 %{ endif }
 %{ endif }
--- a/backend_modules/libvirt/host/main.tf
+++ b/backend_modules/libvirt/host/main.tf
@@ -263,10 +263,6 @@ resource "terraform_data" "provisioning" {
 
   count = var.provision ? var.quantity : 0
 
-  provisioner "local-exec" {
-    command = "sleep 20"
-  }
-
   connection {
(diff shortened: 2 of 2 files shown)


backend_modules/libvirt/host/network_config.yaml
```diff
@@ -19,6 +19,10 @@ network:
       - type: static
         address: ${dhcp_dns_address}
 %{ else }
+%{ if ipv6.accept_ra ~}
```
Bischoff: Simplify - there is no case where we need `dhcp`

Just change to `dhcp4`, no questions asked.

Bischoff (changes requested):

Bischoff (approved):
As a final comment: we went down from 35 seconds to 5 seconds. We only won 30 seconds.

The remaining 5 seconds come from the time to release any previous DHCPv4 lease. I suggest we don't try to skip them, even though that would be technically possible.

PR #2143, opened 2026-05-12
Wire in BV module public ssh key for arm and feilong

maximenoel8:
## What does this PR change?

Some wirings are missing in Build Validation module.
Add the ssh key path to feilong module and arm module.


Bischoff (approved):
Title and commit title: feiling => feilong

PR #2145, opened 2026-05-13
Fix dhcp4 for -ign base OS

maximenoel8:
## What does this PR change?

dhcp4 is not applied to ignition base OS because only ignition is executed. Update ignition base to set dhcp4.


Bischoff (approved):
wow

PR #2155, opened 2026-05-19
Add Kubernetes storage backend selection

AkashKumar7902:
## What does this PR change?

Adds configurable Kubernetes storage handling for `server_kubernetes`, `proxy_kubernetes`, and the Kubernetes testsuite path.

The default behavior remains unchanged: `local-path` still installs Rancher's local-path provisioner, marks the configured StorageClass as default, and creates the static `var-spacewalk` hostPath PV by default.

Non-`local-path` backends (`external`, `nfs`, `longhorn`, `openebs`, `rook-ceph`, `cloud`) skip local-path provisioner setup and the static hostPath PV by default, so deployments can use an existing StorageClass. The server Helm values also receive the configured StorageClass for its PVCs when one is set.

Backend-specific provisioner installers are intentionally kept out of this shared base change and are proposed in separate PRs.

## Validation

- Rendered the affected Salt/Jinja templates for `local-path`, explicit external class, and default StorageClass modes.
- Parsed the rendered SLS/YAML successfully.
- Ran `terraform validate`.

`terraform fmt -check` still reports formatting differences in the existing touched HCL files; no broad formatting pass was applied to keep the functional diff focused.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -935,6 +935,42 @@ module "server" {
 }
 ```
 
+## Kubernetes storage
+
+The `server_kubernetes` and `proxy_kubernetes` modules install Rancher's `local-path-provisioner` by default and use the `local-path` StorageClass for Kubernetes persistent volume claims.
+This preserves the single-node test-lab behavior.
+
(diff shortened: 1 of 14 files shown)


AkashKumar7902:
Closing this in favor of #2198, which implements the Kubernetes StorageClass configuration with local-path preserved as default and external StorageClass support through the correct Helm values path.

PR #2163, opened 2026-05-27
Install kubectl and helm on the controller

AkashKumar7902:
## What does this PR change?

Adds two new variables on the `cucumber_testsuite` module:

- `install_kubectl_helm` (bool, default `false`) — when true, installs `kubectl` (`kubernetes-client` package) and `helm` (via the official `get-helm-4` script) on the controller VM.
- `kubeconfig_path` (string, default `null`) — path to a kubeconfig file on the machine running Terraform; its contents are base64-encoded into a grain and written to `/root/.kube/config` on the controller with `0600` permissions.

This is the first step toward driving cluster operations (helm install, kubectl exec, etc.) from the controller rather than from the `server_kubernetes` VM, which is a prerequisite for supporting an existing (user-provided) Kubernetes cluster — per discussion with @cbosdo and @Pablogoliva.

Default behavior is unchanged: both variables default to false/null, so existing deployments are unaffected.

## Usage

```hcl
module "cucumber_testsuite" {
  source = "./modules/cucumber_testsuite"
  # ...
  install_kubectl_helm = true
  kubeconfig_path      = "/home/user/.kube/config"
}
```

## Test plan

- [ ] `terraform apply` with `install_kubectl_helm = true` and `kubeconfig_path` pointing at a valid kubeconfig: verify `kubectl` and `helm` are installed on the controller and `/root/.kube/config` exists with `0600` perms.
- [ ] `kubectl get nodes` from the controller succeeds against the configured cluster.
- [ ] `terraform apply` without the new variables: verify the controller behaves exactly as before (no `kubectl`, no `/root/.kube/` directory).

### DIFF
--- a/modules/controller/main.tf
+++ b/modules/controller/main.tf
@@ -80,6 +80,8 @@ module "controller" {
     catch_timeout_message     = var.catch_timeout_message
     beta_enabled              = var.beta_enabled
     web_server_hostname       = var.web_server_hostname
+    install_kubectl_helm      = var.install_kubectl_helm
+    kubeconfig_content        = var.kubeconfig_path != null ? base64encode(file(var.kubeconfig_path)) : ""
 
     sle12sp5_paygo_minion    = length(var.sle12sp5_paygo_minion_configuration["hostnames"]) > 0 ? var.sle12sp5_paygo_minion_configuration["hostnames"][0] : null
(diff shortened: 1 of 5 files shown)


maximenoel8 (approved):
LGTM, thank you !

cbosdo (approved):

PR #2167, opened 2026-06-02
QE: Parametrize the use of the containers

Pablogoliva:
## What does this PR change?

Parametrize the use of the coco attestation, tftp and saline containers. Also cleaning up some useless variables for the server_kubernetes module.


modules/cucumber_testsuite/main.tf
```diff
@@ -208,19 +208,6 @@ module "server_kubernetes" {
   db_container_repository        = lookup(local.db_container_repositories, "server_kubernetes", "")
   db_container_image             = lookup(local.db_container_images, "server_kubernetes", "")
   db_container_tag               = lookup(local.db_container_tags, "server_kubernetes", "")
```
Pablogoliva, line 210: These were useless variables for this kubernetes module, here I'm just doing some cleanup.

salt/server_kubernetes/values_server.yaml
```diff
@@ -43,11 +43,21 @@ certManagerNamespace: {{ cert_manager_namespace }}
...
   tftp:
     enable: true
+{% endif %}
 {% if app_armor_name %}
   server:
     apparmorProfile: {{ app_armor_name }}
 {% endif %}
+{% if deploy_coco_attestation %}
+  coco:
+      replicas: 2
```
ktsamis: indentation here needs a fix, use two less

```suggestion
{% if deploy_coco_attestation %}
  coco:
    replicas: 2
{% endif %}
```

ktsamis (changes requested):
just a small nitpick

modules/cucumber_testsuite/main.tf
ktsamis: Why were these useless? They will now default to the the defaults which most are true

Pablogoliva: Because were variables inherited from the server_containerized modules that were not even used, and there's no way to used them on RKE2.

Pablogoliva: Clarified.

salt/server_kubernetes/values_server.yaml
Pablogoliva: Done

ktsamis (approved):

PR #2169, opened 2026-06-03
Remove Rocky8  AMI entry

NamelessOne91:
## What does this PR change?

Rocky 8 AMI seems to be no longer available in the marketplace


nodeg (approved):
LGTM. Nice catch!

ktsamis (approved):

PR #2173, opened 2026-06-04
QE: Restart server in large_deployments

Pablogoliva:
## What does this PR change?

Restart server in large deployments to avoid this issue: `Error while inserting data to the table: server closed the connection unexpectedly` in /var/log/salt/master.
As Vladimir reported [here](https://suse.slack.com/archives/C02D134507M/p1780564216180799).

### DIFF
--- a/salt/server_containerized/large_deployment.sls
+++ b/salt/server_containerized/large_deployment.sls
@@ -28,13 +28,6 @@ large_deployment_tune_tomcat_maxthreads:
     - onchanges:
       - cmd: large_deployment_tune_tomcat_stylesheet_copy
 
-large_deployment_tomcat_restart:
-  cmd.run:
-    - name: mgrctl exec systemctl restart tomcat
-    - watch:
-      - manage_config: large_deployment_edit_rhn
-      - cmd: large_deployment_tune_tomcat_maxthreads
-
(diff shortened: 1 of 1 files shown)


maximenoel8 (approved):

salt/server_containerized/large_deployment.sls
```diff
@@ -91,4 +91,14 @@ large_deployment_postgresql_restart:
       - cmd: large_deployment_increase_database_max_connections
       - cmd: large_deployment_increase_database_work_memory
 {% endif %}
+
+restart_whole_server:
+  cmd.run:
+    - name: mgradm restart
+    - requiere:
```
NamelessOne91: typo? requiEre --> require

Also, I am not sure this will work if only 1 of `large_deployment_postgresql_restart_db_container` and `large_deployment_postgresql_restart` is expected to run.

NamelessOne91 (changes requested):

salt/server_containerized/large_deployment.sls
Pablogoliva: Good catch, it was working and didn't fail. But it didn't work as intended either.

Pablogoliva: Not relevant, anymore. Vladimir said that we have to delete the rest of the services.

ktsamis (approved):

NamelessOne91 (approved):

PR #2174, opened 2026-06-05
Cucumber testsuite rocky9

NamelessOne91:
## What does this PR change?

- Adds SSH option for Rocky 9 and 10
- Sets Rocky 9 as default image for the RHlike minion

### DIFF
--- a/backend_modules/aws/host/user_data.yaml
+++ b/backend_modules/aws/host/user_data.yaml
@@ -101,7 +101,7 @@ yum_repos:
 packages: ["venv-salt-minion", "dbus-tools"]
 runcmd:
   # WORKAROUND: rhel9 change ssh security for ssh key
-  - echo "PermitRootLogin yes" >> /etc/ssh/sshd_config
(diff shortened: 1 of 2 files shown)


ktsamis (approved):

backend_modules/aws/host/user_data.yaml
```diff
@@ -132,6 +132,9 @@ runcmd:
   - yum clean all
   - yum repolist
   - dnf -y install venv-salt-minion
+  # WORKAROUND: cloud-init in Rocky9 does not take care of the following
+  - echo "PermitRootLogin yes" >> /etc/ssh/sshd_config
```
alichiel: I would prefer to use /etc/ssh/sshd_config.d/PermitRootLogin.conf (or differen .conf file) since (from sshd_config):
```

# The strategy used for options in the default sshd_config shipped with
# OpenSSH is to specify options with their default value where
# possible, but leave them commented.  Uncommented options override the
# default value.

Include /etc/ssh/sshd_config.d/*.conf
```

alichiel (approved):

PR #2196, opened 2026-07-06
Fix ip resolution on local deployment to handle nsswitch

maximenoel8:
## What does this PR change?

Fix dns resolution to also accept local deployment.

Tested by @CDellaGiusta for local deployment
Tested by @maximenoel8 for testsuite/dhcp deployment


CDellaGiusta (approved):
positively tested building a local 5.1 VM, in an environment without avahi and with the IPs hardcoded in /etc/hosts

NamelessOne91 (approved):

PR #2198, opened 2026-07-07
Add Kubernetes StorageClass configuration

AkashKumar7902:
## What does this PR change?

Adds opt-in Kubernetes StorageClass configuration for RKE2 server and proxy deployments while preserving the existing `local-path` default behavior.

- Adds `kubernetes_storage_backend` and `kubernetes_storage_class` to `server_kubernetes`, `proxy_kubernetes`, and `cucumber_testsuite`, including per-host `host_settings` overrides.
- Keeps `local-path` as the default backend. In this mode Sumaform still installs the local-path provisioner, marks it as default, and creates the static hostPath PVs for `var-spacewalk` and `var-pgsql18`.
- Skips local-path provisioner setup and static hostPath PV creation by default for non-`local-path` backends, so external StorageClasses can provision the Helm PVCs.
- Passes StorageClass values through the chart-supported paths: `server-helm.volumes.storageClass` and `proxy-helm.volumes.squid.storageClass`.
- Makes the local-path provisioner path, class name, default-class flag, reclaim policy, and static server PV host paths configurable.
- Documents default local-path behavior and external StorageClass usage in `README_ADVANCED.md`.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -981,46 +981,44 @@ module "server" {
 
 ## Kubernetes storage
 
-The `server_kubernetes`, `proxy_kubernetes`, and `cucumber_testsuite` modules use Rancher's `local-path` StorageClass by default.
+The `server_kubernetes` and `cucumber_testsuite` modules use Rancher's `local-path` StorageClass by default.
 This preserves the existing single-node test-lab behavior: sumaform installs the local-path provisioner, marks the StorageClass as default, and creates static hostPath PersistentVolumes for the server database and `/var/spacewalk` volumes.
 
-To use an existing Kubernetes StorageClass from another storage provider, set the storage backend to a non-`local-path` value and provide the StorageClass name:
+To use an existing Kubernetes StorageClass from another storage provider, set the storage backend to `external` and provide the StorageClass name:
 
 ```hcl
-module "server" {
+module "server_kubernetes" {
   source             = "./modules/server_kubernetes"
   base_configuration = module.base.configuration
 
   name                       = "server"
   kubernetes_storage_backend = "external"
-  kubernetes_storage_class   = "longhorn"
+  kubernetes_storage_class   = "my-storage-class"
 }
 ```
 
-With any non-`local-path` backend, sumaform does not install the local-path provisioner and does not create the static local hostPath PersistentVolumes by default.
-The configured StorageClass is passed to the Uyuni Helm chart PVCs under `server-helm.volumes.storageClass` and `proxy-helm.volumes.squid.storageClass`, so the selected storage provider must already exist in the cluster.
+With the `external` backend, sumaform does not install the local-path provisioner and does not create the static local hostPath PersistentVolumes by default.
+The `external` backend is externally managed storage: sumaform does not install any storage provider for it.
+The configured StorageClass is passed to the Uyuni Helm chart PVCs under `server-helm.volumes.storageClass`, so the selected storage provider must already exist in the cluster.
 Set `kubernetes_storage_class = null` to rely on the cluster's default StorageClass.
-
-The backend value can be one of `local-path`, `external`, `nfs`, `longhorn`, `openebs`, `rook-ceph`, or `cloud`.
-The non-`local-path` values are labels for externally managed storage; they do not install the corresponding storage provider.
 
 The local-path backend can also be tuned without changing the default behavior:
 
 ```hcl
-module "server" {
+module "server_kubernetes" {
   source             = "./modules/server_kubernetes"
   base_configuration = module.base.configuration
 
-  name                                 = "server"
-  kubernetes_storage_backend           = "local-path"
-  kubernetes_storage_class             = "local-path"
-  local_path_provisioner_path          = "/opt/local-path-provisioner"
+  name                                  = "server"
+  kubernetes_storage_backend            = "local-path"
+  kubernetes_storage_class              = "local-path"
+  local_path_provisioner_path           = "/opt/local-path-provisioner"
   local_path_provisioner_default_class  = true
   local_path_provisioner_reclaim_policy = "Delete"
 }
 ```
 
-The same settings can be set globally on the `cucumber_testsuite` module or per host in `host_settings` for `server_kubernetes` and `proxy_kubernetes`.
+The same settings can be set globally on the `cucumber_testsuite` module or per host in `host_settings` for `server_kubernetes`.
 
 ## Large deployments
 
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -979,6 +979,49 @@ module "server" {
 }
 ```
 
+## Kubernetes storage
+
+The `server_kubernetes`, `proxy_kubernetes`, and `cucumber_testsuite` modules use Rancher's `local-path` StorageClass by default.
+This preserves the existing single-node test-lab behavior: sumaform installs the local-path provisioner, marks the StorageClass as default, and creates static hostPath PersistentVolumes for the server database and `/var/spacewalk` volumes.
+
+To use an existing Kubernetes StorageClass from another storage provider, set the storage backend to a non-`local-path` value and provide the StorageClass name:
+
+```hcl
+module "server" {
+  source             = "./modules/server_kubernetes"
+  base_configuration = module.base.configuration
+
+  name                       = "server"
+  kubernetes_storage_backend = "external"
+  kubernetes_storage_class   = "longhorn"
+}
+```
+
+With any non-`local-path` backend, sumaform does not install the local-path provisioner and does not create the static local hostPath PersistentVolumes by default.
+The configured StorageClass is passed to the Uyuni Helm chart PVCs under `server-helm.volumes.storageClass` and `proxy-helm.volumes.squid.storageClass`, so the selected storage provider must already exist in the cluster.
+Set `kubernetes_storage_class = null` to rely on the cluster's default StorageClass.
+
+The backend value can be one of `local-path`, `external`, `nfs`, `longhorn`, `openebs`, `rook-ceph`, or `cloud`.
+The non-`local-path` values are labels for externally managed storage; they do not install the corresponding storage provider.
+
+The local-path backend can also be tuned without changing the default behavior:
+
+```hcl
+module "server" {
+  source             = "./modules/server_kubernetes"
+  base_configuration = module.base.configuration
+
+  name                                 = "server"
+  kubernetes_storage_backend           = "local-path"
+  kubernetes_storage_class             = "local-path"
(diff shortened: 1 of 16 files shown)


modules/server_kubernetes/variables.tf
```diff
@@ -104,10 +104,61 @@ variable "install_traefik" {
...
+  description = "true to install local-path-provisioner when kubernetes_storage_backend is local-path"
   default = true
 }
 
+variable "kubernetes_storage_backend" {
+  description = "Storage backend for Kubernetes PVCs. local-path installs Rancher's local-path provisioner; other values expect an existing StorageClass."
+  default     = "local-path"
+
+  validation {
+    condition     = contains(["local-path", "external", "nfs", "longhorn", "openebs", "rook-ceph", "cloud"], var.kubernetes_storage_backend)
```
cbosdo:
```suggestion
    condition     = contains(["local-path", "external"], var.kubernetes_storage_backend)
```

Restrict to what is currently supported, not what could be added in the future

README_ADVANCED.md
```diff
@@ -979,6 +979,49 @@ module "server" {
 }
 ```
 
+## Kubernetes storage
+
+The `server_kubernetes`, `proxy_kubernetes`, and `cucumber_testsuite` modules use Rancher's `local-path` StorageClass by default.
+This preserves the existing single-node test-lab behavior: sumaform installs the local-path provisioner, marks the StorageClass as default, and creates static hostPath PersistentVolumes for the server database and `/var/spacewalk` volumes.
+
+To use an existing Kubernetes StorageClass from another storage provider, set the storage backend to a non-`local-path` value and provide the StorageClass name:
```
cbosdo: Isn't it enough to mark the backend as `external`? From what I can tell, this means the backend is not installed by sumaform, right?

README_ADVANCED.md
```diff
@@ -979,6 +979,49 @@ module "server" {
...
+  source             = "./modules/server_kubernetes"
+  base_configuration = module.base.configuration
+
+  name                       = "server"
+  kubernetes_storage_backend = "external"
+  kubernetes_storage_class   = "longhorn"
+}
+```
+
+With any non-`local-path` backend, sumaform does not install the local-path provisioner and does not create the static local hostPath PersistentVolumes by default.
```
cbosdo: Same here, may be `external` is equivalent to non-local-path.

README_ADVANCED.md
```diff
@@ -979,6 +979,49 @@ module "server" {
...
+The configured StorageClass is passed to the Uyuni Helm chart PVCs under `server-helm.volumes.storageClass` and `proxy-helm.volumes.squid.storageClass`, so the selected storage provider must already exist in the cluster.
+Set `kubernetes_storage_class = null` to rely on the cluster's default StorageClass.
+
+The backend value can be one of `local-path`, `external`, `nfs`, `longhorn`, `openebs`, `rook-ceph`, or `cloud`.
```
cbosdo: What is the interest of handling of those values? It seems that all we want is `local-path` as we install it or `external` that is something assumed to be deployed in an other way.

modules/cucumber_testsuite/main.tf
```diff
@@ -361,6 +388,11 @@ module "proxy_kubernetes" {
   install_cert_manager            = var.install_cert_manager
   install_traefik                 = var.install_traefik
   install_local_path_provisioner  = var.install_local_path_provisioner
+  kubernetes_storage_backend            = lookup(local.kubernetes_storage_backend, "proxy_kubernetes", var.kubernetes_storage_backend)
```
Pablogoliva: What's the matter of having these variables in the proxy kubernetes?

modules/proxy_kubernetes/main.tf
```diff
@@ -67,6 +70,11 @@ module "proxy_kubernetes" {
     install_cert_manager            = var.install_cert_manager
     install_traefik                 = var.install_traefik
     install_local_path_provisioner  = var.install_local_path_provisioner
+    kubernetes_storage_backend            = var.kubernetes_storage_backend
```
Pablogoliva: The storage backend is something fully related with server, not with the proxy. Why are we leaving it here?

salt/kubernetes_common/local-path-storage.yaml
```diff
@@ -1,3 +1,7 @@
+{% set storage_class = grains.get('kubernetes_storage_class') or 'local-path' %}
+{% set local_path = grains.get('local_path_provisioner_path') or '/opt/local-path-provisioner' %}
+{% set reclaim_policy = grains.get('local_path_provisioner_reclaim_policy') or 'Delete' %}
```
Pablogoliva: Use better default values instead of ors there. Just to leave as it is in the rest of sumaform.

README_ADVANCED.md
```diff
@@ -979,6 +979,49 @@ module "server" {
...
 
+## Kubernetes storage
+
+The `server_kubernetes`, `proxy_kubernetes`, and `cucumber_testsuite` modules use Rancher's `local-path` StorageClass by default.
+This preserves the existing single-node test-lab behavior: sumaform installs the local-path provisioner, marks the StorageClass as default, and creates static hostPath PersistentVolumes for the server database and `/var/spacewalk` volumes.
+
+To use an existing Kubernetes StorageClass from another storage provider, set the storage backend to a non-`local-path` value and provide the StorageClass name:
+
+```hcl
+module "server" {
```
Pablogoliva: server_kubernetes

modules/server_kubernetes/variables.tf
```diff
@@ -104,10 +104,61 @@ variable "install_traefik" {
 }
 
 variable "install_local_path_provisioner" {
-  description = "true to install local-path-provisioner"
+  description = "true to install local-path-provisioner when kubernetes_storage_backend is local-path"
   default = true
 }
 
+variable "kubernetes_storage_backend" {
```
Pablogoliva, line 151: Good the variables have default values.

modules/server_kubernetes/variables.tf
AkashKumar7902: Done, restricted to just `local-path` and `external`.

README_ADVANCED.md
AkashKumar7902: Right, that is exactly what it means. I reworded the section to say `external` explicitly instead of talking about "non-`local-path` values".

README_ADVANCED.md
AkashKumar7902: Reworded to `external` here too.

README_ADVANCED.md
AkashKumar7902: Agreed, they did not really add anything. Only `local-path` and `external` are accepted now, and a concrete provider like longhorn only shows up as the StorageClass name in `kubernetes_storage_class`.

modules/cucumber_testsuite/main.tf
AkashKumar7902: Removed, the proxy does not take any storage settings anymore.

modules/proxy_kubernetes/main.tf
AkashKumar7902: You are right, this is really a server concern. I dropped all the storage variables from proxy_kubernetes, so the proxy keeps installing local-path on its own cluster unconditionally like before (its squid cache PVC binds via the default StorageClass there). If configurable proxy storage is ever needed, the chart already exposes `proxy-helm.volumes.squid.storageClass`, but that would be a separate PR.

salt/kubernetes_common/local-path-storage.yaml
AkashKumar7902: Changed them to the `| default(value, true)` filter that is used elsewhere in the salt tree. I could not use the plain `grains.get('key', default)` form here: terraform writes null variables into the grains file as `key: null`, so the grain exists with a None value and the two-arg default never applies. That is why I originally had the `or`s.

README_ADVANCED.md
AkashKumar7902: Fixed, in the second example below as well.

AkashKumar7902:
Rebased on master to resolve the conflict with the container_registry rename; no functional changes from the rebase.

salt/server_kubernetes/set-persistent-volumes.sls
```diff
@@ -1,29 +1,51 @@
...
+{% set create_static_var_pgsql_pv = storage_backend == 'local-path' %}
+{% endif %}
 
 create_uyuni_namespace:
-  cmd.run: 
+  cmd.run:
   - name: kubectl create namespace uyuni
   - env:
     - KUBECONFIG: {{ kubeconfig }}
+  - unless: KUBECONFIG={{ kubeconfig }} kubectl get namespace uyuni
```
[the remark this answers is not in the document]
AkashKumar7902: This is a false positive — a block sequence at the key's indentation is valid YAML, and Salt parses it to the same cmd.run argument list (this state has also run fine in real deployments). Reindented in 099617fd anyway for consistency with the rest of the salt tree, and simplified the unless while touching it: Salt forwards the state's env to unless checks, so the KUBECONFIG= prefix was redundant.

salt/kubernetes_common/set_up_local-path-provisioner.sls
```diff
@@ -20,30 +23,37 @@ copy_local-path-storage_installation_file:
...
       - KUBECONFIG: {{ kubeconfig }}
+    - require:
+      - file: copy_local-path-storage_installation_file
 
+{% if default_class %}
 set_local-path-storage-file-as-default:
   cmd.run:
     - name: |
-        kubectl patch storageclass local-path -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
+        kubectl patch storageclass {{ storage_class }} -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
```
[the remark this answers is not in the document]
AkashKumar7902: Done in 099617fd.

salt/kubernetes_common/set_up_local-path-provisioner.sls
```diff
@@ -20,30 +23,37 @@ copy_local-path-storage_installation_file:
...
 
 create_local-path-provisioner_directory:
   file.directory:
-    - name: /opt/local-path-provisioner
+    - name: {{ local_path }}
 
 set_local-path-provisioner_selinux_tags:
   cmd.run:
-    - name: restorecon -R -v /opt/local-path-provisioner
+    - name: restorecon -R -v {{ local_path }}
```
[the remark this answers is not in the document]
AkashKumar7902: Done in 099617fd, quoted the file.directory path there as well.

Pablogoliva (approved):

PR #2199, opened 2026-07-08
Add external Kubernetes cluster server install

AkashKumar7902:
## What does this PR change?

Adds an external Kubernetes cluster mode for the Kubernetes testsuite server deployment.

In the default `managed` mode, behavior is unchanged: `kubernetes = true` creates the Sumaform-managed `server_kubernetes` and `proxy_kubernetes` RKE2 VMs.

In the new `external` mode, `kubernetes = true` remains the top-level switch, but Sumaform skips the `server_kubernetes` and `proxy_kubernetes` VMs. The controller installs kubectl and Helm, writes the provided kubeconfig to `/root/.kube/config`, and runs the Uyuni server Helm install against the existing cluster.

The target cluster must already provide the required ingress/load-balancer and storage setup. If `install_cert_manager` is enabled, Sumaform installs cert-manager and trust-manager before installing Uyuni. This mode currently covers the Uyuni server deployment only; Kubernetes proxy deployments remain part of the managed RKE2 workflow and are skipped in external mode.


modules/cucumber_testsuite/main.tf
```diff
@@ -660,8 +677,8 @@ module "controller" {
 output "configuration" {
   value = {
     base = module.base.configuration
```
Pablogoliva, line 733: DOn't delete it either.

modules/cucumber_testsuite/main.tf
```diff
@@ -198,7 +203,7 @@ module "server_containerized" {
 module "server_kubernetes" {
   source                         = "../server_kubernetes"
 
-  count = var.kubernetes ? 1 : 0
+  count = local.kubernetes_managed_cluster ? 1 : 0
```
Pablogoliva: Leave kubernetes variable as it is, otherwise it completely break the RKE2 pipelines we have.

salt/controller/install_kubernetes_server.sls
```diff
@@ -0,0 +1,170 @@
+{% set kubeconfig = "/root/.kube/config" %}
```
Pablogoliva: Why don't we reuse the one that we already have?

Pablogoliva (changes requested):

modules/cucumber_testsuite/main.tf
Pablogoliva: This kubernetes variables is very key for out automation. You can't delete it.

modules/cucumber_testsuite/main.tf
```diff
@@ -198,7 +203,7 @@ module "server_containerized" {
 module "server_kubernetes" {
   source                         = "../server_kubernetes"
 
-  count = var.kubernetes ? 1 : 0
```
Pablogoliva, line 253: Can't be deleted

modules/cucumber_testsuite/main.tf
```diff
@@ -319,7 +324,7 @@ module "proxy_kubernetes" {
   depends_on = [module.server_kubernetes]
   source = "../proxy_kubernetes"
 
-  count = var.kubernetes ? 1 : 0
```
Pablogoliva, line 392: Nest things like this: var.kubernetes ? 0 : (var.container_proxy ? 1 : 0)

modules/cucumber_testsuite/main.tf
AkashKumar7902: Restored in 3b0d8fb9 — the output keeps the original `var.kubernetes ? ... : ...` expressions, with the external mode handled as a nested condition inside the kubernetes branch.

modules/cucumber_testsuite/main.tf
AkashKumar7902: Done in 3b0d8fb9. `var.kubernetes` stays as the outer condition and the new mode check is nested inside it: `count = var.kubernetes ? (var.kubernetes_cluster_mode == "external" ? 0 : 1) : 0`. Since `kubernetes_cluster_mode` defaults to `managed`, existing RKE2 pipelines with `kubernetes = true` behave exactly as before.

salt/controller/install_kubernetes_server.sls
AkashKumar7902: Fair point, there was a lot of duplication here. Reworked in 3b0d8fb9: this file now only keeps the external-only bits (kubeconfig check, cert-manager/trust-manager, uyuni namespace) and includes `server_kubernetes.install_kubernetes_server` for the actual chart/values/helm install. The shared state got an `is_external_cluster` branch, and I compared the rendered output for the managed/RKE2 grain combinations to make sure it is unchanged.

modules/cucumber_testsuite/main.tf
AkashKumar7902: The variable itself was never removed, but agreed the counts should keep referencing it directly instead of going through a local — fixed in 3b0d8fb9.

modules/cucumber_testsuite/main.tf
AkashKumar7902: Restored in 3b0d8fb9.

modules/cucumber_testsuite/main.tf
AkashKumar7902: Done, nested it that way in 3b0d8fb9.

salt/controller/init.sls
```diff
@@ -1,6 +1,9 @@
 include:
   - repos
   - controller.apache_https
+  {% if grains.get('install_kubernetes_server_on_external_cluster') == true %}
```
Pablogoliva, line 4: It should be changed. It must point to the server_install:kubernetes sls on server_kubernetes instead of here.

salt/controller/init.sls
```diff
@@ -1,6 +1,9 @@
...
+  - controller.install_kubernetes_server
```
Pablogoliva: server_kubernetes.install_kubernetes_server

salt/controller/install_kubernetes_server.sls
```diff
@@ -0,0 +1,66 @@
...
+{% set cert_manager_namespace = "cert-manager" %}
```
Pablogoliva: Reuse the file on server_kubernetes pls, instead of creating a new one here.

salt/controller/install_kubernetes_server.sls
```diff
@@ -0,0 +1,66 @@
...
+      - KUBECONFIG: {{ kubeconfig }}
+    - require:
+      - cmd: install_helm_on_controller
+      - file: external_kubernetes_kubeconfig
+{% endif %}
+
+create_external_kubernetes_uyuni_namespace:
+  cmd.run:
+    - name: kubectl create namespace uyuni
+    - unless: KUBECONFIG={{ kubeconfig }} kubectl get namespace uyuni
```
Pablogoliva: Not neccesary

salt/controller/install_kubernetes_server.sls
AkashKumar7902: done

salt/server_kubernetes/install_kubernetes_server.sls
```diff
@@ -44,6 +50,74 @@ key_exchange_kubernetes_server:
...
+      - file: external_kubernetes_kubeconfig
+
+install_trust_manager_on_external_kubernetes:
+  cmd.run:
+    - name: |
+        helm upgrade --install \
+          trust-manager oci://quay.io/jetstack/charts/trust-manager \
+          --namespace {{ cert_manager_namespace }} \
+          --wait
+    - unless: KUBECONFIG={{ kubeconfig }} helm status trust-manager --namespace {{ cert_manager_namespace }}
```
Pablogoliva: The environment is already set below, there's no need to set the env variable here.

salt/server_kubernetes/install_kubernetes_server.sls
```diff
@@ -44,6 +50,74 @@ key_exchange_kubernetes_server:
     - source: salt://proxy_kubernetes/proxy_keys/id_ed25519_proxy.pub
     - makedirs: True
 
+{% else %}
+
+install_external_kubernetes_dependencies:
+  pkg.installed:
+    - pkgs:
+      - python3-PyYAML
```
Pablogoliva: Above we already have a place to install packages above in the line 24, please use it to parameterize it depending of the base OS. Also if you are using an external cluster why do you need to install something in the controller?

salt/server_kubernetes/install_kubernetes_server.sls
```diff
@@ -44,6 +50,74 @@ key_exchange_kubernetes_server:
...
+  file.exists:
+    - name: {{ kubeconfig }}
+    {% if grains.get('kubeconfig_content') %}
+    - require:
+      # write_kubeconfig is defined in controller/init.sls, which includes this state
+      - cmd: write_kubeconfig
+    {% endif %}
+
+{% if grains.get('install_cert_manager') == true %}
+install_cert_manager_on_external_kubernetes:
```
Pablogoliva: There's a salt sls in kubernetes_common inside the salt folder called install_helm. Can you please reuse it
?

salt/server_kubernetes/install_kubernetes_server.sls
AkashKumar7902: Done in c83de273 — dropped the `KUBECONFIG=` prefix here and in the other `unless` commands too: the state's `env:` is applied to unless checks as well (verified on Salt 3006: with `env: KUBECONFIG=/nonexistent` the unless fails and the state runs; with the real path it skips).

salt/server_kubernetes/install_kubernetes_server.sls
AkashKumar7902: Done in c83de273 — python3-PyYAML now comes from the pkg_map install point, keyed by base OS (`Leap`, since the controller image is pinned to opensuse156o). It is needed on the controller because in external mode the controller is the machine that runs kubernetes_common/helm_chart.py (which does `import yaml`) to resolve the chart app version before the helm install — same reason install_helm.sls ships python3-yaml on Ubuntu. Note this also moves it from pkg.installed to the pkg_map's pkg.latest + refresh semantics.

One thing I noticed while testing on a Leap 15.6 controller: helm_chart.py uses `subprocess.run(..., text=True)`, which needs Python 3.7+, so update_oci_app_version fails there with system python3 3.6 (pre-existing, unrelated to this change). I can send a tiny separate PR switching it to `universal_newlines=True` if you want.

salt/server_kubernetes/install_kubernetes_server.sls
AkashKumar7902: Done in c83de273 — the duplicated external cert-manager/trust-manager states are deleted; controller/init.sls now includes kubernetes_common/install_helm.sls (gated on the external-cluster + install_cert_manager grains) and that file is parameterized with the same `is_external_cluster` switch already used in install_kubernetes_server.sls: kubeconfig path, the cert/trust gate, and external-only `unless` guards so re-runs against a long-lived external cluster stay no-ops (managed rendering is untouched). External mode also picks up the check_cert_manager_installation / check_trust_manager_installation verification states, and the uyuni namespace state now requires them.

The helm binary itself still comes from the pre-existing unless-guarded install_helm_on_controller state in controller/init.sls (install_kubectl_helm is forced on in external mode), so helm_install_package is skipped externally to avoid a second unguarded install of the same thing.

Two small behavior notes vs the previous head: the external cert-manager install is now the pinned `helm install` (v1.19.2, --timeout 10m0s) guarded by the external-only unless instead of `helm upgrade --install` — identical on the only path where it executes (release absent) — and trust-manager keeps its external unless so re-applies do not bump the release. Verified managed renders identical across the OS/grain matrix (both with a render harness and the real Salt 3006 renderer), and live-tested a fresh external install (cert-manager + trust-manager + uyuni namespace) plus an idempotent second apply against a disposable cluster.

salt/server_kubernetes/install_kubernetes_server.sls
```diff
@@ -150,20 +163,37 @@ transfer_python_management_file:
...
     - name: helm upgrade --install uyuni ./selfsigned -f ./selfsigned/values.yaml -n uyuni
     - cwd: {{ helm_chart_directory }}
     - env:
       - KUBECONFIG: {{ kubeconfig }}
+    {% if is_external_cluster %}
+    - require:
+      - cmd: build_helm_dependencies
+      - file: copy_value_yaml_file
+      - cmd: create_external_kubernetes_uyuni_namespace
+    {% endif %}
```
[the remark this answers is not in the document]
AkashKumar7902: Managed-mode ordering is intentionally left unchanged in this PR: with state_auto_order on, the managed states already execute in declaration order, so declaration-mirroring requires would be behavior-neutral there, and the bar for this change is that managed renders stay identical. The external-only requires are load-bearing, though — write_kubeconfig/install_kubectl live in controller/init.sls, which is ordered after the included files, so external mode cannot rely on auto-order. The managed requisites could be a small follow-up PR so that delta can be reviewed in isolation.

salt/server_kubernetes/install_kubernetes_server.sls
```diff
@@ -150,20 +163,37 @@ transfer_python_management_file:
 update_oci_app_version:
   cmd.run:
     - name: python3 {{ python_helm_chart_path }} -o {{ helm_chart_url }}/{{ helm_chart_name }} --chart-file {{ self_signed_path }}/Chart.yaml {{ devel_flag }}
+    {% if is_external_cluster %}
+    - require:
+      - pkg: install_external_kubernetes_dependencies
+      - cmd: install_helm_on_controller
+      - file: transfer_python_management_file
+      - file: copy_chart_yaml_file
+    {% endif %}
```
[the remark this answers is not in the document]
AkashKumar7902: Managed ordering is intentionally left as-is in this PR (same reasoning as on the install_uyuni_on_kubernetes thread): with state_auto_order on, the managed states already execute in declaration order, and the bar for this change is that managed renders stay identical. The external-only requires are there because in external mode the referenced states live in controller/init.sls, which is ordered after the included files. Can pick up the managed requisites in a follow-up PR.

salt/server_kubernetes/install_kubernetes_server.sls
```diff
@@ -150,20 +163,37 @@ transfer_python_management_file:
...
+{% if grains.get('install_mlm_server') == true or is_external_cluster %}
 
 build_helm_dependencies:
   cmd.run:
     - name: helm dependencies build
     - cwd: {{ self_signed_path }}
+    {% if is_external_cluster %}
+    - require:
+      - cmd: update_oci_app_version
+    {% endif %}
```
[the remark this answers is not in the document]
AkashKumar7902: Same as above — in managed mode build_helm_dependencies already runs right after update_oci_app_version via declaration order (state_auto_order), so this require is only added where it is load-bearing (external mode). Keeping the managed render untouched is the constraint for this PR; happy to add the explicit managed requisites in a follow-up.

salt/controller/install_kubernetes_server.sls
```diff
@@ -0,0 +1,66 @@
...
+    - pkgs:
+      - python3-PyYAML
+
+external_kubernetes_kubeconfig:
+  file.exists:
+    - name: {{ kubeconfig }}
+    {% if grains.get('kubeconfig_content') %}
+    - require:
+      - cmd: write_kubeconfig
+    {% endif %}
```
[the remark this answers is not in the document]
AkashKumar7902: Deferring this one to keep the diff scoped to the review: file.exists already fails naming the missing path (/root/.kube/config). A friendlier fail-fast — non-empty check with an actionable message, or better, validating kubeconfig_path at plan time when kubernetes_cluster_mode = "external" — is a good follow-up together with the pre-existing write_kubeconfig path.

Pablogoliva (approved):

modules/cucumber_testsuite/main.tf
```diff
@@ -121,8 +121,11 @@ locals {
     host_key => lookup(var.host_settings[host_key], "scc_slmicro_pass", null) if var.host_settings[host_key] != null }
 
   minimal_configuration     = { hostname = contains(local.hosts, "proxy") ? local.proxy_full_name : local.server_full_name }
-  server_configuration      = var.kubernetes ? module.server_kubernetes[0].configuration : ( var.container_server ? module.server_containerized[0].configuration : module.server[0].configuration)
-  proxy_configuration       = var.kubernetes ? module.proxy_kubernetes[0].configuration : (var.container_proxy ? module.proxy_containerized[0].configuration : module.proxy[0].configuration)
+  external_kubernetes_server    = var.kubernetes_external_server_hostname != null ? var.kubernetes_external_server_hostname : local.server_full_name
+  external_server_configuration = { id = null, hostname = local.external_kubernetes_server, username = "admin", password = "admin", runtime = "rke2", product_version = module.base.configuration["product_version"], first_user_present = true }
```
Pablogoliva: It should be in a main.tf.

modules/cucumber_testsuite/main.tf
```diff
@@ -121,8 +121,11 @@ locals {
...
+  empty_proxy_configuration     = { id = null, hostname = null, private_mac = null, private_ip = 254, private_name = "proxy", username = "admin", password = "admin" }
```
Pablogoliva: These parameter must be in the main.tf in the ci.

Pablogoliva (changes requested):

Pablogoliva (approved):

Pablogoliva (approved):

PR #2204, opened 2026-07-12
Correctly detect avahi ip during provisioning

maximenoel8:
## What does this PR change?

Fix SSH wait provisioner to support Avahi (mDNS) deployments

## Problem

When use_avahi = true, the VM hostname resolves to a .local mDNS address (e.g. uyuni-server.tf.local). The previous provisioner relied on getent ahostsv4 and host -t A to resolve it,  both go through standard DNS and cannot resolve mDNS multicast names, causing the SSH wait to always time out: `ERROR: timed out waiting for SSH on uyuni-server.tf.local (uyuni-server.tf.local)`

##  Root cause

mDNS (.local) is multicast-based and requires nss-mdns to be configured on the host running Terraform. This is not guaranteed and standard DNS tools cannot see it.

##  Fix

Since VMs already have qemu_agent = true and wait_for_lease = true, libvirt knows the DHCP-assigned IP via the QEMU guest agent, even for Avahi VMs (Avahi is an additional naming layer on top of DHCP, not a replacement). We now pass that IP directly to the provisioner via an environment variable, bypassing name resolution entirely.

##  Resolution order:
  1. libvirt DHCP lease IP (new — primary path, works for local/static-DHCP/Avahi)
  2. getent ahostsv4 — nsswitch chain
  3. avahi-resolve --name — explicit mDNS fallback if lease not available
  4. host -t A — standard DNS fallback
  5. Raw hostname — last resort (previous behavior)

### DIFF
--- a/backend_modules/libvirt/host/main.tf
+++ b/backend_modules/libvirt/host/main.tf
@@ -265,18 +265,40 @@ resource "terraform_data" "provisioning" {
   }
 
   provisioner "local-exec" {
+    environment = {
+      HOST          = local.overwrite_fqdn != "" ? local.overwrite_fqdn : "${libvirt_domain.domain[count.index].name}.${var.base_configuration["domain"]}"
+      IP_FROM_LEASE = local.overwrite_fqdn == "" ? join(" ", libvirt_domain.domain[count.index].network_interface[0].addresses) : ""
+    }
     command = <<-EOF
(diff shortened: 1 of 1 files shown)


cbosdo:
I would have been good to add the analysis from the PR description to the Git commit… And with one single commit GH would have used it as PR description too

PR #2205, opened 2026-07-13
Add optional /etc/hosts entry for the external Kubernetes server FQDN

AkashKumar7902:
## What does this PR change?

> **Note:** this PR is stacked on top of #2199 and only its last commit (`Add optional /etc/hosts entry for the external Kubernetes server FQDN`) is new. It will be rebased once #2199 is merged.

In external cluster mode (#2199) the chosen server FQDN may not be covered by real DNS — common in test labs. The controller then cannot resolve the server it installs and tests, and the user has to add an `/etc/hosts` entry on the controller by hand, even though the controller is a machine sumaform itself provisions.

This adds an optional `kubernetes_external_server_ip` variable to `cucumber_testsuite` (exposed as `kubernetes_server_ip` on the `controller` module, following the same naming split as `kubernetes_external_server_hostname` / `kubernetes_server_fqdn`):

- When set, the controller gets a matching `/etc/hosts` entry for the server FQDN, managed with Salt's `host.present` state (idempotent, no duplicate entries on re-provisioning).
- When unset (the default), nothing changes.

The FQDN for the entry uses the same fallback chain as the install state: `kubernetes_server_fqdn`, then the pre-existing `server` grain.

Tested with:
- `tofu validate` and a full null-backend `apply` of the RKE2 example with external mode and the new variable set.
- Rendering the modified Salt state with the grain set (entry present, correct name/ip), unset (state absent), and with the `server`-grain fallback.
- The equivalent manual `/etc/hosts` entry was what an end-to-end external-cluster deployment (RKE2 on bare metal, controller-driven Uyuni install, storage benchmark run) required in practice; this automates exactly that step.


cbosdo:
> ## What does this PR change?
> 
> > **Note:** this PR is stacked on top of #2199 and only its last commit (`Add optional /etc/hosts entry for the external Kubernetes server FQDN`) is new. It will be rebased once #2199 is merged.
> 
> In external cluster mode (#2199) the chosen server FQDN may not be covered by real DNS — common in test labs. The controller then cannot resolve the server it installs and tests, and the user has to add an `/etc/hosts` entry on the controller by hand, even though the controller is a machine sumaform itself provisions.

You probably will face other problems using the server if you don't have a properly setup DNS with the server FQDN… How are you handling the DNS / DHCP in your VMs? Do you use Avahi or https://github.com/uyuni-project/uyuni/wiki/Libvirt-DNS-and-DHCP-without-Avahi?

salt/controller/install_kubernetes_server.sls
```diff
@@ -0,0 +1,73 @@
+{% set kubeconfig = "/root/.kube/config" %}
```
Pablogoliva: Pls. Do not create another install_kubernetes_sever sls inside the controller, parametrize the one that already exists and execute it from the init.sls as server_kubernetes.install_kubernetes_server.sl as I told you.

Pablogoliva (changes requested):

AkashKumar7902:
we have decided to go on with modifying libvirt with static entry like:
```
<host ip='192.168.122.219'>
      <hostname>rke2-node1</hostname>
      <hostname>rke2</hostname>
    </host>
```

PR #2210, opened 2026-07-15
Remove EOL opensuse156o client tools repo

ktsamis:
## What does this PR change?

openSUSE Leap 15.6 (`opensuse156o`) is EOL, so this drops its client tools repository configuration in `backend_modules/libvirt/host/user_data.yaml`.

The block previously handled both `opensuse156o` and `opensuse160o`; it now only handles `opensuse160o` (the redundant inner conditional was collapsed).

Follow-up to a review comment from @mcalmer on #2172.

PR #2214, opened 2026-07-16
Revert "Remove EOL opensuse156o client tools repo"

ktsamis:
Reverts uyuni-project/sumaform#2210

The previous PR broke deployments because it affected the controller and no venv-salt-minion was available.


Pablogoliva (approved):

nodeg (approved):
LGTM. And yes we still need the repos for the controller.

PR #2224, opened 2026-07-21
Log a list of installed packages

alichiel:
## What does this PR change?

It logs the list of installed packages and their version. The packages can change over time but in this case there should be another information logged.

### DIFF
--- a/backend_modules/libvirt/host/user_data.yaml
+++ b/backend_modules/libvirt/host/user_data.yaml
@@ -1058,4 +1058,4 @@ runcmd:
 
 runcmd:
   # Log a list of installed RPM/DEB packages for the debuigging purposes
-  - "rpm --version &>/dev/null && rpm -qa | sort >/installed_packages.txt || dpkg --list | grep 'ii' >/installed_packages.txt"
+  - "rpm --version &>/dev/null && rpm -qa | sort >/installed_packages.txt || dpkg --list | grep '^ii' >/installed_packages.txt"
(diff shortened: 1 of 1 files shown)


backend_modules/libvirt/host/user_data.yaml
```diff
@@ -1055,3 +1055,7 @@ runcmd:
   - echo "PermitRootLogin yes" > /etc/ssh/sshd_config.d/root.conf
   - systemctl restart sshd.service
 %{ endif ~}
+
+runcmd:
+  # Log a list of installed RPM/DEB packages for the debuigging purposes
+  - "rpm --version &>/dev/null && rpm -qa | sort >/installed_packages.txt || dpkg --list | grep 'ii' >/installed_packages.txt"
```
ktsamis: I would do a check for what kind of system it is, like we do everywhere to run the commands in the right system. But as it is it's correct, so I think it's fine to merge this as well

ktsamis (approved):

backend_modules/libvirt/host/user_data.yaml
juliogonzalez:
```suggestion
  - "rpm --version &>/dev/null && rpm -qa | sort >/installed_packages.txt || dpkg --list | grep '^ii' >/installed_packages.txt"
```
You want to match only installed packages (the status is in the first two characters of the line), otherwise your current grep would match packages with `ii` in the name, even if they are uninstalled and with leftover config files.

alichiel: Thanks, you are absolutely right. Fixed.

alichiel: I was thining about that but also this check might be quicker and smaller since we have just 2 systems - RPM or DEB.
I can still fix it but would prefere to go with the current version.

NamelessOne91 (approved):

PR #2230, opened 2026-07-22
Move test suite controller to openSUSE Leap 16.0

ktsamis:
## What does this PR change?

Migrates the test suite **controller** VM image from openSUSE Leap 15.6 (EOL since April 2026) to openSUSE Leap 16.0.

- `modules/controller/main.tf` now uses `opensuse160o` (already defined in the libvirt and AWS base backends). This is the single, central place the controller image is set — both the cucumber testsuite and build-validation paths call `../controller` with no image override, so this covers all environments.
- `backend_modules/libvirt/host/main.tf`: updated a stale comment that still referenced `opensuse156o` as the controller image (the `opensuse156o` entry itself stays in the `x86_64_v2_images` list — it's still used by Leap 15.6 minion clients).

Only the controller is affected; client/minion, dhcp_dns, proxy, server and mirror images are untouched.

### Repository changes (salt)

**`salt/repos/tools.sls` — controller no longer pulls `systemsmanagement:sumaform:tools`.**
On Leap 16 this repo was only ever added on the **controller** — never on minions. tools.sls is guarded by role (`controller` / `grafana` / `mirror` / evil-minion / `monitored`), and the only one of those that runs on Leap 16 is the controller; plain minions (including the Leap 16 minions in the build validation) do not match the condition and have never referenced this repo. The only package the controller needs from that family is `node_exporter`, which openSUSE Leap 16.0 ships in its base repo, so the controller does not need the tools repo at all on Leap 16. It is therefore removed from the list of roles that add it. `tools` remains in place unchanged for the `grafana` / `mirror` / evil-minion / `monitored` roles, which run on SLE / Leap 15.x where the repo builds normally.

Note: `systemsmanagement:sumaform:tools` has no working Leap 16 build (its packages are decade-old frozen copies whose Go bootstrap sources were removed from Factory, and modern equivalents now live in the Leap 16 base repo). Porting it is only relevant if the grafana/monitoring/evil-minion roles are ever moved to Leap 16, which is out of scope here.

**`salt/controller/init.sls` — health-check repo now uses the Leap 16 naming.**
The `systemsmanagement:Uyuni:healthcheck:Stable` repo names its Leap repositories `openSUSE_Leap_<version>`, so on Leap 16 the controller now points at `openSUSE_Leap_16.0` (older distros keep the previous bare-`osrelease` path). `mgr-health-check` is published and installable there.

This is part of migrating the test-suite controller host OS to Leap 16.0. Companion PRs:
- uyuni: `controller-dev` container Dockerfile → Leap 16.0 — **merged**: https://github.com/uyuni-project/uyuni/pull/12313
- susemanager-ci: environment image lists + Uyuni Master AWS controller override + README matrix → Leap 16.0: https://github.com/SUSE/susemanager-ci/pull/2131

Controller provisioning on Leap 16.0 is validated via the manual core-stage terracumber run alongside the uyuni controller-dev image. Azure `opensuse160o` parity is a deferred follow-up (not required for libvirt-based CI).

Links: https://github.com/SUSE/spacewalk/issues/31270

### DIFF
--- a/backend_modules/libvirt/host/main.tf
+++ b/backend_modules/libvirt/host/main.tf
@@ -4,7 +4,7 @@ locals {
   manufacturer = lookup(var.provider_settings, "manufacturer", "Intel")
   product      = lookup(var.provider_settings, "product", "Genuine")
   // List of images that can benefit from cpu passthrough or host-model.
-  // In particular, opensuse156o covers the controller and sles15sp7o is required for the upgrade test to sles160o
+  // In particular, opensuse160o covers the controller and sles15sp7o is required for the upgrade test to sles160o
   x86_64_v2_images = ["opensuse156o", "almalinux10o", "almalinux9o", "amazonlinux2023o", "libertylinux9o", "openeuler2403o", "oraclelinux9o", "oraclelinux10o", "rocky9o", "rocky10o", "sles15sp7o", "sles160o", "opensuse160o", "slemicro55o", "slmicro60o", "slmicro61o", "slmicro62o", "ubuntu2404o", "ubuntu2604o", "tumbleweedo"]
   combustion_images  = ["slmicro60o", "slmicro61o", "slmicro62o", "sles160o"]
   gpg_keys = [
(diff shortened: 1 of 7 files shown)


Bischoff (approved):

Bischoff:
About the sumaform tools, I suggest we remove them all apart from apt-mirror and evil-minions. I created a card.

I think the controller was using them only because of twopence.

PR #2233, opened 2026-07-23
Add 5.2 case

Bischoff:
## What does this PR change?

Add 5.2 case in conditionals

This fixes a file not found in 5.2 CI:
```
cucumber_testsuite.server_containerized[0].provisioning[0] (remote-exec):
[INFO   ] Executing state cmd.run for [mgrctl exec 'sed -i "s/max_connections = (.*)/max_connections = 400/" 
```

### DIFF
--- a/salt/server_containerized/large_deployment.sls
+++ b/salt/server_containerized/large_deployment.sls
@@ -28,7 +28,7 @@ large_deployment_tune_tomcat_maxthreads:
     - onchanges:
       - cmd: large_deployment_tune_tomcat_stylesheet_copy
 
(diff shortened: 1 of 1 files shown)


ktsamis (approved):
We could expand that if statement as well because the regex kinda worries me but if this works then it's fine as well

Bischoff:
> We could expand that if statement as well because the regex kinda worries me but if this works then it's fine as well

To be very honest, I am a bit afraid too, but Gemini triple-checked it with me.

Expanded versions would be really long here with the additional 5.2 case. It would make the evaluation more complicated with 4 calls to `grains.get()` and 3 `or` operations. It would be harder to maintain.

image

I will merge it.

PR #2241, opened 2026-07-27
Support multiple SUSE minions in the test suite

AkashKumar7902:
## What does this PR change?

Allows the libvirt cucumber testsuite module to honor `host_settings.suse_minion.quantity`, so a benchmark configuration can provision every requested SUSE minion.

Passes the complete generated hostname list to the controller as the `minions` grain and exports it as the JSON array `UYUNI_BENCH_MINIONS`. The existing `MINION` export remains the first hostname for compatibility.

Validation:

- OpenTofu and Terraform example validation
- Controller Salt highstate compilation and exact two-minion JSON rendering
- Null-backend plan assertion that `suse_minion.quantity = 2` reaches the minion module


.github/workflows/ci-validation/salt-controller-validation
```diff
@@ -46,4 +48,12 @@ EOF
 cat testconfig/grains
 echo
 
+rendered_bashrc="$PWD/testconfig/controller-bashrc"
```
Pablogoliva, line 51: This is for other stuff. This workflow is for testing things in the GH actions pipeline, not in the common pipeline. Please don't modify it.

Pablogoliva (changes requested):

Pablogoliva (changes requested):
After a clarification with Akash we have figured out that Uyuni is already able to use more than one minon. So, it can be closed,

Pablogoliva:
After a clarification with Akash we have figured out that Uyuni is already able to use more than one minon. So, it can be closed.

PR #2243, opened 2026-07-28
QE: Move the installation of HEAD to RKE2

Pablogoliva:
## What does this PR change?

Adapt sumaform to move the installation of HEAD to RKE2. The purpose of this PR is to not do anything regarding the installation of MLM or RKE2 during sumaform, just configure the machines for it.

### DIFF
--- a/salt/kubernetes_common/install_helm.sls
+++ b/salt/kubernetes_common/install_helm.sls
@@ -6,15 +6,13 @@
 {% set is_slmicro_6_2 = osfullname == 'SL-Micro' and osrelease == '6.2' %}
 {% set is_ubuntu = osfullname == 'Ubuntu' %}
 {% set is_tumbleweed = osfullname == 'openSUSE Tumbleweed' %}
-{% set is_supported_os = is_sles_15_7 or is_slmicro_6_2 or is_ubuntu or is_tumbleweed %}
-{% if is_supported_os or is_external_cluster %}
+{% set is_supported_os = is_sles_15_7 or is_ubuntu or is_tumbleweed %}
 {% set kubeconfig = "/root/.kube/config" if is_external_cluster else "/etc/rancher/rke2/rke2.yaml" %}
 {% set cert_manager_version = "v1.19.2" %}
 {% set cert_manager_namespace = "cert-manager" %}
 {% set pkg_map = {
     'Ubuntu': ['python3-yaml'],
-    'openSUSE Tumbleweed': ['openssl', 'git'],
-    'SUSE Linux Micro': ['git']
+    'openSUSE Tumbleweed': ['openssl', 'git']
 } %}
 
 {% if osfullname in pkg_map %}
@@ -23,6 +21,8 @@ install_dependencies_helm:
     - pkgs: {{ pkg_map.get(osfullname) }}
     - refresh: True
 {% endif %}
+
+{% if is_supported_os or is_external_cluster %}
 
 {% if not is_external_cluster %}
 helm_install_package:
@@ -88,3 +88,10 @@ check_trust_manager_installation:
 {% endif %}
 
 {% endif %}
+
+variables_helm:
+  file.managed:
+    - name: /etc/profile.d/helm_vars.sh
+    - contents: |
+        export CERT_MANAGER_VERSION={{ cert_manager_version }}
+        export CERT_MANAGER_NAMESPACE={{ cert_manager_namespace }}
--- a/salt/kubernetes_common/install_rke2.sls
+++ b/salt/kubernetes_common/install_rke2.sls
@@ -46,29 +46,12 @@ tls-san_setup_file:
         {% endif %}
     - makedirs: True
 
+{% if not is_slmicro_6_2 %}
 
 rke2_install:
   cmd.run:
     - name: curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_VERSION="{{ rke2_version }}" sh -
-    - env:
-        {% if is_slmicro_6_2 %}
-        - INSTALL_RKE2_METHOD: rpm
-        - INSTALL_RKE2_SELINUX: true
-        {% endif %}
     - unless: systemctl is-active rke2-server
-    {% if is_slmicro_6_2  and grains.get('scc_slmicro_pass') %} 
-    - require:
-      - cmd: register_to_scc
-      - cmd: apply_transition_scc
-    {% endif %}
-
-{% if is_slmicro_6_2 %}
-apply_transition_rke2:
-  cmd.run:
-    - name: transactional-update apply
-    - require:
-      - cmd: rke2_install
-{% endif %}
 
 rke2_server_enable:
   service.enabled:
@@ -76,10 +59,6 @@ rke2_server_enable:
     - require:
       - cmd: rke2_install
       - file: tls-san_setup_file
-      {% if is_slmicro_6_2 and grains.get('scc_slmicro_pass') %}
-      - cmd: register_to_scc
-      - cmd: apply_transition_scc
-      {% endif %}
 
 {% if is_tumbleweed %}
 rke2_selinux_install:
@@ -129,12 +108,15 @@ link_ctr_rke2:
     - require:
       - cmd: rke2_server_wait_until_active
 
+{% endif %}
+
 variables_rke2:
   file.managed:
     - name: /etc/profile.d/rke2_vars.sh
     - contents: |
         export PATH=$PATH:/opt/rke2/bin
         export KUBECONFIG={{ kubeconfig }}
+        export RKE2_VERSION={{ rke2_version }}
 
 
 {% endif %}
--- a/salt/kubernetes_common/install_traefik.sls
+++ b/salt/kubernetes_common/install_traefik.sls
@@ -6,6 +6,7 @@
 {% set is_ubuntu = osfullname == 'Ubuntu' %}
 {% set is_tumbleweed = osfullname == 'openSUSE Tumbleweed' %}
 {% set is_supported_os = is_sles_15_7 or is_slmicro_6_2 or is_ubuntu or is_tumbleweed %}
+{% set traefik_file = "/root/kubernetes-crd-definition-v1.yml" %}
 {% if is_supported_os %}
 {% set pkg_map = {
     'openSUSE Tumbleweed': ['iptables']
@@ -20,14 +21,22 @@ install_dependencies_traefik:
 
 copy_traefik_installation_file:
   file.managed:
-  - name: /root/kubernetes-crd-definition-v1.yml
+  - name: {{ traefik_file }}
   - source: salt://kubernetes_common/kubernetes-crd-definition-v1.yml
   - makedirs: true
 
+{% if not is_slmicro_6_2 %}
 install_traefik:
   cmd.run:
-    - name: kubectl apply -f /root/kubernetes-crd-definition-v1.yml
+    - name: kubectl apply -f {{ traefik_file }}
     - env:
       - KUBECONFIG: {{ kubeconfig }}
+{% endif %}
+
+variables_traefik:
+  file.managed:
+    - name: /etc/profile.d/traefik_vars.sh
+    - contents: |
+        export TRAEFIK_FILE={{ traefik_file }}
 
 {% endif %}
--- a/salt/kubernetes_common/set_up_local-path-provisioner.sls
+++ b/salt/kubernetes_common/set_up_local-path-provisioner.sls
@@ -4,13 +4,16 @@
 {% set is_slmicro_6_2 = osfullname == 'SL-Micro' and osrelease == '6.2' %}
 {% set is_ubuntu = osfullname == 'Ubuntu' %}
 {% set is_tumbleweed = osfullname == 'openSUSE Tumbleweed' %}
-{% set is_supported_os = is_sles_15_7 or is_slmicro_6_2 or is_ubuntu or is_tumbleweed %}
-{% if is_supported_os %}
+{% set is_supported_os = is_sles_15_7 or is_ubuntu or is_tumbleweed %}
 {% set kubeconfig = "/etc/rancher/rke2/rke2.yaml" %}
+{% set local_path_storage_file = "/root/local-path-storage.yaml" %}
+{% set local_path_namespace = "local-path-storage" %}
 {% set storage_class = grains.get('kubernetes_storage_class') | default('local-path', true) %}
 {% set local_path = grains.get('local_path_provisioner_path') | default('/opt/local-path-provisioner', true) %}
 {% set default_class = grains.get('local_path_provisioner_default_class', true) %}
 {% set pkg_map = {} %}
+
+{% if is_supported_os %}
 
 {% if osfullname in pkg_map %}
 install_dependencies_helm_provisioner:
@@ -21,14 +24,14 @@ install_dependencies_helm_provisioner:
 
 copy_local-path-storage_installation_file:
   file.managed:
-    - name: /root/local-path-storage.yaml
+    - name: {{ local_path_storage_file }}
     - source: salt://kubernetes_common/local-path-storage.yaml
     - template: jinja
     - makedirs: true
 
 install_local_path_provisioner:
   cmd.run:
-    - name: kubectl apply -f /root/local-path-storage.yaml
+    - name: kubectl apply -f {{ local_path_storage_file }}
     - env:
       - KUBECONFIG: {{ kubeconfig }}
     - require:
@@ -62,7 +65,7 @@ set_local-path-provisioner_selinux_tags:
 
 restart_local-path-provisioner_pods:
   cmd.run:
-    - name: kubectl delete pods --all -n local-path-storage
+    - name: kubectl delete pods --all -n {{ local_path_namespace }}
     - require:
       - cmd: set_local-path-provisioner_selinux_tags
       - file: create_local-path-provisioner_directory
@@ -71,4 +74,15 @@ restart_local-path-provisioner_pods:
 
 {% endif %}
 
+
 {% endif %}
+
+variables_local-path-provisioner:
+  file.managed:
+    - name: /etc/profile.d/local-path-provisioner_vars.sh
+    - contents: |
+        export LOCAL_PATH_PROVISIONER_PATH={{ local_path_storage_file }}
+        export LOCAL_PATH_PROVISIONER_STORAGE_CLASS={{ storage_class }}
+        export LOCAL_PATH={{ local_path }}
+        export LOCAL_PATH_NAMESPACE={{ local_path_namespace }}
+
--- a/salt/proxy_kubernetes/config_node.sls
+++ b/salt/proxy_kubernetes/config_node.sls
@@ -75,11 +75,6 @@ ssh_config_proxy_kubernetes:
     - group: root
     - mode: 700
 
-create_uyuni_namespace_proxy:
-  cmd.run:
-  - name : kubectl create namespace {{ Namespace }}
-  - env:
-      - KUBECONFIG: {{ kubeconfig }}
 
 #####################################################
 ################ Setup proxy certs ##################
@@ -102,6 +97,14 @@ copy_certs_generator:
         ProxyFQDN: {{ ProxyFQDN }}
         Namespace: {{ Namespace }}
         ProxyName: {{ ProxyName }}
+
+{% if not is_slmicro_6_2 %}
+
+create_uyuni_namespace_proxy:
+  cmd.run:
+  - name : kubectl create namespace {{ Namespace }}
+  - env:
+      - KUBECONFIG: {{ kubeconfig }}
 
 check_ssh_communication:
   cmd.run:
@@ -179,5 +182,7 @@ copy_uyuni_ca:
   - require:
       - file: ssh_private_key_proxy_kubernetes_for_server
       - file: ssh_config_proxy_kubernetes
+
+{% endif %}
 
 {% endif %}
--- a/salt/proxy_kubernetes/install_kubernetes_proxy.sls
+++ b/salt/proxy_kubernetes/install_kubernetes_proxy.sls
@@ -60,16 +60,18 @@ transfer_python_management_file:
 
 {% if grains.get('install_helm') == true %}
 
+copy_manifest_uyuni_ingress_proxy:
+  file.managed:
+    - name: /var/lib/rancher/rke2/server/manifests/uyuni-ingress-proxy.yaml
+    - source: salt://proxy_kubernetes/uyuni-ingress-proxy.yaml
+    - makedirs: True
+
+{% if not is_slmicro_6_2 %}
 update_oci_app_version_proxy:
   cmd.run:
     - name: python3 {{ python_helm_chart_path }} -o {{ helm_chart_url }}/{{ helm_chart_name }} --chart-file {{ self_signed_path }}/Chart.yaml {{ devel_flag }}
 
 {% if grains.get('install_rke2') == true and grains.get('install_mlm_proxy') == true %}
-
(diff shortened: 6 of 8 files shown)


salt/kubernetes_common/set_up_local-path-provisioner.sls
```diff
@@ -71,4 +73,14 @@ restart_local-path-provisioner_pods:
 
 {% endif %}
 
+
+variables_local-path-provisioner:
```
ktsamis, line 80: This block is inside `{% if is_supported_os %}` on line 8, and line 7 takes SL Micro out of `is_supported_os`.

So these variables never get written on SL Micro, which is the only place that needs them. The feature in uyuni #12223 checks `LOCAL_PATH` and `LOCAL_PATH_NAMESPACE` and would fail.

`variables_helm` and `variables_traefik` are outside their guards, this one needs the same.

salt/kubernetes_common/set_up_local-path-provisioner.sls
```diff
@@ -71,4 +73,14 @@ restart_local-path-provisioner_pods:
...
+  file.managed:
+    - name: /etc/profile.d/local-path-provisioner_vars.sh
+    - contents: |
+        export LOCAL_PATH_PROVISIONER_PATH={{ local_path_storage_file }}
+        export LOCAL_PATH_PROVISIONER_STORAGE_FILE={{ storage_class }}
```
ktsamis: This holds a storage class name, not a file. `LOCAL_PATH_PROVISIONER_STORAGE_CLASS` would be clearer.

salt/kubernetes_common/install_traefik.sls
```diff
@@ -20,14 +21,22 @@ install_dependencies_traefik:
...
+    - name: kubectl apply -f {{ traefik_file }}
     - env:
       - KUBECONFIG: {{ kubeconfig }}
+{% endif %}
+
+variables_traefik:
+  file.managed:
+    - name: /etc/profile.d/traefik_vars.sh
+    - contents: |
+        export TRAEFIK_PATH={{ traefik_file }}
```
ktsamis: The feature in uyuni #12223 reads `TRAEFIK_FILE`. One of the two names needs to change.

salt/server_kubernetes/install_kubernetes_server.sls
```diff
@@ -234,6 +240,26 @@ save_script_to_get_pod_name:
     - user: root
     - group: root
 
+
+variables_server_kubernetes:
+  file.managed:
+    - name: /etc/profile.d/server_kubernetes_vars.sh
+    - contents: |
+        export PYTHON_HELM_CHART_PATH={{ python_helm_chart_path }}
+        export HELM_CHART_DIRECTORY={{ helm_chart_directory }}
```
ktsamis, line 265: `SELF_SIGNED_PATH` is missing here. The feature in uyuni #12223 checks for it, and uses it to cd into the chart directory before the helm install.

salt/server_kubernetes/set-persistent-volumes.sls
```diff
@@ -1,3 +1,10 @@
+{% set osfullname = grains['osfullname'] %}
+{% set osrelease = grains['osrelease'] %}
+{% set is_sles_15_7 = osfullname == 'SLES' and osrelease == '15.7' %}
+{% set is_slmicro_6_2 = osfullname == 'SL-Micro' and osrelease == '6.2' %}
+{% set is_ubuntu = osfullname == 'Ubuntu' %}
+{% set is_tumbleweed = osfullname == 'openSUSE Tumbleweed' %}
+{% set is_supported_os = is_sles_15_7 or is_ubuntu or is_tumbleweed %}
```
ktsamis, line 6: Taking SL Micro out here means `create_uyuni_namespace` and the two PVs no longer run on it.

The feature in uyuni #12223 runs `helm upgrade --install uyuni ... -n uyuni` without `--create-namespace`, so the namespace will not exist. Either this state stays, or the feature creates it.

Are the static PVs still needed on SL Micro, or does local-path cover that?

salt/server_kubernetes/set-persistent-volumes.sls
```diff
@@ -49,3 +58,5 @@ apply_var_pgsql_file:
       - cmd: create_uyuni_namespace
       - file: copy_var_pgsql_file
 {% endif %}
+
+{% endif %}
```
ktsamis: The file lost its trailing newline.

backend_modules/libvirt/host/combustion
```diff
@@ -190,7 +190,7 @@ PACKAGES="$PACKAGES helm"
 PACKAGES="$PACKAGES bash-completion ca-certificates-suse"
 %{ endif }
 
-%{ if container_server || server_kubernetes }
+%{ if container_server }
```
ktsamis: This drops mgradm and mgrctl from every kubernetes server, not only SL Micro.

remote_node.rb sets `@has_mgrctl` from `which mgrctl`, and uses it to decide whether a command runs inside the pod or on the host. So this changes how the whole RKE2 suite runs on SLES, Tumbleweed and Ubuntu as well.

They are installed on mlm-ci-head-rke2-server right now:

```
mgradm-5.3.1-150750.1.1.develHead.x86_64
mgrctl-5.3.1-150750.1.1.develHead.x86_64
```

Was that the plan, or should it be scoped to SL Micro?

salt/proxy_kubernetes/install_kubernetes_proxy.sls
```diff
@@ -85,13 +87,14 @@ uncompress_config_tar:
     - name: tar -xf {{ helm_chart_directory }}/config.tar.gz -C {{ helm_chart_directory }}/
     - cwd: {{ helm_chart_directory }}
 
-install_uyuni_on_kubernetes:
+install_uyuni_proxy_on_kubernetes:
   cmd.run:
-    - name: helm upgrade --install uyuni ./selfsigned -f ./selfsigned/values.yaml -n uyuni --set-file global.ssh=ssh.yaml --set-file global.config=config.yaml --set-file global.httpd=httpd.yaml
+    - name: helm upgrade --install uyuni-proxy ./selfsigned -f ./selfsigned/values.yaml -n uyuni --set-file global.ssh=ssh.yaml --set-file global.config=config.yaml --set-file global.httpd=httpd.yaml
```
ktsamis, line 92: This renames the helm release from `uyuni` to `uyuni-proxy`. Looks like a real fix, but it is not really part of moving the RKE2 install, and anything looking up the old release name would need updating too.

Would you mind splitting it into its own PR?

ktsamis (changes requested):
I like the direction, taking the RKE2 install out of sumaform for SL Micro solves the reboot problem cleanly.

I went through this together with uyuni #12223, since the two have to agree on the profile.d variables. A few of them do not match yet, and one block ends up in the wrong guard. Comments inline.

One thing I would rather ask about than change: the combustion edit looks wider than the rest of the PR.

ktsamis:
Just a note that this needs a rebase before it can merge, it conflicts with master at the moment.

salt/proxy_kubernetes/install_kubernetes_proxy.sls
Pablogoliva: That's a name, to distinguish them. Not a real fix, but yes, I can move it for sure.

salt/server_kubernetes/set-persistent-volumes.sls
Pablogoliva: The PVs can not be created in a RKE2 cluster that doesn't exists. And the purpouse of this PR is to prepare sumaform to not install MLM and nothing regarding kubernetes to do it in the testsuite in a cucumber feature, so it's something to do after the installation in a cucumber feature.

salt/server_kubernetes/set-persistent-volumes.sls
Pablogoliva: Solved

backend_modules/libvirt/host/combustion
Pablogoliva: I was deleting those packages as a way to avoid the use of mgrctl, command that was failing because MLM/Uyuni was not installed, but it's not the solution. I have to find a way to workaround the initialization of cucumber without the api_test, the cobbler test and the chrome driver. So reverting it.

salt/proxy_kubernetes/install_kubernetes_proxy.sls
Pablogoliva: Also because I discovered that there were two different salt blocks with the same name.

Pablogoliva: So we can leave it here, it's not a big deal.

salt/server_kubernetes/install_kubernetes_server.sls
Pablogoliva: Added.

salt/kubernetes_common/set_up_local-path-provisioner.sls
Pablogoliva: So moving the env variables out of the if.

salt/kubernetes_common/set_up_local-path-provisioner.sls
Pablogoliva: Ok, changing the name.

salt/kubernetes_common/install_traefik.sls
Pablogoliva: Renaming it as traefik file then.

salt/kubernetes_common/set_up_local-path-provisioner.sls
ktsamis: Guard is fixed now, thanks. The names are still off though, see the other thread.

salt/kubernetes_common/install_traefik.sls
ktsamis: Fixed.

salt/server_kubernetes/set-persistent-volumes.sls
ktsamis: Fixed.

backend_modules/libvirt/host/combustion
ktsamis: Reverted, thanks.

salt/proxy_kubernetes/install_kubernetes_proxy.sls
ktsamis: Fine, leave it here. Nothing else in sumaform looks up the release name, so it is safe.

salt/kubernetes_common/set_up_local-path-provisioner.sls
ktsamis: Renamed, but the two sides still do not agree. The feature in uyuni#12223 checks `LOCAL_PATH_PROVISIONER_STORAGE_FILE`, and it checks `LOCAL_PATH_STORAGE_FILE` for the yaml instead of `LOCAL_PATH_PROVISIONER_PATH`. So both of those scenarios fail as it is.

I would go with `LOCAL_PATH_STORAGE_FILE` for the yaml and `LOCAL_PATH_PROVISIONER_STORAGE_CLASS` for the class, and change the feature to match.

salt/server_kubernetes/install_kubernetes_server.sls
ktsamis: It is there now, but it landed inside `{% if grains.get('install_mlm_server') == true or is_external_cluster %}` on line 207.

The whole point of the PR is that sumaform stops installing MLM on SL Micro. The moment that grain goes false, the file is never written and `SELF_SIGNED_PATH`, `HELM_CHART_DIRECTORY` and `PYTHON_HELM_CHART_PATH` all disappear. Three scenarios in uyuni#12223 fail then.

Can you move the block out of that guard, next to the other `variables_*` ones?

salt/server_kubernetes/set-persistent-volumes.sls
ktsamis: Fair point on the PVs.

But the namespace is still nobody's job. uyuni#12223 runs `helm upgrade --install uyuni ... -n uyuni` with no `--create-namespace`, and no scenario creates it. So the install fails on a missing namespace.

Can the feature create the namespace, and the PVs too? And I am still curious about the other half: are the static PVs needed at all on SL Micro, or does local-path cover it?

salt/kubernetes_common/install_helm.sls
```diff
@@ -24,7 +22,9 @@ install_dependencies_helm:
     - refresh: True
 {% endif %}
 
+{% if is_supported_os %}
```
ktsamis: This one is a regression, and it is not about SL Micro.

It used to be `{% if is_supported_os or is_external_cluster %}`, and line 9 just dropped SL Micro from `is_supported_os`. In external cluster mode this state runs on the controller, which is `opensuse160o`. That was never in `is_supported_os` - it only got in through the `or is_external_cluster`.

So cert-manager and trust-manager now quietly never install on the controller. Can you put the `or is_external_cluster` back?

salt/kubernetes_common/install_rke2.sls
```diff
@@ -46,40 +46,19 @@ tls-san_setup_file:
         {% endif %}
     - makedirs: True
 
+{% if not is_slmicro_6_2 %}
 
 rke2_install:
   cmd.run:
     - name: curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_VERSION="{{ rke2_version }}" sh -
```
ktsamis, line 53: `INSTALL_RKE2_SELINUX=true` is gone here, and the feature in uyuni#12223 only passes `INSTALL_RKE2_METHOD=rpm`. But `config.yaml` above still sets `selinux: true` for SL Micro.

Was dropping it on purpose, or did it just get lost in the move?

salt/kubernetes_common/install_rke2.sls
```diff
@@ -46,40 +46,19 @@ tls-san_setup_file:
         {% endif %}
     - makedirs: True
 
+{% if not is_slmicro_6_2 %}
```
ktsamis, line 49: `apply_transition_rke2` is gone, but the SCC block on lines 23-33 still runs `transactional-update register` and `transactional-update apply` on SL Micro, and nothing reboots after the highstate.

Then the testsuite runs its own `transactional-update` for the rpm install. That one is based on the running root, not on the pending snapshot, so the registration can get thrown away.

If SCC registration is still needed on this host, there has to be a reboot in between. Or move the registration into the feature too.

salt/proxy_kubernetes/config_node.sls
```diff
@@ -103,6 +98,14 @@ copy_certs_generator:
         Namespace: {{ Namespace }}
         ProxyName: {{ ProxyName }}
 
+{% if not is_slmicro_6_2 %}
```
ktsamis, line 101: This switches off the namespace creation and the whole cert exchange for the proxy on SL Micro, but uyuni#12223 only covers the server. So the proxy ends up half done: sumaform stops doing it and nothing else picks it up.

Is the proxy in scope for this PR, or does it get its own one?

salt/server_kubernetes/set-persistent-volumes.sls
ktsamis, line 6: Nit. Everywhere else in the PR you use `{% if not is_slmicro_6_2 %}`. Here it is a whitelist instead. Same result today, but it is inconsistent, and `is_slmicro_6_2` on line 4 ends up unused.

ktsamis (changes requested):
Second pass, went through it together with uyuni#12223 again.

Fixed since last time: the traefik variable name matches, the combustion change is gone, the newline is back, and the local-path variables are out of the guard. Thanks.

Still open, I reopened those threads (or tried to, they were marked resolved):

- the local-path variable names still do not match the feature
- `SELF_SIGNED_PATH` and `HELM_CHART_DIRECTORY` landed inside a guard that will be false on SL Micro
- nothing creates the `uyuni` namespace any more

And four new things inline. The first one is a real regression outside SL Micro.

One aside for uyuni#12223, since you own both: the install scenario does `cd $SELF_SIGNED_PATH && helm upgrade --install uyuni ./selfsigned`. From inside `selfsigned` there is no `./selfsigned`. sumaform runs that with `cwd: {{ helm_chart_directory }}`, so it should be `cd $HELM_CHART_DIRECTORY`.

Nit, no need to answer: stray double blank lines at install_rke2.sls:120, set_up_local-path-provisioner.sls:76, install_kubernetes_server.sls:254 and config_node.sls:77.

salt/kubernetes_common/install_helm.sls
Pablogoliva: Yes, my bad. The is_external_cluster is missing, good catch.

Pablogoliva: Done

salt/kubernetes_common/install_rke2.sls
Pablogoliva: The variable INSTALL_RKE2_SELINUX doesn't actually exists during the installation of RKE2, it was added by an AI and I believed that it did exist, but I discovered that no during this PR. 
Meanwhile the variable: INSTALL_RKE2_METHOD it's just neccesary for transactional systems, in non-transactional systems this variable is useless that's why I removed it from here.

salt/proxy_kubernetes/config_node.sls
Pablogoliva: Yes, it's intended. We can not create a namespace if RKE2 is not installed and the matter of the following PR is to install RKE2 and MLM there. So we can not create the namespace here and I will create it later in the testsuite PR.

salt/server_kubernetes/install_kubernetes_server.sls
Pablogoliva: Moving it out of all the ifs.

salt/server_kubernetes/set-persistent-volumes.sls
Pablogoliva: Because here I have noting to do with sl_micro 6.2, I can not create the persistent volumes because RKE2 is not installed and ther's nothing else done here. I can remove the check to not leave useless things.

salt/kubernetes_common/set_up_local-path-provisioner.sls
Pablogoliva: I have switched the testsuite PR to a draft because there are things to do whit cucumber before. So the name of the variables will be addressed in a future PR once I solve the problem of cucumber without MLM installed.

salt/server_kubernetes/set-persistent-volumes.sls
Pablogoliva: It will be done in the testsuite, after the installation of RKE2 for SLmicro 6.2. And for the rest of the OSes it's creating the namespace.

salt/kubernetes_common/install_rke2.sls
Pablogoliva: The SCC registration can not be moved to the testsuite because to safely manage the SCC credential I have to save them in the jenkins worker, and the testsuite has no easy access to the jenkins-worker, so I have to do this step here. The `transactional-update apply` I execute after the registration to SCC I think that is enough to register it to SCC and I can do a reboot before starting the feature in the installation feature if I see that something fails.

Pablogoliva:
> Second pass, went through it together with uyuni#12223 again.
> 
> Fixed since last time: the traefik variable name matches, the combustion change is gone, the newline is back, and the local-path variables are out of the guard. Thanks.
> 
> Still open, I reopened those threads (or tried to, they were marked resolved):
> 
> * the local-path variable names still do not match the feature
> * `SELF_SIGNED_PATH` and `HELM_CHART_DIRECTORY` landed inside a guard that will be false on SL Micro
> * nothing creates the `uyuni` namespace any more
> 
> And four new things inline. The first one is a real regression outside SL Micro.
> 
> One aside for uyuni#12223, since you own both: the install scenario does `cd $SELF_SIGNED_PATH && helm upgrade --install uyuni ./selfsigned`. From inside `selfsigned` there is no `./selfsigned`. sumaform runs that with `cwd: {{ helm_chart_directory }}`, so it should be `cd $HELM_CHART_DIRECTORY`.
> 
> Nit, no need to answer: stray double blank lines at install_rke2.sls:120, set_up_local-path-provisioner.sls:76, install_kubernetes_server.sls:254 and config_node.sls:77.

The matter of this PR is to avoid the installation of MLM and RKE2 on SLmicro 6.2 to do it in the testsuite, so the name of the variables and things like this can be addressed in the next PR. I have to look for a way to start just cucumber wihout the chrome driver, the api_test and the cobbler_test instances before actually installing MLM and RKE2. So it will be addressed later.

maximenoel8:
Nice idea overall, the pattern of always writing the variables_* states outside the OS guards makes sense.

One thing that looks wrong in install_helm.sls:


```
  {% if is_supported_os and is_external_cluster%}
  {% if not is_external_cluster %}
  helm_install_package:
```

The outer condition requires is_external_cluster to be true, but the inner one requires it to be false — so helm_install_package can never run. Pretty sure you want or here, same as the original. The old logic worked because SL-Micro 6.2 is already excluded from is_supported_os in this file.

Also tiny: missing space before %} on that same line.

Pablogoliva:
> {% if is_supported_os and is_external_cluster%}

image
It's just because there's an specific block I don't want to execute because in external cluster helm is already installed.

> Nice idea overall, the pattern of always writing the variables_* states outside the OS guards makes sense.
> 
> One thing that looks wrong in install_helm.sls:
> 
> ```
>   {% if is_supported_os and is_external_cluster%}
>   {% if not is_external_cluster %}
>   helm_install_package:
> ```
> 
> The outer condition requires is_external_cluster to be true, but the inner one requires it to be false — so helm_install_package can never run. Pretty sure you want or here, same as the original. The old logic worked because SL-Micro 6.2 is already excluded from is_supported_os in this file.
> 
> Also tiny: missing space before %} on that same line.

It's just for a specific salt block  I want to avoid because in external cluster helm is already isntalled. 
image

salt/kubernetes_common/install_helm.sls
ktsamis: Reopening, this is still `and` on line 25 so the block never runs at all.

On a normal RKE2 server `is_external_cluster` is false, and on the external cluster controller `is_supported_os` is false because the controller is Leap. Both cases skip `helm_install_package` and cert-manager, and `build_helm_dependencies` then fails with `helm: command not found`.

```suggestion
{% if is_supported_os or is_external_cluster %}
```

The inner `{% if not is_external_cluster %}` already keeps `helm_install_package` off the controller, so `or` is enough.

Pablogoliva: Done

Pablogoliva: And regarding the leap thing, it's true I have to add leap as a supported OS. I will once I start the other pipeline for becnhmarking.

ktsamis: Fixed in 41bf375c.

ktsamis (approved):
The `and`/`or` regression in `install_helm.sls` is fixed, so helm and cert-manager install again on both the RKE2 servers and the external cluster controller. Everything else from the earlier passes is addressed, and the variable names, the `uyuni` namespace and the PVs are fine to settle in the testsuite PR.

Two things to keep in mind for that one, neither blocking here:

- the SCC block still does `transactional-update register` and `apply` with no reboot, so the rke2 install in the feature needs `transactional-update --continue` or a reboot in between, otherwise the registration is based on the running root and gets dropped
- the install scenario does `cd $SELF_SIGNED_PATH && helm upgrade --install uyuni ./selfsigned`, and from inside `selfsigned` there is no `./selfsigned`. sumaform runs it with `cwd: {{ helm_chart_directory }}`, so it should be `cd $HELM_CHART_DIRECTORY`

PR #2244, opened 2026-07-28
Raspi OS 13 minion goes to ARM hypervisor

Bischoff:
## What does this PR change?

Raspi OS 13 minion goes to ARM hypervisor

### DIFF
--- a/README_BUILD_VALIDATION_MODULE.md
+++ b/README_BUILD_VALIDATION_MODULE.md
@@ -131,7 +131,7 @@ Some SLE Micro / SL Micro `*_sshminion` modules exist only as commented-out work
 
 - `salt_migration_minion` is a dedicated SLES 15 SP5 minion used to test migration from OS Salt to Salt bundle and is deployed with `install_salt_bundle = false`.
 - `slemicro52_minion`, `slemicro53_minion`, and `slemicro54_minion` are kept with `install_salt_bundle = false` because of the in-file workaround comment.
-- `opensuse156arm_*` and `opensuse160arm_*` get their location-specific extension and FQDN handling from the module code.
+- `opensuse156arm_*`, `opensuse160arm_*`, and `raspios13_*` get their location-specific extension and FQDN handling from the module code.
 - `opensuse160arm_minion` and `opensuse160arm_sshminion` currently use `./salt/controller/id_ed25519.pub` directly for `ssh_key_path` in the module code.
 
 ### `BASE_CONFIGURATIONS`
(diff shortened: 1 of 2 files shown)


ktsamis (commented):
`raspios13_minion` keeps `ssh_key_path = var.controller_public_ssh_key_path` (line 725)
while `raspios13_sshminion` now uses `"./salt/controller/id_ed25519.pub"` (line 1302).
Before this PR both used the variable, and every other pair in the file is consistent
(`opensuse156arm_*` both use the variable, `opensuse160arm_*` both use the ed25519 path).

Is the split intentional, or did the sshminion pick up the hardcoded path from the
`opensuse160arm_sshminion` block above it? If it should be the ed25519 key, the minion needs the same change, and the README note on line 135 should list raspios13 too.

Bischoff:
> `raspios13_minion` keeps `ssh_key_path = var.controller_public_ssh_key_path` (line 725) while `raspios13_sshminion` now uses `"./salt/controller/id_ed25519.pub"` (line 1302). Before this PR both used the variable, and every other pair in the file is consistent (`opensuse156arm_*` both use the variable, `opensuse160arm_*` both use the ed25519 path).
> 
> Is the split intentional, or did the sshminion pick up the hardcoded path from the `opensuse160arm_sshminion` block above it? If it should be the ed25519 key, the minion needs the same change, and the README note on line 135 should list raspios13 too.

No, good catch, it comes from a hasty copy and paste. Fixing.

README_BUILD_VALIDATION_MODULE.md
```diff
@@ -131,7 +131,7 @@ Some SLE Micro / SL Micro `*_sshminion` modules exist only as commented-out work
 
 - `salt_migration_minion` is a dedicated SLES 15 SP5 minion used to test migration from OS Salt to Salt bundle and is deployed with `install_salt_bundle = false`.
 - `slemicro52_minion`, `slemicro53_minion`, and `slemicro54_minion` are kept with `install_salt_bundle = false` because of the in-file workaround comment.
-- `opensuse156arm_*` and `opensuse160arm_*` get their location-specific extension and FQDN handling from the module code.
+- `opensuse156arm_*`, `opensuse160arm_*`, and raspios13_* get their location-specific extension and FQDN handling from the module code.
```
ktsamis:
```suggestion
- `opensuse156arm_*`, `opensuse160arm_*`, and `raspios13_*` get their location-specific extension and FQDN handling from the module code.
```

Bischoff: Right.

ktsamis (approved):

PR #2249, opened 2026-07-30
Add use_dhcp_ip provider setting for VMs not declared in DNS

maximenoel8:
###  Problem

Terraform provisioning fails for libvirt VMs that are not declared in DNS infrastructure.
The SSH connection and SSH-readiness check both rely on resolving the VM's FQDN (e.g.
toto-mlm-5-2-server.mgr.suse.de), which fails when developers intentionally deploy without registering the host in DNS.

The FQDN-based approach was introduced to handle deployments where the libvirt DHCP lease returns an unroutable IP — DNS provides the correct routable address in those cases. Both approaches are valid depending on the infrastructure setup, but there was no way to choose between them.

###  Solution

Add a use_dhcp_ip boolean flag to provider_settings (default: false) that controls how Terraform connects to VMs during provisioning:

- use_dhcp_ip = false (default): existing behavior — resolves FQDN via DNS. Safe for VMs declared in DNS, handles unroutable DHCP lease IPs.
- use_dhcp_ip = true: uses the IPv4 address from the libvirt DHCP lease directly. Required for VMs not declared in DNS, assumes the lease IP is routable.

Both the local-exec SSH-readiness check and the connection block respect the flag, so the behavior is consistent end-to-end.

##  Changes

  - backend_modules/libvirt/host/main.tf: add use_dhcp_ip = false default to provider_settings; branch local-exec script and connection host on the flag
  - backend_modules/libvirt/README.md: document use_dhcp_ip in the host modules provider_settings table
  - README_ADVANCED.md: add "Connecting to VMs not in DNS" section explaining when to use each mode

## Tests

 - CI: https://ci.suse.de/user/manager/my-views/view/RRTG/job/manager-4.3-dev-acceptance-tests/3568/
 Need test on dev env.

### DIFF
--- a/README_ADVANCED.md
+++ b/README_ADVANCED.md
@@ -14,7 +14,7 @@ Legal values for released software are:
 - `5.0-released` (latest released maintenance update for SUSE Manager 5.0 and Tools)
 - `5.1-released` (latest released maintenance update for Multi Linux Manager 5.1 and Tools)
 - `5.2-released` (latest released maintenance update for Multi Linux Manager 5.2 and Tools)
-- `uyuni-released` (latest released version for Uyuni Server, Proxy and Tools, from systemsmanagement:Uyuni:Stable)
+- `uyuni-released` (latest released version for Uyuni Server, Proxy, and Tools, from systemsmanagement:Uyuni:Stable)
 
 Legal values for work-in-progress software are:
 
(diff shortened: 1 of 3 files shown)


Bischoff (approved):

m-czernek (approved):
+1, tested with this `main.tf`:

```hcl
module "suma-server" {
  source = "./modules/server_containerized"
  name = "mlm-5-2-server"
  product_version = "5.2-released"
  image = "slmicro62o"
  provider_settings = {
    use_dhcp_ip = true
  }
}
```

and this finally works as expected! Thank you!

PR #2268, opened 2026-08-19
Fix expose vm_password variable

maximenoel8:
## What does this PR change?

Add `sensitive = true` to credential variables

  Reported by Gap Hunter Labs (private disclosure).

  Their scan flagged `cc_password` missing `sensitive = true` in the aws,
  azure, and feilong backends. We confirmed and extended the scope:

  - `cc_username` and `cc_password` were both missing `sensitive = true`
    across all backends (the null backend was also affected, not in the
    original report). Without this flag, Terraform prints the SCC password
    in plain text in `plan`/`apply` output and stores it unmasked in the
    state file.
  - Two `connection {}` blocks in libvirt/host and ssh/host had
    `password = "linux"` hardcoded with no variable. Extracted to a
    `vm_password` variable (default: `"linux"`, sensitive: true).

  Note: `bastion_password` is pulled from the `provider_settings` map via
  `lookup()`. Terraform does not support marking a single map key as
  sensitive; addressing that would require a larger refactor and is left
  for a follow-up.

### DIFF
--- a/backend_modules/null/host/variables.tf
+++ b/backend_modules/null/host/variables.tf
@@ -131,3 +131,10 @@ variable "product_version" {
   type        = string
   default     = null
 }
+
+variable "vm_password" {
+  description = "password for SSH connection to provisioned VMs"
+  type        = string
+  default     = "linux"
+  sensitive   = true
+}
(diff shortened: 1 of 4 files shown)


NamelessOne91 (approved):

alichiel (approved):

backend_modules/null/host/variables.tf
```diff
@@ -131,3 +131,10 @@ variable "product_version" {
   type        = string
   default     = null
 }
+
+variable "vm_password" {
+  description = "password for SSH connection to provisioned VMs"
+  type        = string
+  default     = "linux"
```
Bischoff, line 138: ouch

Bischoff (changes requested):

Bischoff (approved):
after discussion

PR #2270, opened 2026-08-20
QE: Add flag variable to not install MLM when true

Pablogoliva:
## What does this PR change?

Add flag variable to not initialize the user, the api_test and the cobbler_tests instances if MLM/Uyuni is not installed. For the installation of MLM on RKE2 and RKE2 by itself in the testuite

### DIFF
--- a/modules/controller/variables.tf
+++ b/modules/controller/variables.tf
@@ -936,6 +936,11 @@ variable "install_kubectl_helm" {
   default = false
 }
 
+variable "install_uyuni_via_testsuite" {
+  description = "true to install mlm with the testsuite"
+  default = false
+}
+
 variable "kubeconfig_path" {
   description = "Path to a kubeconfig file on the host running Terraform; its contents will be copied to /root/.kube/config on the controller. Leave null to skip."
   default = null
--- a/modules/controller/variables.tf
+++ b/modules/controller/variables.tf
@@ -937,7 +937,7 @@ variable "install_kubectl_helm" {
 }
 
 variable "install_uyuni_via_testsuite" {
-  description = "true to install mlm with the testsuite"
+  description = "true to install mlm with the testsuite. This only exports UYUNI_NOT_INSTALLED to the controller, it does not gate the salt states: install_helm.sls, install_traefik.sls, install_kubernetes_server.sls and set_up_local-path-provisioner.sls are still keyed on is_slmicro_6_2, so setting this to true on any other OS installs mlm twice"
   default = false
 }
 
--- a/salt/controller/bashrc
+++ b/salt/controller/bashrc
@@ -268,6 +268,10 @@ export PRODUCT="Uyuni"
 export PRODUCT="SUSE Multi-Linux Manager"
 {% endif %}
 
+{% if grains.get('install_uyuni_via_testsuite') == true %}
+export UYUNI_NOT_INSTALLED=true
+{% endif %}
+
 # Ruby specific settings
 export GEM_PATH="/usr/lib64/ruby/gems/3.4.0"
 export PLAYWRIGHT_CLI_EXECUTABLE_PATH=/usr/local/bin/playwright
--- a/salt/kubernetes_common/set_up_local-path-provisioner.sls
+++ b/salt/kubernetes_common/set_up_local-path-provisioner.sls
@@ -1,7 +1,6 @@
 {% set osfullname = grains['osfullname'] %}
 {% set osrelease = grains['osrelease'] %}
 {% set is_sles_15_7 = osfullname == 'SLES' and osrelease == '15.7' %}
-{% set is_slmicro_6_2 = osfullname == 'SL-Micro' and osrelease == '6.2' %}
 {% set is_ubuntu = osfullname == 'Ubuntu' %}
 {% set is_tumbleweed = osfullname == 'openSUSE Tumbleweed' %}
 {% set is_supported_os = is_sles_15_7 or is_ubuntu or is_tumbleweed %}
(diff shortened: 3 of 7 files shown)


salt/controller/bashrc
```diff
@@ -268,6 +268,10 @@ export PRODUCT="Uyuni"
 export PRODUCT="SUSE Multi-Linux Manager"
 {% endif %}
 
+{% if grains.get('install_uyuni_via_testsuite') == false %}
+export UYUNI_NOT_INSTALLED=true
```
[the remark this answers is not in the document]
Pablogoliva: It's applies to the cucumber testsuite during runtime, I have tested  it, so this comment is not relevant.

modules/controller/variables.tf
```diff
@@ -936,6 +936,11 @@ variable "install_kubectl_helm" {
   default = false
 }
 
+variable "install_uyuni_via_testsuite" {
```
ktsamis, line 939: The name and the behaviour are opposites. `install_uyuni_via_testsuite = true` is the default and means sumaform installs it, and https://github.com/SUSE/susemanager-ci/pull/2191 sets it to `false` in exactly the two environments where the testsuite does the install.

The description says the real meaning: "true to install mlm with sumaform". Please rename it `install_uyuni_with_sumaform`, or flip the polarity so the name is true.

Same block in `modules/cucumber_testsuite/variables.tf`, and both tf files in https://github.com/SUSE/susemanager-ci/pull/2191 follow.

salt/controller/bashrc
```diff
@@ -268,6 +268,10 @@ export PRODUCT="Uyuni"
 export PRODUCT="SUSE Multi-Linux Manager"
 {% endif %}
 
+{% if grains.get('install_uyuni_via_testsuite') == false %}
```
ktsamis: The flag does not gate anything in salt. Every state that skips the install is still keyed on a hardcoded `is_slmicro_6_2`:

- `salt/server_kubernetes/install_kubernetes_server.sls` - `update_oci_app_version`, `build_helm_dependencies`, `install_uyuni_on_kubernetes`
- `salt/kubernetes_common/install_helm.sls` - `helm_install_package`, cert-manager, trust-manager
- `salt/kubernetes_common/install_traefik.sls` - `install_traefik`
- `salt/kubernetes_common/set_up_local-path-provisioner.sls` - `install_local_path_provisioner`

So one decision now has two independent switches. Set the flag on an RKE2 environment that is not SL Micro and sumaform installs MLM anyway, while the controller tells the testsuite it did not, and the testsuite then installs it a second time.

Driving those guards off the grain instead of the OS would make it one knob. If that is too much for this PR, please say so in the variable description so the next person does not assume the flag is authoritative.

salt/kubernetes_common/set_up_local-path-provisioner.sls
```diff
@@ -65,7 +73,7 @@ set_local-path-provisioner_selinux_tags:
 
 restart_local-path-provisioner_pods:
   cmd.run:
-    - name: kubectl delete pods --all -n {{ local_path_namespace }}
+    - name: kubectl delete pods --all -n local-path-storage
```
ktsamis: This was `{{ local_path_namespace }}` and is now hardcoded, four lines below where the same variable is exported to the shell as `LOCAL_PATH_NAMESPACE`. Nothing to do with the flag. Was it intentional?

ktsamis (changes requested):
Moving the `variables_*` states out of the SL Micro guards is the right fix, that part works. I checked every variable the new testsuite features assert and they all get exported now.

Two things on the flag itself and one unrelated change.

Order: this has to merge before https://github.com/SUSE/susemanager-ci/pull/2191, otherwise those tf files set a module variable that does not exist yet and terraform fails on the RKE2 job.

salt/kubernetes_common/set_up_local-path-provisioner.sls
Pablogoliva: No, it was a regression I didn't see. Good catch.

modules/controller/variables.tf
Pablogoliva: True, in the beginning it was going to be different and I changed the name. Let's change the logic then.

Pablogoliva: Solved here and in the CI.

salt/controller/bashrc
Pablogoliva: Good point. That's true, maybe we can also remove the flag of install_mlm_via_testsuite and leave it just for the slmicro 6.2 as default.

Pablogoliva: Yes, because those states don't work on SLmicro 6.2, because they need reboots between them. So, I decided to skip them on Slmicro 6.2.
Regarding the flag `install_uyuni_via_testsuite` I wanted to have it apart because Maxime said to reuse it for podman as well.

modules/controller/variables.tf
Pablogoliva: Yup, that's true. Changing the logic then

salt/controller/bashrc
Pablogoliva: I can skip the installation in the other OSes when that flag is true. That's a good option.

ktsamis (commented):
Re-reviewed against the current head. The rename/polarity and the hardcoded namespace are both fixed, thanks. One leftover from the flag-versus-OS-guard thread.

modules/controller/variables.tf
```diff
@@ -936,6 +936,11 @@ variable "install_kubectl_helm" {
...
+  description = "true to install mlm with the testsuite"
```
ktsamis: Could the description say that this flag does not gate the salt states? They are still keyed on `is_slmicro_6_2` in `install_helm.sls`, `install_traefik.sls`, `install_kubernetes_server.sls` and `set_up_local-path-provisioner.sls`, so setting this to `true` on a supported OS installs MLM twice.

One line here saves the next person from assuming the flag is authoritative.

ktsamis: Done in c2ca511a, on both `modules/controller/variables.tf` and `modules/cucumber_testsuite/variables.tf`.

ktsamis (approved):
Let's merge

PR #2277, opened 2026-09-07
Make "base_arm" a first class citizen

Bischoff:
## What does this PR change?

Make "base_arm" a first class citizen.

See also https://github.com/SUSE/susemanager-ci/pull/2216 .

DO NOT MERGE. I am not convinced of the merits of this PR. Maxime will tell.

### DIFF
--- a/modules/build_validation/main.tf
+++ b/modules/build_validation/main.tf
@@ -5,17 +5,13 @@ locals {
   host_rhlike  = lookup(var.module_base_configurations, "rhlike",  local.base_core)
   host_deblike = lookup(var.module_base_configurations, "deblike", local.base_core)
   base_retail  = lookup(var.module_base_configurations, "retail",  local.base_core)
+  base_arm     = lookup(var.module_base_configurations, "arm",     local.base_core)
 
   server_configuration = length(module.server_containerized) > 0 ? module.server_containerized[0].configuration : module.server[0].configuration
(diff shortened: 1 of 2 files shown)


Pablogoliva (approved):
LGTM

modules/build_validation/main.tf
```diff
@@ -5,6 +5,7 @@ locals {
   host_rhlike  = lookup(var.module_base_configurations, "rhlike",  local.base_core)
   host_deblike = lookup(var.module_base_configurations, "deblike", local.base_core)
   base_retail  = lookup(var.module_base_configurations, "retail",  local.base_core)
+  base_arm     = lookup(var.module_base_configurations, "arm",     local.base_core)
```
ktsamis, line 8: Nothing reads `local.base_arm` — the module's own `module "base_arm"` below is still what all six ARM minion/sshminion resources use, with `var.location`'s mirror. So the PRG pin in susemanager-ci#2216 has no effect, and with the multi-provider template the ARM base is built twice on the same host/pool (colliding `${name_prefix}${image}` volumes). Drop the internal `module "base_arm"` and switch those six to `local.base_arm`, like `deblike`/`retail`.

ktsamis (changes requested):
Dropping the hardcoded provider is right, but this has to land right before SUSE/susemanager-ci#2216 — on its own it breaks every BV deployment, since no caller passes `libvirt.host_arm` yet.

modules/build_validation/main.tf
Bischoff: See my answer to same comment on the other PR.

Bischoff:
Too complicated, too little benefit

PR #2282, opened 2026-09-09
Refresh expired systemsmanagement OBS signing key

ktsamis:
The copy of the systemsmanagement OBS key (`0xACAA9CF7E6E5A213`) in [salt/default/gpg_keys/saltstack.key](https://github.com/uyuni-project/sumaform/blob/master/salt/default/gpg_keys/saltstack.key) expired on 2026-02-13. Combustion imports it with `rpm --import` before adding any repo, so rpm keeps the expired copy and rejects `venv-salt-minion` as `NOTTRUSTED` after it was re-signed on 2026-09-08. That broke the first boot of SL Micro hosts in [manager-5.2-dev-acceptance-tests build 306](https://jenkins.mgr.suse.de/job/manager-5.2-dev-acceptance-tests/306/console).

Replaced with the current key from https://download.opensuse.org/repositories/systemsmanagement:/saltstack:/bundle:/next:/SLFO/SLFO/repodata/repomd.xml.key — same keyid, valid until 2028-04-09.


Pablogoliva (approved):

antogaro (approved):
lgtm

